Issue #34, opened 2017-01-16
Add .editorconfig


iocanel:
This will help having a consistent format, effortless for all IDEs.
It will also help reviewing too.

PR #35, opened 2017-01-16
#34: Add .editorconfig that defines 4 char ident for all files, but p…

iocanel:
…om.xml where ident uses tabs.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,7 @@
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+indent_style = tab
+indent_size = 4


.editorconfig
```diff
@@ -0,0 +1,10 @@
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+indent_style = space
```
spencergibb: All Spring Projects use tabs

iocanel: Even better! I'll fix that asap! :-)

iocanel:
@spencergibb: does it look ok now?

spencergibb:
Looks good now. We could probably add rules that are currently defined in https://github.com/spring-cloud/spring-cloud-build/blob/master/spring-cloud-dependencies-parent/eclipse-code-formatter.xml as an enhancement.

Issue #26, opened 2016-12-17
Add unit tests


spencergibb:

jrrickard:
@spencergibb I enjoy writing tests. Are PRs with additional tests welcome?

spencergibb:
absolutely!

jrrickard:
@spencergibb sorry for the delay, we had a baby. Working on this now, would you prefer multiple PRs (as well as issues) for tests across different modules (i.e. for tests for spring-cloud-kubernetes-discovery, tests for spring-cloud-kubernetes-config), or just in one big PR?

spencergibb:
NP. Congratulations!

I think it's easier to review smaller PRs, so I'd lean that way. @iocanel, thoughts?

iocanel:
Congratulations @jrrickard!

@spencergibb: I also prefer smaller PRs.

jrrickard:
@iocanel @spencergibb couple of follow up questions. Existing tests are mostly written in Groovy, but most other spring-cloud projects seem to have tests written in Java. I started writing in Java but realized I should check if the preference is Java or Groovy for these. 

Related question. If Groovy, should I upgrade the Spock version (1.0 -> 1.1) to support the use of SpringBootTest instead of the deprecated SpringApplicationConfiguration/IntegrationTest? Looks like the 1.1 version of Spock is still RC. 

Thanks!

spencergibb:
Java.

bparry02:
@jrrickard I'd like to contribute some unit tests as well. Which projects are you focusing on initially?

cmoulliard:
As we have started to add Groovy and/or Java Unit Tests, can we close this ticket ?

PR #40, opened 2017-01-29
Add null param check for StandardPodUtils and Unit test

bparry02:
A small start to #26!

Added a null check in StandardPodUtils, since an instance with a null Kube client would have limited usefulness and we should protect developers from a configuration error. If a developer decides not to use a KubernetesClient, they should probably implement the PodUtils interface and supply an alternate implementation instead of passing null to StandardPodUtils.

Holding off on additional tests for StandardPodUtils for now, since it will be tricky to simulate running inside a pod because of the file system checks, and we can't guarantee that the test is running outside of a pod.


iocanel:
LGTM

PR #50, opened 2017-03-11
#26: First draft of integration tests example.

iocanel:
This pull request adds an example which demonstrates a simple integration-test.

The test is using [arquillian cube for kubernetes](https://github.com/arquillian/arquillian-cube/blob/master/docs/kubernetes.adoc) to install the app in an isolated namesapce and then lookup and access the actual service (via portforwarding).

Note: The pull request introduces a snapshot dependency on arquillian-cube, but hopefully will be replaced with a released version shortly.


iocanel:
As soon as https://github.com/arquillian/arquillian-cube/issues/629 we need to replace the snapshot with the actual release.

PR #51, opened 2017-03-22
#26: Create unit test for LazilyInstantiate

bparry02:
Contribution toward #26


bparry02:
@spencergibb The new file uses tabs now. I did a force push to update my branch--let me know if that works.

PR #61, opened 2017-03-28
Add profile kubernetes for local build/deployment on minikube

cmoulliard:
Add profile kubernetes for local build/deployment on minikube and avoid CI build failure if the goal build is included within the f-m-p configuration section


iocanel:
@cmoulliard: shouldn't we also define the profile in the pom?

cmoulliard:
This is my fault. It has been included in the pom.xml file pushed within another pull request.

PR #63, opened 2017-03-30
Define service type using enricher section of the f-m-p config

cmoulliard:
Add the NodePort service type using enricher section of the f-m-p config


iocanel:
@cmoulliard: can you please add the same config to the other example, too?

PR #64, opened 2017-03-30
Define service type using enricher section of the f-m-p config

cmoulliard:
Define NodePort service type using enricher section of the f-m-p config


iocanel:
@cmoulliard: Thanks you very much! You rock :+1:

Issue #74, opened 2017-05-03
Add a more real Test case to demonstrate/test configMap


cmoulliard:
Add a Java Junit Test case containing a SpringBoot application exposing a HTTP APi where the message is found from the Kubernetes ConfigMap deployed within a KubernetesMock Server

iocanel:
I think we can close this now.

PR #75, opened 2017-05-03
Add a more real Java testcase

cmoulliard:
See Issue #74

### DIFF
--- a/spring-cloud-kubernetes-config/pom.xml
+++ b/spring-cloud-kubernetes-config/pom.xml
@@ -76,6 +76,11 @@
 			<artifactId>spock-spring</artifactId>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>io.rest-assured</groupId>
+			<artifactId>rest-assured</artifactId>
+			<scope>test</scope>
+		</dependency>
 
 		<dependency>
 			<groupId>org.springframework.boot</groupId>
--- a/spring-cloud-kubernetes-config/src/test/java/org/springframework/cloud/kubernetes/config/ConfigMapsSpringBootTest.java
+++ b/spring-cloud-kubernetes-config/src/test/java/org/springframework/cloud/kubernetes/config/ConfigMapsSpringBootTest.java
@@ -1,0 +1,113 @@
+/*
+ * Copyright (C) 2016 to the original authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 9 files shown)


spring-cloud-kubernetes-config/src/test/java/org/springframework/cloud/kubernetes/config/ConfigMapsSpringBootTest.java
```diff
@@ -75,11 +76,13 @@ public void beforeTest() {
 			.once();
 	}
 
+	@Ignore
```
iocanel: The ignored test cases, are actually pretty valid for the core module, which is responsible for injecting the client. I would move them there (unignored ofc).

spring-cloud-kubernetes-config/pom.xml
```diff
@@ -76,6 +76,12 @@
 			<artifactId>spock-spring</artifactId>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>io.rest-assured</groupId>
+			<artifactId>rest-assured</artifactId>
+			<version>3.0.2</version>
```
iocanel: Can we use a placeholder for that?

cmoulliard: > placeholder

Do you mean "add this dependency within a parent pom file" ?

spring-cloud-kubernetes-config/src/test/java/org/springframework/cloud/kubernetes/config/ConfigMapsSpringBootTest.java
cmoulliard: Do you want that we move this test 

```
	@Test
	public void testConfig() {
		assertEquals(config.getMasterUrl(),mockClient.getConfiguration().getMasterUrl());
		assertEquals(config.getNamespace(),mockClient.getNamespace());
	}
```
within another a new Test Class of the spring-cloud-kubernetes-core project ?

spring-cloud-kubernetes-config/pom.xml
iocanel: I mean adding a property like 

    <restassured.version>3.0.2</restassured.version>

either on the top of the current pom, or maybe in the parent (if we are using restassured in more than one places).

spring-cloud-kubernetes-config/src/test/java/org/springframework/cloud/kubernetes/config/ConfigMapsSpringBootTest.java
iocanel: Short answer: Yes

Longer answer: Having an ignored test, that doesn't test anything specific to the current module is not really clean. So I was thinking that we could move it to the core where its more relevant.

cmoulliard:
Squashed into one commit. There are no longer @Ignore annotations with the latest commit. Do you want to do something else ?

PR #86, opened 2017-05-18
Change ServerListRefreshInterval and Hystrix timeoutInMilliseconds

cmoulliard:
- Change ServerListRefreshInterval and Hystrix timeoutInMilliseconds.
- Fix typo error within the README file


iocanel:
@cmoulliard: It looks good to me, but can you pleaase, tell us why are you tuning the refresh interval?
Does it have to do with the unit tests?

cmoulliard:
> are you tuning

I have reduced the refreshTime to avoid that when the user will play with this scenario on Kubernetes/OpenShift platform he/she has to wait too long till the endpoints will be registered again after the pods were killed and to verify that the circuit breaker doesn't reply anymore with "Fallback" message

iocanel:
That's good enough for me.

Issue #91, opened 2017-05-22
Add zipkin test case


cmoulliard:
The module spring-cloud-kubernetes-zipkin doesn't contain any test or mock test case. I propose to add such a test to validate the logic implemented

PR #99, opened 2017-05-29
Add example supporting zikin

cmoulliard:
- Instead of a test case, a real example has been created to cover #91 
- Project has been tested using minishift or minikube


iocanel:
I am not so sure about the example.

- It doesn't feel concise. (lot of different methods that are not obvious how they fit together: start, call, hi etc). 
- It does a lot of things manually, that are not needed (no need to create spans, by hand, not sure about tags).

cmoulliard:
Code is coming from Spring Cloud Sleuth example. I will improve it

iocanel:
@cmoulliard: Then I may miss something obvious here. 
@spencergibb: Your thoughts?

cmoulliard:
@iocanel I think that the idea of Spencer was to demonstrate the Spring Sleuth possibilities as presented within this [publication](http://www.baeldung.com/spring-cloud-sleuth-single-application). This is the reason why different endpoints have been created ....

spencergibb:
Using the sleuth api's shouldn't be the focus, just a sample running in kubernetes is the value I see.

cmoulliard:
> focus

I will revamp the code to just expose 2 endpoints to show a trace and trace + span when a service call another one in order to simplify the lines of code

cmoulliard:
@spencergibb @iocanel Can you review my modifications please ?

iocanel:
This looks ok to me!

I am not crazy about mixing the explicit (using the Tracer) and aop style together (cause its not obvious which is the minimum required configuration to make it work.
 
Maybe in the future they could be two distinct examples.

PR #101, opened 2017-06-06
Add a new module supporting jaeger - opentracing

cmoulliard:
- Added a new maven module supporting jaeger - opentracing
- Added a starter to deploy a Spring Boot Jaeger application
- Project has been tested using the ola [spring boot application](https://github.com/cmoulliard/ola/tree/jaeger)

### DIFF
--- a/spring-cloud-kubernetes-dependencies/pom.xml
+++ b/spring-cloud-kubernetes-dependencies/pom.xml
@@ -20,6 +20,9 @@
 		<mockwebserver.version>0.0.12</mockwebserver.version>
 		<restassured.version>3.0.2</restassured.version>
 		<spock-spring.version>1.0-groovy-2.3</spock-spring.version>
+		<version.jaeger>0.19.0</version.jaeger>
+		<io.opentracing.version>0.22.0</io.opentracing.version>
+		<io.opentracing.contrib.spring.autoconfigure.version>0.0.4</io.opentracing.contrib.spring.autoconfigure.version>
 	</properties>
 	<dependencyManagement>
 		<dependencies>
@@ -74,6 +77,11 @@
 				<version>${project.version}</version>
 			</dependency>
 
+			<dependency>
+				<groupId>org.springframework.cloud</groupId>
+				<artifactId>spring-cloud-kubernetes-jaeger</artifactId>
+				<version>${project.version}</version>
+			</dependency>
 
 			<!-- Own dependencies - Starters -->
 			<dependency>
@@ -105,6 +113,27 @@
 				<artifactId>spring-cloud-starter-kubernetes-all</artifactId>
 				<version>${project.version}</version>
 			</dependency>
+
+			<!-- OpenTracing -->
+			<dependency>
+				<groupId>io.opentracing</groupId>
+				<artifactId>opentracing-noop</artifactId>
+				<version>${io.opentracing.version}</version>
+			</dependency>
+
+			<dependency>
+				<groupId>io.opentracing.contrib</groupId>
+				<artifactId>opentracing-spring-web-autoconfigure</artifactId>
+				<version>${io.opentracing.contrib.spring.autoconfigure.version}</version>
+			</dependency>
+
+			<!-- OpenTracing implementation - Jaeger -->
+			<dependency>
+				<groupId>com.uber.jaeger</groupId>
+				<artifactId>jaeger-core</artifactId>
+				<version>${version.jaeger}</version>
+			</dependency>
+
 
 			<!-- Testing Dependencies -->
 			<dependency>
@@ -159,6 +188,8 @@
 				<artifactId>spock-spring</artifactId>
 				<version>${spock-spring.version}</version>
 			</dependency>
+
+
 		</dependencies>
 	</dependencyManagement>
 	<profiles>
--- a/spring-cloud-kubernetes-jaeger/src/main/java/org/springframework/cloud/kubernetes/jaeger/JaegerKubernetesAutoConfiguration.java
+++ b/spring-cloud-kubernetes-jaeger/src/main/java/org/springframework/cloud/kubernetes/jaeger/JaegerKubernetesAutoConfiguration.java
@@ -17,6 +17,9 @@
 
 package org.springframework.cloud.kubernetes.jaeger;
 
+import java.net.InetAddress;
+import java.net.MalformedURLException;
+import java.net.URL;
 import java.util.List;
 import java.util.Optional;
 import java.util.stream.Collectors;
@@ -54,7 +57,7 @@ public class JaegerKubernetesAutoConfiguration {
 	private static final Log LOG = LogFactory.getLog(JaegerKubernetesAutoConfiguration.class);
 
     @Bean
-    public Tracer tracer(KubernetesClient client, KubernetesJaegerDiscoveryProperties discoveryProperties) {
+    public Tracer tracer(KubernetesClient client, KubernetesJaegerDiscoveryProperties discoveryProperties) throws Exception {
 
     	String serviceName = discoveryProperties.getServiceName();
     	String traceServerName = discoveryProperties.getTracerServerName();
@@ -85,8 +88,9 @@ public class JaegerKubernetesAutoConfiguration {
                 .collect(Collectors.toList());
     }
 
-	private Tracer jaegerTracer(String url, String serviceName) {
-		Sender sender = new UdpSender(url, 0, 0);
+	private Tracer jaegerTracer(String serviceURL, String serviceName) throws Exception {
+		URL url = new URL(serviceURL);
+		Sender sender = new UdpSender(url.getHost(), url.getPort(), 0);
 		return new com.uber.jaeger.Tracer.Builder(serviceName,
 			new RemoteReporter(sender, 100, 50,
 				new Metrics(new StatsFactoryImpl(new NullStatsReporter()))),
--- a/spring-cloud-kubernetes-jaeger/src/main/java/org/springframework/cloud/kubernetes/jaeger/JaegerKubernetesAutoConfiguration.java
+++ b/spring-cloud-kubernetes-jaeger/src/main/java/org/springframework/cloud/kubernetes/jaeger/JaegerKubernetesAutoConfiguration.java
@@ -17,13 +17,24 @@
 
 package org.springframework.cloud.kubernetes.jaeger;
 
-import java.net.InetAddress;
-import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.List;
 import java.util.Optional;
 import java.util.stream.Collectors;
 
+import com.uber.jaeger.metrics.Metrics;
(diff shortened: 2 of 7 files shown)


spring-cloud-kubernetes-dependencies/pom.xml
```diff
@@ -20,6 +20,9 @@
 		<mockwebserver.version>0.0.12</mockwebserver.version>
 		<restassured.version>3.0.2</restassured.version>
 		<spock-spring.version>1.0-groovy-2.3</spock-spring.version>
+		<version.jaeger>0.19.0</version.jaeger>
+		<io.opentracing.version>0.22.0</io.opentracing.version>
```
adriancole: fyi 0.22 is api incompatible with 0.30

adriancole:
my 2p: this seems like more a community project than something we'd have in spring-cloud-kubernetes. sleuth is the spring-cloud's tracing solution. We don't want to inherit maintenance for third party apis, particularly api unstable ones never requested by an end user

spring-cloud-kubernetes-dependencies/pom.xml
pavolloffay: This is fine here, these versions match

spring-cloud-kubernetes-jaeger/src/main/java/org/springframework/cloud/kubernetes/jaeger/KubernetesJaegerDiscoveryProperties.java
```diff
@@ -0,0 +1,56 @@
...
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+/**
+ * @author <a href="mailto:cmoullia@redhat.com">Charles Moulliard</a>
+ */
+@ConfigurationProperties("spring.cloud.kubernetes.jaeger.discovery")
+public class KubernetesJaegerDiscoveryProperties {
+
+	private String tracerServerName = "jaeger-all-in-one";
```
pavolloffay: Should be called `agentServiceName` data could be sent to agent and collector.

spring-cloud-kubernetes-jaeger/src/main/java/org/springframework/cloud/kubernetes/jaeger/JaegerKubernetesAutoConfiguration.java
```diff
@@ -0,0 +1,97 @@
...
+                .stream()
+                .flatMap(s -> s.getAddresses().stream().map(a -> (ServiceInstance) new KubernetesServiceInstance(name, a ,s.getPorts().stream().findFirst().orElseThrow(IllegalStateException::new), false)))
+                .collect(Collectors.toList());
+    }
+
+	private Tracer jaegerTracer(String serviceURL, String serviceName) throws Exception {
+		URL url = new URL(serviceURL);
+		Sender sender = new UdpSender(url.getHost(), url.getPort(), 0);
+		return new com.uber.jaeger.Tracer.Builder(serviceName,
+			new RemoteReporter(sender, 100, 50,
```
pavolloffay: Some of these properties should be configurable, see https://github.com/uber/jaeger-client-java/tree/master/jaeger-core#configuration-via-environment at least queue size and flush interval.

pavolloffay:
slueth is not OpenTracing compatible and we would like to go that direction.

adriancole:
>
> slueth is not OpenTracing compatible and we would like to go that
> direction.
>
feel free to host your own integration. If users ask about it, we can
mention how to find yours!

adriancole:
>
> +		<version.jaeger>0.19.0</version.jaeger>
> +		<io.opentracing.version>0.22.0</io.opentracing.version>
>
> This is fine here, these versions match
>
it isn't fine. the point is that this api breaks constantly and will
introduce thrash. bringing unstable components into a repo isn't something
to take lightly.

adriancole:
There's community effort (by a user of sleuth) here
https://github.com/DealerDotCom/sleuth-opentracing if opentracing
becomes stable and/or activity picks up there, we can consider
formalizing.

Adding this to the core repo ahead of demand and with known unstable
apis is just not a right time thing

cmoulliard:
> it isn't fine. the point is that this api breaks constantly and will
> introduce thrash. bringing unstable components into a repo isn't something
> to take lightly.

It's true about api breaks (like it has also been the case for AMQP or WebSocket till 1.0 have been released). However, the OT API version 1.0 should [be out soon]( https://github.com/opentracing/opentracing-java/milestone/1) (although the date is not settled) and will stabilize existing situation. Those breaking changes happen because of major new features like active span in-process propagation.

Sleuth combines a variety of instrumentations and is harder to plug in instrumentation for a different framework, on the other hand, pure OT instrumentation allows us easily plug in different instrumentation together.

So, if you refuse this PR because the OT Api isn't stable enough, I will argue you that you have implemented a Spring Boot starter for AMQP - 0.9 while the spec wasn't stable enough. This is not because a spec API is not yet stable that we should not propose/provide an implementation as proposed here based on the Uber Jaeger project which is not at all a "community project" as you suggest but a Microservices Distributed Tracing Project [1] [2] offering more features than Zipkin

Moreover, as an API adapter has been proposed to the OpenTracing foundation [3] for spring Cloud Sleuth, that makes a lot of sense to continue to develop this PR, module. Do you agree ? 

[1] : https://eng.uber.com/distributed-tracing/
[2] : https://uber.github.io/jaeger/
[3] : https://github.com/opentracing-contrib/meta/issues/5

spring-cloud-kubernetes-jaeger/src/main/java/org/springframework/cloud/kubernetes/jaeger/JaegerKubernetesAutoConfiguration.java
cmoulliard: +1

spring-cloud-kubernetes-jaeger/src/main/java/org/springframework/cloud/kubernetes/jaeger/KubernetesJaegerDiscoveryProperties.java
cmoulliard: Will do that.

adriancole:
I'd recommend closing this issue

spring-cloud-kubernetes-dependencies/pom.xml
cmoulliard: I will update the version as jaeger-core - 0.19.0 has a dep with opentracing 0.22

```
[INFO] com.uber.jaeger:jaeger-core:jar:0.19.0
[INFO] +- com.uber.jaeger:jaeger-thrift:jar:0.19.0:compile
[INFO] |  \- org.apache.thrift:libthrift:jar:0.9.2:compile
[INFO] |     +- org.apache.httpcomponents:httpclient:jar:4.2.5:compile
[INFO] |     |  +- commons-logging:commons-logging:jar:1.1.1:compile
[INFO] |     |  \- commons-codec:commons-codec:jar:1.6:compile
[INFO] |     \- org.apache.httpcomponents:httpcore:jar:4.2.4:compile
[INFO] +- io.opentracing:opentracing-api:jar:0.22.0:compile
[INFO] +- com.google.code.gson:gson:jar:2.8.0:compile
[INFO] +- org.slf4j:slf4j-api:jar:1.7.16:compile
```

cmoulliard:
> I'd recommend closing this issue

Why ?

adriancole:
The "community project" reference was not a qualitative reference about jaeger. It was that you can absolutely host this integration on your own, as a person in a community, in jaeger's org or opentracing's org or somewhere else.

I'm not going to restate the myriad of reasons why it is not the right time, both from a stability pov, or a relevance POV to put something like this in the core repo!

adriancole:
ps I'm not a maintainer of this project, so if contributors here decide to merge and maintain this ahead of user demand etc, that's its prerogative. I'm only commenting as a couple people were confused by this out-of-the-blue PR

iocanel:
I will not comment about api stability, as it refers to things I am not familiar with.

I am however, familiar with `which integration, should be hosted where` kind of talks. So I'll comment on that:

I don't see why we can't host this module here, same way we do for ribbon, hystrix, zipkin.
I also don't see why a feature has to be requested by a user first (especially for an incubating project).

Now, if spring-cloud has a policy like: "we host only one solution per problem (e.g. for tracing we will only host sleuth based stuff)", then I guess as a project that aspires to graduate, we will have to respect that.

cmoulliard:
This PR offers the possibility for Spring project running on Google Kubernetes or Openshift platform to really benefit about a modern OpenTracing API, where instrumentation exists, is independent of the implementation, is vendor neutral.

If you choose zipkin you have to stick with zipkin, OT allows you to change tracing backend with literally zero configuration change. There is no tracer API in zipkin - e.g. zipkin java tracer (brave) has totally different API that tracer in ruby or python.

Because OpenTracing is vendor neutral it is possible (and it is a goal of OT) is to write instrumentation only once. OpenTracing instrumentation can be reused for Zipkin or Jaeger. 

The PR module proposed here could be turned into zipkin with minimal changes only tracer creation.

So far all tracing system had to write instrumentation by themselves, the same over and over (instana, datadog, new relic...). OT allows you to write it only once and then focus on thighs which really matters.

So why opentracing --> http://opentracing.io/documentation/#why-opentracing

maxromanovsky:
@cmoulliard I don't get why it's k8s-specific. What about consul or netflix users? Also, IMHO there would be little to no people to support it, and it doesn't belong to k8s repo.

spencergibb:
It's been marked for team discussion, which we will discuss tomorrow.

adriancole:
@cmoulliard I've read the website and the tag line. Show me the tests that prove any of this works in practice. You do know stagemonitor needed to write a shim to work with two opentracing implementations. Now, they are rev-locked.. until everyone updates the same version.

screen shot 2017-06-07 at 6 58 51 pm

What you are doing now is wasting more time, and honestly I'm done responding to your trolls.

spencergibb:
I'll reopen discussion after our meeting tomorrow

spencergibb:
Thanks for the PR. There are a few things we'll comment on:
 
1) This is not the appropriate project for this PR. We’re focusing our tracing efforts in the Spring Cloud Sleuth project: https://github.com/spring-cloud/spring-cloud-sleuth.
 
2) There have been no requests for specific Jaeger support. (Also, Jaeger specifically supports Zipkin spans: http://jaeger.readthedocs.io/en/latest/getting_started/#migrating-from-zipkin.)
 
3) On sleuth specifically (because this isn't the sleuth repo), we don't have demand to justify a strict dependency on OpenTracing. We would prefer that exposing OpenTracing apis optionally within Sleuth start as a community project, such as the existing OpenTracing integration for Spring Cloud Sleuth (https://github.com/DealerDotCom/sleuth-opentracing). We feel that we should wait to incorporate this support into a core project. We can revisit the topic once the community repository is more popular and apis are stable.

PR #112, opened 2017-07-10
Review README file.

cmoulliard:
Review content of the README file to correct mistakes, improve content, add missing stuffs

PR #118, opened 2017-07-26
#112: Fixed logger.

metacosm:


spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/ConfigUtils.java
```diff
@@ -1,18 +1,17 @@
 package org.springframework.cloud.kubernetes.config;
 
+import io.fabric8.kubernetes.client.KubernetesClient;
```
spencergibb: Please revert import change

metacosm:
Import change reverted. Sorry about that.

PR #125, opened 2017-08-30
allow disabling reading config maps over api

apixandru:
We're blocked by company policy to read config maps from API so the way that I read them is to mount the config map as an external file (similar to the way secrets are dealt with if we have the spring.cloud.kubernetes.secrets.paths set up) and then read it as a regular 'application.yml' file with no extra code required to achieve this.

The problem is that there is no way to disable config map api checks so we end up getting the warning stacktrace that the service account doesn't have view permission.

Currently, the only solution offered for this particular case is to raise the log level to error but I think it would be better to just have the option to skip that part if you want.

### DIFF
--- a/spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/ConfigMapConfigProperties.java
+++ b/spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/ConfigMapConfigProperties.java
@@ -24,18 +24,9 @@ public class ConfigMapConfigProperties extends AbstractConfigProperties {
 
     private static final String TARGET = "Config Map";
 
-	private boolean readConfigsFromApi = true;
-
-	public boolean isReadConfigsFromApi() {
-		return readConfigsFromApi;
-	}
-
-	public void setReadConfigsFromApi(boolean readConfigsFromApi) {
-		this.readConfigsFromApi = readConfigsFromApi;
-	}
-
     @Override
     public String getConfigurationTarget() {
         return TARGET;
     }
+
 }
--- a/spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/BootstrapConfiguration.java
+++ b/spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/BootstrapConfiguration.java
@@ -36,20 +36,28 @@ public class BootstrapConfiguration {
 
     @Configuration
     @Import(KubernetesAutoConfiguration.class)
-    @EnableConfigurationProperties({ ConfigMapConfigProperties.class, SecretsConfigProperties.class })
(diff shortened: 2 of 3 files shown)


spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/ConfigMapConfigProperties.java
```diff
@@ -24,6 +24,16 @@
 
     private static final String TARGET = "Config Map";
 
+	private boolean readConfigsFromApi = true;
```
spencergibb: I would go with `disabled = false`.

apixandru:
While 'disabled' is gives quite a direct indication of what it does, the reason why I chose such a specific name is that we also have 'enabled' in the same class and it doesn't do what I would expect.

The current behavior of spring.cloud.kubernetes.config.enabled doesn't limit itself to config maps, [it also expands to secrets](https://github.com/apixandru/spring-cloud-kubernetes/blob/32d982977932401db6744709e4406afb79b28ab6/spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/BootstrapConfiguration.java#L40
).  

So if i were to enable the disabled variables, to only get secrets i would have to have spring.cloud.kubernetes.config.enabled=true and spring.cloud.kubernetes.config.disabled=true which is quite counter intuitive.

Now, since the enabled property is in the abstract class and is never used, I suspect that the original intention of the author was to separate enabling of config maps and secrets by having two inner configuration classes KubernetesConfigMapPropertySourceConfiguration and KubernetesSecretsPropertySourceConfiguration that are enabled individually and i think that they were bundled together because of an oversight.

The correct thing to do in my opinion would be to separate them but that may break some existing code out there when they update if they relied on the fact that spring.cloud.kubernetes.config.enabled set to false also disables secrets.

What do you think?

apixandru:
@spencergibb did you get a chance to read my previous message?

spencergibb:
I'd prefer to use the `enabled` property then. If the original author could chime in @iocanel?

apixandru:
I went ahead and separated the property sources configuration to what I suspect was the original intention.

apixandru:
any updates on this?

iocanel:
It looks good to me.

apixandru:
@spencergibb anything that you would like to add?

stela:
I just noticed this pull request, #142 also allows disabling ConfigMap API requests (in case you know you'll only be using a ConfigMap mounted as a volume or exported as environment variables). This pull request still seems useful though in providing more global "feature toggles" for the entire ConfigMaps and Secrets PropertySources (if I understand the code correctly). Also, looks like the README should be fixed to mention only the ConfigMap PropertySource for `spring.cloud.kubernetes.config.enabled`. `spring.cloud.kubernetes.secrets.enabled` already has the correct description.

PR #143, opened 2018-01-10
Remove duplicate logic from addKubernetesProfile

gytis:
First part of the `addKubernetesProfile` method executes the same logic as the second part, just without logging.

### DIFF
--- a/spring-cloud-kubernetes-core/src/main/java/org/springframework/cloud/kubernetes/profile/KubernetesProfileApplicationListener.java
+++ b/spring-cloud-kubernetes-core/src/main/java/org/springframework/cloud/kubernetes/profile/KubernetesProfileApplicationListener.java
@@ -16,8 +16,6 @@
 
 package org.springframework.cloud.kubernetes.profile;
 
-import io.fabric8.kubernetes.api.model.Pod;
-
 import org.apache.commons.logging.Log;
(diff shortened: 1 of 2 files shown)


spring-cloud-kubernetes-core/src/main/java/org/springframework/cloud/kubernetes/profile/KubernetesProfileApplicationListener.java
```diff
@@ -47,13 +45,6 @@ public void onApplicationEvent(ApplicationEnvironmentPreparedEvent event) {
     }
 
     void addKubernetesProfile(ConfigurableEnvironment environment) {
-        Pod current = utils.currentPod().get();
```
spencergibb: Is this the same as `utils.isInsideKubernetes()`?

gytis: Yes it is (https://github.com/spring-cloud-incubator/spring-cloud-kubernetes/blob/master/spring-cloud-kubernetes-core/src/main/java/org/springframework/cloud/kubernetes/StandardPodUtils.java#L55):
```java
@Override
public Boolean isInsideKubernetes() {
    return currentPod().get() != null;
}
```

PR #145, opened 2018-01-17
Use configmap name when configuring an archaius watcher

gytis:


iocanel:
Thanks for spotting it!

Issue #150, opened 2018-02-14
Publish heartbeat events when service catalogue changes


joshiste:
The KubernetesDiscoveryClient should publish HearbeatEvents when there are changes to the service catalog to actively notify the application.

We need this for example in Spring Boot Admin...
... and the other discovery implementations I know (eureka, consul, zookeeper, cloudfoundry) are implementing this...

kaidowei:
The PR #154 did not solve the problem, because the `KubernetesCatalogWatch.java` publishes `HeartBeatEvents` with the list of services as the state. This list will rarely/never change and therefore a listener of those events cannot decide if an update is needed or not. Other implementations like [consul](https://github.com/spring-cloud/spring-cloud-consul/blob/master/spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulCatalogWatch.java) or [eureka](https://github.com/spring-cloud/spring-cloud-netflix/blob/eb0681f9c25c62c15b368674a2e0bf082663923c/spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/CloudEurekaClient.java) publish an ever increasing counter in those `HeartBeatEvents`.

kaidowei:
@akaGelo thanks for the fast response. Where can I download the newest snapshot so that I can verify the correct behaviour with spring boot admin?

akaGelo:
[repo](https://github.com/spring-cloud-incubator/spring-cloud-kubernetes/blob/3ad6fa14e14daa4d6b49d81e4662e5ccbf74a916/pom.xml#L193)

maven:
```
<repository>
           <id>spring-snapshots</id>
            <name>Spring Snapshots</name>
            <url>https://repo.spring.io/libs-snapshot-local</url>
</repository>
```

[last snapshot build version 0.3.0.BUILD-SNAPSHOT ](https://repo.spring.io/libs-snapshot-local/org/springframework/cloud/spring-cloud-kubernetes-discovery/0.3.0.BUILD-SNAPSHOT/)

spencergibb:
Closed by #209

PR #154, opened 2018-02-24
KubernetesCatalogWatch

akaGelo:
#150  and [https://github.com/codecentric/spring-boot-admin/issues/643](https://github.com/codecentric/spring-boot-admin/issues/643)

Publish HeartbeatEvent if services list changed.

Inspired by ConsulCatalogWatch.java


spring-cloud-kubernetes-discovery/src/main/java/org/springframework/cloud/kubernetes/discovery/KubernetesCatalogWatch.java
```diff
@@ -0,0 +1,66 @@
...
+import org.springframework.scheduling.annotation.Scheduled;
+
+import java.util.List;
+import java.util.concurrent.atomic.AtomicReference;
+
+
+/**
+ * @author Oleg Vyukov
+ */
+@Slf4j
```
shuraa: Oleg, I don't think that Lombok really useful here. It is an only case then you use it.

akaGelo: I have removed the lombok.

akaGelo:
Yes. I'm sorry that I did not do it right away. Problems with unavailability of repo.spring.io in Russia lead to problems with local build. I had to use CI for testing

PR #161, opened 2018-03-09
Populate ServiceInstance Metadata from Service Labels

shahbour:
I wanted to run `spring-boot-admin` on kubernetes and found that for services with the context-map set to anything other than /  it will not work.  

I needed to populate some data in `Metadata` in `org.springframework.cloud.client.ServiceInstance` and found that currently it is set to `EmptyMap` so I thought we can customize this by getting them from service labels for example
```
apiVersion: v1
kind: Service
metadata:
  labels:
    app: ui
    management.context-path: ui
  name: ui
  namespace: default
spec:
  clusterIP: 10.103.4.107
  ports:
  - port: 8085
    protocol: TCP
    targetPort: 8085
  selector:
    app: ui
  sessionAffinity: None
  type: ClusterIP
```
In the above example it is read from `management.context-path`

### DIFF
--- a/spring-cloud-kubernetes-discovery/src/main/java/org/springframework/cloud/kubernetes/discovery/KubernetesDiscoveryClient.java
+++ b/spring-cloud-kubernetes-discovery/src/main/java/org/springframework/cloud/kubernetes/discovery/KubernetesDiscoveryClient.java
@@ -17,6 +17,8 @@
 package org.springframework.cloud.kubernetes.discovery;
 
 import io.fabric8.kubernetes.api.model.Endpoints;
+import io.fabric8.kubernetes.api.model.ObjectMeta;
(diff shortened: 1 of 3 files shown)


iocanel:
Can you rebase this one please?

iocanel:
Also, there is a relevant discussion going on on:#166 where you might want to share your view.

Salaboy:
@shahbour @iocanel I think that we can close this, because we are already doing this. I'm also testing with spring-boot-admin so this should be working already. But this PRs are related: https://github.com/spring-cloud-incubator/spring-cloud-kubernetes/pull/166
and 
https://github.com/spring-cloud-incubator/spring-cloud-kubernetes/pull/187

when #187 is merged the spring boot app admin should work as expected.

shahbour:
Ok cool , thanks for the update

Issue #106, opened 2017-06-22
K8S ConfigMap doesn't work very well with profile


cmoulliard:
When the following configMap is defined 

```
metadata:
  name: store
data:
  application.yml: |-
    greeting:
      message: Say Hello to the World
    ---
    spring:
      profiles: development
    greeting:
      message: Say Hello to the Developers
    ---
    spring:
      profiles: production
    greeting:
      message: Say Hello to the Ops
```

and the profile `development` specified

```
exec java -Dspring.profiles.active=development -javaagent:/opt/jolokia/jolokia.jar=config=/opt/jolokia/etc/jolokia.properties -cp . -jar /deployments/store-1.0-SNAPSHOT.jar
I> No access restrictor found, access to any MBean is allowed
Jolokia: Agent started with URL https://172.17.0.13:8778/jolokia/
14:40:35.904 [background-preinit] INFO  o.h.validator.internal.util.Version - HV000001: Hibernate Validator 5.3.5.Final
14:40:36.195 [main] INFO  o.s.c.a.AnnotationConfigApplicationContext - Refreshing org.springframework.context.annotation.AnnotationConfigApplicationContext@7e0babb1: startup date [Thu Jun 22 14:40:36 UTC 2017]; root of context hierarchy
14:40:36.524 [main] INFO  o.s.c.s.PostProcessorRegistrationDelegate$BeanPostProcessorChecker - Bean 'configurationPropertiesRebinderAutoConfiguration' of type [org.springframework.cloud.autoconfigure.ConfigurationPropertiesRebinderAutoConfiguration$$EnhancerBySpringCGLIB$$f55351cc] is not eligible for getting processed by all BeanPostProcessors (for example: not eligible for auto-proxying)
  .   ____          _            __ _ _
 /\\ / ___'_ __ _ _(_)_ __  __ _ \ \ \ \
( ( )\___ | '_ | '_| | '_ \/ _` | \ \ \ \
 \\/  ___)| |_)| | | | | || (_| |  ) ) ) )
  '  |____| .__|_| |_|_| |_\__, | / / / /
 =========|_|==============|___/=/_/_/_/
 :: Spring Boot ::        (v1.5.3.RELEASE)
14:40:37.000 [main] INFO  o.s.c.b.c.PropertySourceBootstrapConfiguration - Located property source: ConfigMapPropertySource [name='configmap.store.jbcnf']
[...]

PR #162, opened 2018-03-09
Support Global or Common application config map

shahbour:
Support Global config map called application
Support Specific config map per active profile
So a service with name sample and active profiles production , mysql will try to get the following config maps

application
sample
sample-production
sample-mysql

This is related to [issue](https://github.com/spring-cloud-incubator/spring-cloud-kubernetes/issues/153) create by me

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,9 +88,11 @@ The [Spring Cloud Kubernetes Config](./spring-cloud-kubernetes-config) project m
 during application bootstrapping and triggers hot reloading of beans or Spring context when changes are detected on 
 observed `ConfigMap`s.
 
-`ConfigMapPropertySource` will search for a Kubernetes `ConfigMap` which `metadata.name` is either the name of 
-your Spring application (as defined by its `spring.application.name` property) or a custom name defined within the
+`ConfigMapPropertySource` will search for a Kubernetes `ConfigMap` as below
+- `global` is a global common config map (This can be used to hold global configuration for all pods)
+- `metadata.name` is either the name of your Spring application (as defined by its `spring.application.name` property) or a custom name defined within the
 `bootstrap.properties` file under the following key `spring.cloud.kubernetes.config.name`.
+- `metadata.name-profile` same name 2nd step but with active profiles added
 
 If such a `ConfigMap` is found, it will be processed as follows:
 
--- a/spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/ConfigMapPropertySource.java
+++ b/spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/ConfigMapPropertySource.java
@@ -37,49 +37,78 @@ import org.springframework.beans.factory.config.YamlPropertiesFactoryBean;
 import org.springframework.core.io.ByteArrayResource;
 import org.springframework.util.StringUtils;
 
+/**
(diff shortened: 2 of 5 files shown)


iocanel:
I have mixed feelings about this.

I do like the idea of being able to have shared configuration between multiple pods, this is really great.

I am not sure if doing that via the application configmap is the way to go. Why?
- the name is misleading
- it changes the way that application is used so far.

So I'd rather use a different name, like `shared` or `global`. @spencergibb: Your thoughts?

shahbour:
@spencergibb @iocanel Any update on this?

spencergibb:
I'm fine with a different name as long as it is documented

shahbour:
Ok great, I will go with global and update the code

shahbour:
I changed the name to global and updated documentation and tests, below are how it works now

Support Global config map called global
Support Specific config map per active profile
So a service with name sample and active profiles production, mysql, will try to get the following config maps

global
sample
sample-production
sample-mysql

geoand:
I am working on more generalized solution to this problem that would allow users to specify arbitrary `ConfigMaps` (name and namespace) instead of requiring specific names.
I am hoping to have a working prototype soon (supporting context reload might be a challenge, otherwise it seems doable)

shahbour:
After rebasing and getting to latest commit I noticed that if the config map does not have an active profile (spring.profiles) is empty or not set then it is ignored and not added, this was not the case before.

Is this on purpose or it is a bug?

iocanel:
@geoand do u have an idea?

geoand:
@iocanel I will check in a few hours and let you know

geoand:
@iocanel @shahbour The change that was introduced in #174 basically fixes #106 for the 0.3.x release. 
The initial refactoring that was done to bring the project inline with Spring Boot 2.x, had introduced a regression and #106 was once again an issue.

As far as I can tell, the state of the `yaml` handling currently in `master` is the intended one, meaning that it's working as it was for `0.2.0.RELEASE` and not `0.3.0.RC1` which broke the intended functionality.

@shahbour Could you provide an example of what you think might not be working correctly?

shahbour:
I will try to do a test unit that fails. To explain what I faced. After I did the update all my config maps that don't include `spring.profiles` were ignored for example the below config map did not work until I added spring.profiles: kubernetes

```
apiVersion: v1
data:
  application.yml: |-
    logging:
      level:
        com.xxxxxx: INFO
    spring:
      profiles: kubernetes
      cache:
        type: hazelcast
```

I think the reason for this is this line 
```
yamlFactory.setDocumentMatchers(
					(DocumentMatcher) properties ->
						(asList(profiles).contains(properties.getProperty("spring.profiles")) ?
							FOUND : NOT_FOUND)
				);
```

geoand:
@shahbour I can see why that would be a problem, however the current behavior is the same as existed in `0.2.0.RELEASE`.
If anything I would say that automatically adding the `kubernetes` is problematic. @iocanel Do you remember what the reasoning  was for automatically adding the `kubernetes` profile?

geoand:
@shahbour  @iocanel  PR #193 seems like something that would make profile handling feel a little more natural

iocanel:
@shahbour @geoand: Is this still relevant?

geoand:
I would say that since #176 was merged, this PR becomes obsolete

shahbour:
Ok cool, I think we can close this. I will update my code to and try #176 and if anything I will open issue.

PR #166, opened 2018-04-04
changed from .getLabels() to .getAnnotations()

doernbrackandre:


Salaboy:
+1 to this PR, using annotations instead of labels for service metadata is the right way to go. 
We will need to find a better way to deal with labels for service selections in kubernetes and make sure that the abstractions are in the Spring Cloud Common space.

Salaboy:
@spencergibb I will appreciate if you can merge this.

spencergibb:
I'm hesitant to because of 
>  and make sure that the abstractions are in the Spring Cloud Common space.

What do you mean?

doernbrackandre:
@Salaboy, @spencergibb  exactly. Annotations are what we know as metadata in spring-cloud, for example. Labels are part of the service discovery in k8s.

The goal should be to utilize the kubernetes client like this:
```
client.services()
    .withName(name)
    .withLabels(labels)
```
In this scenario one of the labels could be the stage in which the application is running (testing, staging, prod etc).

iocanel:
From my point of view it boils down, on how spring-cloud is using `metadata`.

If its used as supplement to the serviceId for identifying purposes, then we should stick to labels.
If not, then we probably need to switch to annotations.

Salaboy:
@iocanel that was exactly the case, my use case when I updated the module was to use labels to identify the services, but @doernbrackandre had a different use case using annotations. 
I do believe that we need to find out how to support both use cases and not sure that at the Spring Cloud Common level there is a distinction between the two. I do agree that for now, it looks more like annotations as stated by @doernbrackandre , that's why I gave this PR a +1

iocanel:
Maybe we should just make it configurable and support:

- labels
- annotations
- both (default option)

iocanel:
The good thing is that we agree on how labels and annotations are meant to be used.

The point where we disagree is that you are suggesting to dump labels (which hold identifying information) and just use annotations as metadata. 

This approach would be fine if we just needed to perform service lookup by serviceId, but what if you wanted to list all services that correspond to say 'revision 3'. Of course, such queries are not supported by the DiscoveryClient, but possibly listing all and filtering using metadata could be the only workaround. This ability will be lost if we merge.

On the other hand, we are loosing the annotations.

The fact that Kubernetes draws the line between labels and annotations by clarifying how they should be used limits the chances of conflicts (they can still happen, but its more of a user error rather than anything else). So personally I feel that we should merge labels and annotations as metadata.

@spencergibb: Your thoughts?

Salaboy:
@iocanel @spencergibb  totally agree.. I think that we need a new selector concept for dealing with things in the DiscoveryClient. That's what I meant by having different use cases before. I needed to use labels as selectors for picking up services not by name. 
Merging both annotations and labels might work for now, but I do believe that we will need a finer grained control to select services. 
I'm starting to think that we will need to label spring boot applications so the discovery client (let's say the one in the spring cloud gateway) can recognise from services that are spring cloud applications and for other services that might be floating around in the same namespace.

doernbrackandre:
Good morning guys,
> Merging both annotations and labels might work for now, but I do believe that we will need a finer grained control to select services.
I totally agree.

> I'm starting to think that we will need to label spring boot applications so the discovery client (let's say the one in the spring cloud gateway) can recognise from services that are spring cloud applications and for other services that might be floating around in the same namespace.
Exactly. This is the way we did it in our fork. We introduced labels to the DiscoveryClient to that it can utilize them a kind of a filter on discovering services. within a given "scope/range" (a namespace or all namespaces)

In k8s the object name is a metadata field. Is this maybe a valid approach to generalize the way how to handle  labels/annotations in the discovery client?

Salaboy:
@doernbrackandre @iocanel @spencergibb ok so new proposal.. so I can work on a PR for this. 
After considering several options, I would like to propose to extend the getServices() and getServiceInstance(String m) with some filtering capabilities, that can take into account Labels and Annotations. So the proposal can be split into two parts:
- Merge Labels and Annotations inside the serviceInstance.metadata
- Add new methods with lambda support: 
  - getServices(h -> Filter()),
  - getServiceInstances(String serviceName, h -> Filter())

 where the filter allow you to restrict annotations and labels. Notice that these methods should be added to the DiscoveryClient interface. 
Does this make sense to you guys? 
I can definitely spend some time this week on this.

Salaboy:
@doernbrackandre I'm with you on merging the annotations and labels.. I was thinking more in having a key for annotations and another key for labels inside the Metadata Map. We definitely need a more complex structure for that.

iocanel:
Since we all agree on that there is a mismatch between the ServiceInstance design and how Kubernetes works, but we don't have a solution that makes us all  happy, I'd like to propose the following:

Make everything configurable, so that users can tune it however they see fit. In detail:

- add flag to enable/disable addition of labels to metadta (default: true)
- add optional prefix to label keys (default none)
- add flag to enable/disable addition of annotation to metadata (default: true)
- add optional prefix to annotation keys (default none)

This way by default we have both. We can optional choose one or the other, but also we can prefix keys to avoid conflicts in case we the same key in both annotations and labels.

doernbrackandre:
Hi @iocanel,
> add optional prefix to label keys (default none)
> add optional prefix to annotation keys (default none)

what do you mean by this? Namespacing labels is nothing this library should do. The namespace is part of the name. This comes from the user/dev. Also, different labels can have different namespaces regarding to its purpose. It would only makes sense if all labels are within the same namespace. 

The same comes in place for annotations. From the k8s perspective, annotations don't need to be namespaced. However, it is up to the user/dev to add it to the name/key if needed.

Salaboy:
@doernbrackandre @spencergibb Spel Filters added here:  #227 , this should enable us to filter based on different criteria.

geoand:
FWIW I agree with the last proposal put forth by @iocanel and would gladly implement it we have a consensus on it

geoand:
> Since we all agree on that there is a mismatch between the ServiceInstance design and how Kubernetes works, but we don't have a solution that makes us all happy, I'd like to propose the following:
> 
> Make everything configurable, so that users can tune it however they see fit. In detail:
> 
>     * add flag to enable/disable addition of labels to metadta (default: true)
> 
>     * add optional prefix to label keys (default none)
> 
>     * add flag to enable/disable addition of annotation to metadata (default: true)
> 
>     * add optional prefix to annotation keys (default none)
> 
> 
> This way by default we have both. We can optional choose one or the other, but also we can prefix keys to avoid conflicts in case we the same key in both annotations and labels.

@salaboy @iocanel @spencergibb WDYT about the implementing this?

iocanel:
Sounds good

iocanel:
Created an issue to track this: #260.

PR #170, opened 2018-05-15
Relax the name application.(yaml|properties) constraint in ConfigMap

geoand:
When the ConfigMap contains a single entry, then the name of the file
does not have to application.(yaml|properties), but can be any
yaml/properties file.
This facilitates the use case where a confimap is created by something
like:

kubectl create configmap game-config --from-file=/path/app-config.yaml


iocanel:
@geoand: Can you please rebase?

geoand:
@iocanel, Done :)

PR #171, opened 2018-05-17
Add support for @ConditionalOnKubernetes and @ConditionalOnOpenshift

geoand:
These annotations could be used by application specific code that needs
to trigger bean creation only when a certain type of Kubernetes
distribution is being "targeted"


spring-cloud-kubernetes-core/src/main/java/org/springframework/cloud/kubernetes/condition/OnKubernetesOrOpenshiftCondition.java
```diff
@@ -0,0 +1,100 @@
...
+import io.fabric8.openshift.client.OpenshiftAdapterSupport;
+import java.util.Map;
+import org.springframework.boot.autoconfigure.condition.ConditionMessage;
+import org.springframework.boot.autoconfigure.condition.ConditionOutcome;
+import org.springframework.boot.autoconfigure.condition.SpringBootCondition;
+import org.springframework.context.annotation.ConditionContext;
+import org.springframework.core.env.PropertyResolver;
+import org.springframework.core.type.AnnotatedTypeMetadata;
+
+class OnKubernetesOrOpenshiftCondition extends SpringBootCondition {
```
iocanel: Again, I'd like to avoid having Openshift specific stuff in the core. They should go to a module of their own.

Having, said that I think that it might not be feasible to to really make sure if its vanilla. So we might need to check the behaviour to any working k8s environment will do. 

BTW, this is on me, cause I haven't thought of that detail, when we discussed about it. Sorry.

geoand: No problem at all :)

I will move the Openshift stuff to a separate module as you mention.
And I understand correctly, since there is no way of guaranteeing that a k8s cluster is a vanilla cluster, I should remove the `requireVanilla` option from `ConditionalOnKubernetes` (along with the respective tests of course)?

spencergibb:
I'm not sure there should be anything specific to openshift at all in this project

geoand:
I moved the Openshift specific parts to a separate module

spencergibb:
I think having components specific to a kubernetes implementation is a bad idea. I think it's a support burden we do not want to have. Even in a separate module.

Salaboy:
@geoand @iocanel same here.. I do believe that all these modules should be vanilla kubernetes, different distributions of kube can keep their extensions in their own repos.

iocanel:
I don't see why not having extension/favor specific modules here, given that there are people around to maintain them. If there are people willing to maintain them I feel its better to do that under the spirng-cloud-kubernetes banner (where things can be well aligned) than send them off to create their own projects, with their different conventions, styles, practices and varying quality. This will eventually harm this project too.

spring-cloud-kubernetes-core/src/main/java/org/springframework/cloud/kubernetes/KubernetesClientProperties.java
```diff
@@ -16,12 +16,14 @@
...
+import static org.springframework.cloud.kubernetes.Constants.SPRING_CLOUD_KUBERNETES_CLIENT_PROPS;
+
 import org.springframework.boot.context.properties.ConfigurationProperties;
 
-@ConfigurationProperties("spring.cloud.kubernetes.client")
+@ConfigurationProperties(SPRING_CLOUD_KUBERNETES_CLIENT_PROPS)
 public class KubernetesClientProperties {
 
-    private Boolean trustCerts;
+	private Boolean trustCerts;
```
spencergibb: spacing

spring-cloud-kubernetes-core/src/main/java/org/springframework/cloud/kubernetes/Constants.java
```diff
@@ -0,0 +1,9 @@
+package org.springframework.cloud.kubernetes;
+
+public final class Constants {
+
+	public static final String SPRING_CLOUD_KUBERNETES_CLIENT_PROPS = "spring.cloud.kubernetes.client";
```
spencergibb: No need for a separate class is there? I'd assume keep it in  the properties class.

spring-cloud-kubernetes-core/src/main/java/org/springframework/cloud/kubernetes/condition/OnKubernetesCondition.java
```diff
@@ -0,0 +1,39 @@
...
+	@Override
+	public ConditionOutcome getMatchOutcome(ConditionContext context,
+		AnnotatedTypeMetadata metadata) {
+
+		ConditionMessage.Builder message = ConditionMessage.forCondition("");
+
+		if (metadata.isAnnotated(ConditionalOnKubernetes.class.getName())) {
+			KubernetesClient kubernetesClient = getKubernetesClient(context.getEnvironment());
+
+			if (OnKubernetesConditionSupport.isServerUp(kubernetesClient)) {
```
spencergibb: Why the separate class here?

spring-cloud-kubernetes-core/src/main/java/org/springframework/cloud/kubernetes/condition/OnKubernetesConditionSupport.java
```diff
@@ -0,0 +1,50 @@
...
+				)
+			)
+			.build();
+
+		return new DefaultKubernetesClient(properties);
+	}
+
+	static boolean isServerUp(KubernetesClient kubernetesClient) {
+		try {
+			kubernetesClient.rootPaths();
```
spencergibb: Can you explain a bit more what's going on here? Generally, conditions evaluate something in the environment or on the classpath. I'm hesitant to have something that tries to execute against a remote api to determine the condition.

spring-cloud-kubernetes-core/src/main/java/org/springframework/cloud/kubernetes/KubernetesClientProperties.java
geoand: Good catch, fixed

spring-cloud-kubernetes-core/src/main/java/org/springframework/cloud/kubernetes/Constants.java
geoand: Fixed

spring-cloud-kubernetes-core/src/main/java/org/springframework/cloud/kubernetes/condition/OnKubernetesConditionSupport.java
geoand: This is needed because the whole point of `ConditionalOnKubernetes` is to ensure that a bean is registered when a Kubernetes server is accessible. 

The simplest scenario that I envision is that it can be used to differentiate the beans registered between environments that the code runs in.

Another reason that this is needed is that if this annotation is going to be extended to support matching on the server version (which I plan to do once the relevant functionality is added to `KubernetesClient`), then a connection to the server will be needed anyway.

spencergibb: I'm afraid I've never seen a boot condition used that way. Boot doesn't test a database before we create a datasource bean. This is also a potentially very bad for performance. I'd like to see if there's a different way to go about this.

spencergibb: On cloudfoundry, we key off of a spring profile named 'cloud'. We don't hit CF api endpoints.

spencergibb: That seems strange to me as well. @wilkinsona or @philwebb, thoughts on executing network operation on a conditional?

wilkinsona: I don’t have all the context, but I don’t like the sound of making a network call in a condition as it’ll hurt performance. Perhaps there’s something that can be down with environment variables instead? That’s how Boot’s [`CloudPlatform`](https://github.com/spring-projects/spring-boot/blob/d3c34ee3d1bfd3db4a98678c524e145ef9bca51c/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/cloud/CloudPlatform.java) is implemented

geoand:
Thank you very much for the input.

The reasoning behind the network call is to provide conditions based on
some data from the server (for example, the server version).

I totally understand the performance concern, which is why I suggested that
it could be controlled by a property - when the property is set, then the
regular functionality is enabled, otherwise it's not.

spring-cloud-kubernetes-core/src/main/java/org/springframework/cloud/kubernetes/condition/OnKubernetesConditionSupport.java
geoand: Thank you very much for the input. 

The reasoning behind the network call is to provide conditions based on some data from the server (for example, the server version). 

I totally understand the performance concern, which is why I suggested that it could be controlled by a property - when the property is set, then the regular functionality is enabled, otherwise it's not.

iocanel: I see two issues, with environment variables:

- not sure if existing environment variables cover our needs.
- at dev time the user will have to toy with those.

From my point of view its preferable to perform the actual call and provide the functionality in a transparent way at the cost of boot speed (as long as its well documented).

Note, that kubernetes is no database. In fact its unlike any other piece of software out there in that it has frequent releases that all introduce new features all the time. This increases the need to allow users to craft version specific config/beans etc.

geoand:
I have no removed Openshift from the production code and also introduced support for using the Server version in the Conditional.
This of course still assumes that a network call will be made during the evaluation of the Conditional as there is currently no other way to determine such a version from the environment

spring-cloud-kubernetes-core/src/main/java/org/springframework/cloud/kubernetes/condition/OnKubernetesCondition.java
```diff
@@ -0,0 +1,220 @@
...
+		}
+
+		return ConditionOutcome.noMatch(
+			message.because("Unknown Conditional annotation using " +
+				this.getClass().getSimpleName())
+		);
+	}
+
+	// We need to create a KubernetesClient ourselves
+	// since the KubernetesClient Bean is not available when the Condition is evaluated
```
wilkinsona: Unfortunately, I don't think this will work very well. The auto-configuration for `KubernetesClient` marks it as `@ConditionalOnMissingBean`. This is a good thing as it allows a user to provide their own `KubernetesClient` configured however they like. This condition effectively breaks that as it assumes that the user will have used the standard configuration properties to configure the client.

geoand: @wilkinsona That's very good point, thanks for bringing it up!

Unfortunately we have no other way of creating a KubernetesClient in the condition evaluation phase (which is executed well before the user specifies a custom KubernetesClient bean) that I can think of.

I understand that the implementation here is "unusual", in the sense that the output of the `Conditional` depends on some external service. Are you aware of any other Conditional that performs similar functionality?
Is this something that might perhaps make sense to be "standardized" in Spring Boot?

geoand:
I'm going to go ahead and close this, since we will try to come up with a solution that is less demanding on the client / library side

spencergibb:
Thanks, I tend to agree.

geoand:
@spencergibb If all goes well, we will add the `@ConditionalOnKubernetes` later on, but it will just read info from environment variables ( the problem is getting those environment variables into the pods ;) ).

spencergibb:
In the short term, Java system properties for the app would work as well (since they all end up normalized in the spring environment)

PR #175, opened 2018-05-22
Fix issue with ConfigMap with profiles inside and no active profiles

geoand:
Before this fix, when a ConfigMap contained profiles but no active profile was set, then no data was read from the ConfigMap at all


geoand:
@iocanel could you check this out? It's pretty simple and fixes a pretty annoying issue :)

geoand:
Thanks @iocanel!

PR #176, opened 2018-05-22
Allow the use of multiple configmaps as sources

geoand:
This is sort of a generalization of what has been done in #162.
It basically allows users to specify any number of ConfigMaps as property sources while following some sensible defaults

It doesn't introduce any specific profile support, but I am confident that it can be added without much hassle based on the infrastructure code I have put in place


geoand:
Thanks 😀!

I definitely agree that the "feature set" proposed by this PR can be
confusing!

My initial thought was to provide maximum flexibility, but that will
probably cause more pain than it's worth.

I will update the PR based on your suggestion very soon.

geoand:
@iocanel I updated the PR according to your suggestions.
Furthermore I rebased onto the latest `master` to make sure that everything still works as expected

Issue #131, opened 2017-09-20
KubernetesApplicationContextInitializer always tries to parse certificates


marcingrzejszczak:
when you have `spring-cloud-kubernetes` on the classpath and you're running integration tests that start up Boot context, then regardless of the fact whether you use Kubernetes or not, by default it tries to parse certificates from your local folders. That's because the `KubernetesApplicationContextInitializer` does this in the constructor. 

In my opinion either it should be evaulated lazilly or it should be possible to disable loading by a system property or sth (e.g. `spring.cloud.kubernetes.enabled=false`).

snowe2010:
This is a pretty big deal for us, as we want to disable kubernetes for production and just use it for development, but this makes it impossible, due to autoconfiguration ignoring all `spring.cloud.kubernetes.*` properties.

spencergibb:
PRs welcome

snowe2010:
What steps need to be taken to get this working? I would be glad to submit a PR, but am likely to break something due to the number of modules in this project.

snowe2010:
I'm not sure how the Initializer is called (because it implements `ApplicationContextInitializer`?).

snowe2010:
I seem unable to use `@ConditionalOnProperty(value = "spring.cloud.kubernetes.enabled", matchIfMissing = true)` because the properties have not been initialized by the point that the `spring.factories` are loaded (bootstrap). I'm not sure how to get around this.

dsyer:
The `Environment` is available in the `initialize()` method. If this feature needs to be implemented as an `ApplicationContextInitializer` (or similar), you would have to move  the logic to there.

geoand:
@marcingrzejszczak @snowe2010 Which certificates do you mean? I can submit a PR fixing the issue, but I'm not sure I completely understand the issue

spencergibb:
@geoand I think the logic in this line https://github.com/spring-cloud-incubator/spring-cloud-kubernetes/blob/master/spring-cloud-kubernetes-core/src/main/java/org/springframework/cloud/kubernetes/profile/KubernetesApplicationContextInitializer.java#L35 needs to move to the `initialize()`.

geoand:
@spencergibb That's my thought as well :), but I wanted to be sure before making any changes. 
That's why I would like @marcingrzejszczak or @snowe2010 to confirm which certificates are causing there issues.

snowe2010:
[...]

PR #177, opened 2018-05-22
Lazily create the KubernetesProfileApplicationListener

geoand:
This is needed in order to avoid causing problems to environments
that have spring-cloud-kubernetes on the classpath but don't want to
enable it

Fixes: #131


spring-cloud-kubernetes-core/src/main/java/org/springframework/cloud/kubernetes/profile/KubernetesApplicationContextInitializer.java
```diff
@@ -26,30 +27,38 @@
...
-    @Override
+	@Override
     public int getOrder() {
         return ORDER;
     }
 
     @Override
     public void initialize(ConfigurableApplicationContext applicationContext) {
-        listener.addKubernetesProfile(applicationContext.getEnvironment());
+    	if(isKubernetesEnabled(applicationContext)){
```
iocanel: Do we need that? Isn't lazy initialization enough?

geoand: Now that you mention it, that should be enough :).
I will update the PR

geoand: PR updated

PR #181, opened 2018-06-08
Leader election

gytis:
https://github.com/spring-cloud-incubator/spring-cloud-kubernetes/issues/172

I'm raising this pull request to start a review process while finishing up the implementation. I'm still planning to add integration tests, properties validation and more configuration options such as disabling failure events emitting.

@iocanel @nicolaferraro could you please give me your thoughts? @nicolaferraro I used your Camel implementation a lot here.


spring-cloud-kubernetes-leader/src/main/java/org/springframework/cloud/kubernetes/leader/KubernetesHelper.java
```diff
@@ -0,0 +1,106 @@
...
+	private final LeaderProperties leaderProperties;
+
+	private final KubernetesClient kubernetesClient;
+
+	public KubernetesHelper(LeaderProperties leaderProperties, KubernetesClient kubernetesClient) {
+		this.leaderProperties = leaderProperties;
+		this.kubernetesClient = kubernetesClient;
+	}
+
+	public boolean isPodAlive(String id) {
```
iocanel: This just checks if the pod exists, it doesn't do any status check. If its essential for the pod to be alive, I'd suggest we include the check. If not let's just change the name of the method to `podExists`.

spring-cloud-kubernetes-leader/src/main/java/org/springframework/cloud/kubernetes/leader/KubernetesHelper.java
```diff
@@ -0,0 +1,106 @@
...
+import io.fabric8.kubernetes.api.model.ConfigMap;
+import io.fabric8.kubernetes.api.model.ConfigMapBuilder;
+import io.fabric8.kubernetes.api.model.ObjectMeta;
+import io.fabric8.kubernetes.api.model.Pod;
+import io.fabric8.kubernetes.client.KubernetesClient;
+
+/**
+ * @author <a href="mailto:gytis@redhat.com">Gytis Trikleris</a>
+ */
+public class KubernetesHelper {
```
iocanel: Since, this is a class method that is specifc to the leader module and not a generic kubernetes helper, I'd prefer a name like 'LeaderHelper', 'LeaderUtils' (or whatever suffix is most commonly used around. :-)

spring-cloud-kubernetes-leader/src/main/java/org/springframework/cloud/kubernetes/leader/LeaderInitiator.java
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	private void update() {
+		if (leadershipController.acquire(candidate)) {
+			// We're a leader, check-in later
+			scheduleUpdate(leaderProperties.getLeaseDuration());
+		} else {
+			// Couldn't become a leader, retry sooner.
+			// TODO maybe we should separate error and another leader scenarios?
+			scheduleUpdate(leaderProperties.getRetryPeriod());
```
iocanel: Having  a fixed amount of time where we re-evaluate the situation works. But I think we can be awesome and issue a watcher for the actual leader AND the leaders configmap. If any of those resources changes we can call update. The watch thing can complement the scheduled update (which we can still keep as a safety net, just incase an event is missed or something).

spring-cloud-kubernetes-leader/src/main/java/org/springframework/cloud/kubernetes/leader/KubernetesHelper.java
gytis: It is possible for a pod to become a leader during the application boot-up, so having a status check might interfere with this. I'll rename to `podExists`

spring-cloud-kubernetes-leader/src/main/java/org/springframework/cloud/kubernetes/leader/KubernetesHelper.java
gytis: `LeaderKubernetesUtils`? `LeaderUtils` doesn't explain what type of the utils are defined in this class. All of them are specific to Kubernetes.

spring-cloud-kubernetes-leader/src/main/java/org/springframework/cloud/kubernetes/leader/LeaderInitiator.java
gytis: +1, should we implement this as a future improvement?

iocanel: Yeah, no need to do that now.

spring-cloud-kubernetes-leader/src/main/java/org/springframework/cloud/kubernetes/leader/KubernetesHelper.java
iocanel: +1

gytis:
@iocanel I've updated the PR based on your comments

PR #187, opened 2018-06-15
Discovery Client should only list Spring Boot App Services

Salaboy:
This Pull Request add a filter to the DiscoveryClient .getServices and .getServiceInstance to only retrieve services which contains a label: 
- spring-boot-app: "true" 

in their Kubernetes Service Descriptor. This has been tested with the Spring Boot Admin app (codecentric) and also with the spring cloud gateway (dynamic route discovery). 
This allow these implementations to only see spring boot apps and keeps the consistency provided by Eureka (PreKubernetes). We might want to extend the interface to provide another method which list all the services available in the kubernetes namespace. But for services that are not Spring Boot Apps I would rely on the Kubernetes Service catalog. 

I'm up for any discussion or concerns about this approach, but I do believe that we need this for consistency.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -179,7 +179,12 @@
 					<reuseForks>false</reuseForks>
 				</configuration>
 			</plugin>
-		</plugins>
+		
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-deploy-plugin</artifactId>
+        <version>2.8.2</version></plugin>
+    </plugins>
 		<pluginManagement>
 			<plugins>
 			</plugins>
--- a/pom.xml
+++ b/pom.xml
@@ -179,11 +179,6 @@
 					<reuseForks>false</reuseForks>
 				</configuration>
 			</plugin>
-		
-      <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-deploy-plugin</artifactId>
-        <version>2.8.2</version></plugin>
     </plugins>
 		<pluginManagement>
 			<plugins>
--- a/spring-cloud-kubernetes-discovery/src/main/java/org/springframework/cloud/kubernetes/discovery/KubernetesDiscoveryClient.java
+++ b/spring-cloud-kubernetes-discovery/src/main/java/org/springframework/cloud/kubernetes/discovery/KubernetesDiscoveryClient.java
@@ -39,13 +39,13 @@ public class KubernetesDiscoveryClient implements DiscoveryClient {
 
 	private static final Log log = LogFactory.getLog(KubernetesDiscoveryClient.class);
 	private static final String HOSTNAME = "HOSTNAME";
-	private static final String SPRING_BOOT_APP_LABEL = "spring-boot-app";
+
 
 	private KubernetesClient client;
 	private KubernetesDiscoveryProperties properties;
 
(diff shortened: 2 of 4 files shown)


cl4mer:
I think it is a good approach. For extending the API I was thinking about a method which receives a lambda expression in order to allow custom filtering. What do you think?

Salaboy:
@cl4mer +1 I will provide an extra method where you can send a selector object to filter services by metadata in a different PR.

spencergibb:
If this is specifically for gateway, I've got dynamic filtering via SpEL. If we want filtering generically here, I'd go similar to what I have in gateway (ie. the key is configurable).

Salaboy:
@spencergibb done. This is not specific for the gateway, but please share the dynamic filtering mechanism if possible, so I can check it out to see if we can apply something like that for discovery.

Salaboy:
@spencergibb it should be all green now. I've added the property so the key can be configured, I'm accepting suggestions for the name of that key, but the functionality is working correctly.

Salaboy:
@spencergibb @iocanel if there is anything else that you want to see in this PR please let me know. The admin app and gateway are working as expected with this PR code, so I would like to get this merged asap to be able to merge some other PRs in other projects.

sta-szek:
I would say it is good to discover all services. Otherwise, shall I rewrite everything to java to use this autodiscovery?

Salaboy:
@sta-szek not really, if you want to discover a service you just need to add a label to it.  
In Spring Cloud when you do discovery, you are assuming that those other services are Spring Boot applications and that they will expose certain things such as actuators to monitor them. 
Notice that this is replicating the same behaviour of Spring Cloud with Eureka.
But if you want to discover other services and deal with them in another way, you are free to label them accordingly and the DiscoveryClient will pick them up.

iocanel:
It's possible that I am missing something obvious here, so please do correct me if I am wrong.

My main issue, is that I fail to see why would we want to require additonal labels to filter out services. This seems like a tailor made extra step, that would become a burden for most use cases.

I could possibly agree, to an **Optional** and more versatile version of this. What @spencergibb proposed about SpEL sounds good to me.

sta-szek:
From my side:
1. By default discover all services
2. If you want to discover only spring-boot apps, then add property `spring.cloud.kubernetes.discovery.spring-boot-only: true` -- by default false
3. Then add possibility to discover other services which has some annotation like `spring-boot-discover: true`

Salaboy:
@iocanel @sta-szek good discussion guys, and that was the main intention of this PR, to make people aware of what the problem is. 

@iocanel I think that we can automate the setting of the label. Basically, when the spring app is started, we can call the k8s API behind the scenes to add that label to the corresponding services. SpEL sounds like something that we can do the gateway, but it is not general. Do you mean adding an SpEL filter into the DiscoveryClients methods? 

@sta-szek I kinda agree with that approach of having a flag, but it will be confusing anyway. That was the previous default behaviour for Spring Cloud Discovery Client. Imagine that all the applications that were written based on that will be broken in one way or another. If we don't care about those application's assumptions, I'm ok to make it false by default.

Salaboy:
@spencergibb @iocanel @sta-szek @paolocarta 
Closing this one in favor of this new one: https://github.com/spring-cloud/spring-cloud-kubernetes/pull/227

PR #191, opened 2018-06-19
Enable/disable failure events publishing during leader election

gytis:
Adding a property `spring.cloud.kubernetes.leader.publishFailedEvents` which allows to enable/disable event publishing when leader acquisition fails. It is set to `false` by default because this event is very chatty. Acquisition failure is a normal event while there is a current leader. Other members of the cluster constantly try to take over membership in case if the current leader has disappeared.


gytis:
@iocanel ping

PR #195, opened 2018-06-23
refactor: merge spring-cloud-stater-kubernetes-config into spring-cloud-starter-kubernetes.

iocanel:


spencergibb:
Why are you doing this? It is a common pattern across many spring cloud projects. I don't want to change it here.

PR #200, opened 2018-07-01
Option to make health check fail when the Kubernetes API isn't available

will-gtv:
Adds an option (`spring.cloud.kubernetes.health.failOutsidePod`) that
causes the health indicator to report as down when Spring hasn't
communicated with the Kubernetes API correctly.

We have had occasional issues with pods that start up and report as healthy but aren't actually picking things up from the Kubernetes API servers. Applying this setting through a Pod template & environment variable should let k8s fix the problem for us and prevent traffic from ever reaching them.

### DIFF
--- a/spring-cloud-kubernetes-core/src/main/java/org/springframework/cloud/kubernetes/KubernetesAutoConfiguration.java
+++ b/spring-cloud-kubernetes-core/src/main/java/org/springframework/cloud/kubernetes/KubernetesAutoConfiguration.java
@@ -88,11 +88,13 @@ public class KubernetesAutoConfiguration {
 
     @Configuration
     @ConditionalOnClass(HealthIndicator.class)
+	@EnableConfigurationProperties(KubernetesHealthIndicatorProperties.class)
     protected static class KubernetesActuatorConfiguration {
         @Bean
         @ConditionalOnMissingBean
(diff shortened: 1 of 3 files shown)


pivotal-issuemaster:
@will-gtv Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud-incubator/spring-cloud-kubernetes&pullRequestId=200)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud-incubator/spring-cloud-kubernetes&pullRequestId=200) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/about) for frequently asked questions.

Salaboy:
@will-gtv can you elaborate a little bit more about the problem that you were facing? In Kubernetes until the pod is not healthy the service is not enabled, meaning that other services shouldn't be able to communicate with it. Also the Pod itself can use the kubernetes APIs to find other services, but that will only happen when the pod is up and healthy right? 

I am trying to understand when this new property will help .

geoand:
@will-gtv I am also interested to hear more about your issue

spencergibb:
Closing due to lack of requested feedback. If you would like us to look at this issue, please provide the requested information and we will re-open the issue.

PR #204, opened 2018-07-02
Optimise leader pod check

gytis:
@iocanel @steven-sheehy

### DIFF
--- a/spring-cloud-kubernetes-leader/src/main/java/org/springframework/cloud/kubernetes/leader/LeaderProperties.java
+++ b/spring-cloud-kubernetes-leader/src/main/java/org/springframework/cloud/kubernetes/leader/LeaderProperties.java
@@ -24,6 +24,8 @@ import org.springframework.boot.context.properties.ConfigurationProperties;
 @ConfigurationProperties("spring.cloud.kubernetes.leader")
 public class LeaderProperties {
 
+	private static final boolean DEFAULT_ENABLED = true;
+
 	private static final String DEFAULT_LEADER_ID_PREFIX = "leader.id.";
 
 	private static final boolean DEFAULT_AUTO_STARTUP = true;
@@ -35,6 +37,14 @@ public class LeaderProperties {
 	private static final long DEFAULT_RETRY_PERIOD = 5000;
 
 	private static final double DEFAULT_JITTER_FACTOR = 1.2;
+
+	private static final boolean DEFAULT_PUBLISH_FAILED_EVENTS = false;
+
+	/**
+	 * Should leader election be enabled.
+	 * Default: true
+	 */
+	private boolean enabled = DEFAULT_ENABLED;
 
 	/**
 	 * Should leader election be started automatically on startup.
@@ -86,7 +96,15 @@ public class LeaderProperties {
 	 * Enable/disable publishing events in case leadership acquisition fails.
 	 * Default: false
 	 */
-	private boolean publishFailedEvents = false;
+	private boolean publishFailedEvents = DEFAULT_PUBLISH_FAILED_EVENTS;
+
+	public boolean isEnabled() {
+		return enabled;
+	}
+
+	public void setEnabled(boolean enabled) {
+		this.enabled = enabled;
+	}
 
 	public boolean isAutoStartup() {
 		return autoStartup;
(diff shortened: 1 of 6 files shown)


spring-cloud-kubernetes-leader/src/main/java/org/springframework/cloud/kubernetes/leader/LeaderProperties.java
```diff
@@ -61,12 +58,6 @@
 	 */
 	private String configMapName = DEFAULT_CONFIG_MAP_NAME;
 
-	/**
-	 * Kubernetes labels common to all leadership candidates.
-	 * Default: empty
-	 */
-	private Map<String, String> labels = Collections.emptyMap();
-
```
steven-sheehy: Can you add enabled property that is checked in LeaderAutoConfiguration?

gytis: It is already here https://github.com/spring-cloud-incubator/spring-cloud-kubernetes/blob/master/spring-cloud-kubernetes-leader/src/main/java/org/springframework/cloud/kubernetes/leader/LeaderProperties.java#L46

steven-sheehy: You're referring to autoStartup. I'm referring to "spring.cloud.kubernetes.leader.enabled", which is missing from LeaderProperties but checked in this line `@ConditionalOnProperty(value = "spring.cloud.kubernetes.leader.enabled", matchIfMissing = true)`

gytis: Done. Although, it's not strictly necessary.

gytis:
Unrelated Ribbon tests have failed.

gytis:
Retest this please

steven-sheehy:
I know above is for circleci, but I also tested this change and think it may pose problems. If I start two pods at the exact same time then both will transition to leader at the same time. The problem corrects itself at the next poll period, but we should never have more than one leader at a time. I think this occurs due to the ready check since both pods are not ready so they both acquire.

If we want to keep ready check, then logic should ensure leader.isValid() returns true if he is ready OR **if he is not ready but there are no other ready pods**.

Also, it was pretty difficult to diagnose this since there is very little logging. Anyway some more debug statements could be added like when attempting acquire/revoke, updating configmap, etc.?

gytis:
Were you able to reproduce the issue with two leaders? Did you already have a configmap created?

steven-sheehy:
Yes, it occurs consistently as long as your pod's readinessProbe.initialDelaySeconds is high enough. The configmap was already created at the time.

steven-sheehy:
1) Just to expand upon the problem of checking only pod exists: If you have two ready pods and one restarts, it will continue to keep the same pod name and most likely keep the leader lock. But that pod is not really able to respond to requests yet and it would be much better to elect the pod that is currently up and ready instead.

2) "Only allow pod to become a leader if it is ready." This is what your PR currently does. I think it needs to additionally check if not ready, then proceed if no other ready pods.

Maybe we can sidestep the issue and instead of checking ready status (which is mostly applicable to service traffic), it would be better to look at pod.status.phase == 'Running'. Main thing is to ensure pod not elected leader if he's still starting up, not that he became unready after startup. I don't think pod phase can ever go back to another status without a restart. Kubernetes docs say that Running means:

> The pod has been bound to a node, and all of the containers have been created. At least one container is still running, or is in the process of starting or restarting.

steven-sheehy:
Nevermind about the pod.status.phase. That is marked as running even when not ready. So guess we still need to figure out how to use ready.

iocanel:
retest this please

iocanel:
IMHO, we should check if a pod is ready. The client has a utility called `Readiness` that can easily perform the check.

The question is what happens when a pod that was once ready, fails its readiness check....
I think that all candidates should watch for pod events and if the leader is no longer ready, should automatically kick an election.

geoand:
Should the PR be rebased onto the latest master (or have master merged into it) in order to take advantage of the fix to the Ribbon test?

iocanel:
@gytis can u please rebase?

gytis:
OK so I'll have to reimplement the logic how leadership is given and revoked. Will also rebased at the same time.

PR #209, opened 2018-07-05
KubernetesCatalogWatch endpoint based

akaGelo:
Fix problem [https://github.com/spring-cloud-incubator/spring-cloud-kubernetes/issues/150#issuecomment-402499289](https://github.com/spring-cloud-incubator/spring-cloud-kubernetes/issues/150#issuecomment-402499289)

---

org.springframework.cloud.client.ServiceInstance  Instance is a Pod and not Service.

Pod name is [unique](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/)

---
P.S. Watch  resourceVersion  of EndpointsList not effective. Many false positives that do not affect the list of ServiceInstance.

### DIFF
--- a/spring-cloud-kubernetes-discovery/src/main/java/org/springframework/cloud/kubernetes/discovery/KubernetesCatalogWatch.java
+++ b/spring-cloud-kubernetes-discovery/src/main/java/org/springframework/cloud/kubernetes/discovery/KubernetesCatalogWatch.java
@@ -16,6 +16,8 @@
 
 package org.springframework.cloud.kubernetes.discovery;
 
+import io.fabric8.kubernetes.api.model.Endpoints;
+import io.fabric8.kubernetes.client.KubernetesClient;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.cloud.client.discovery.event.HeartbeatEvent;
@@ -25,6 +27,7 @@ import org.springframework.scheduling.annotation.Scheduled;
 
 import java.util.List;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.stream.Collectors;
 
 
 /**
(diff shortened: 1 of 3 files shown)


spring-cloud-kubernetes-discovery/src/main/java/org/springframework/cloud/kubernetes/discovery/KubernetesCatalogWatch.java
```diff
@@ -50,19 +54,27 @@ public void setApplicationEventPublisher(ApplicationEventPublisher publisher) {
 	@Scheduled(fixedDelayString = "${spring.cloud.kubernetes.discovery.catalogServicesWatchDelay:30000}")
 	public void catalogServicesWatch() {
 		try {
-			List<String> previousState = catalogServicesState.get();
```
iocanel: I am wondering if it make sense to have @Scheduled operation. It might be better if we just use a watch. Not sure if its even possible, I am just thinking out loud.

spring-cloud-kubernetes-discovery/src/test/java/org/springframework/cloud/kubernetes/discovery/KubernetesCatalogWatchTest.java
```diff
@@ -23,11 +33,17 @@
...
 
 	@Mock
-	private KubernetesDiscoveryClient kubernetesDiscoveryClient;
+	private KubernetesClient kubernetesClient;
 
 	@Mock
 	private ApplicationEventPublisher applicationEventPublisher;
 
+	@Mock
+	private MixedOperation<Endpoints, EndpointsList, DoneableEndpoints, Resource<Endpoints, DoneableEndpoints>> endpointsOperation;
```
iocanel: It's also possible to mock Kubernetes Api Server itself.

spring-cloud-kubernetes-discovery/src/main/java/org/springframework/cloud/kubernetes/discovery/KubernetesCatalogWatch.java
akaGelo: I considered the use  "kubernetesClient.endpoints().watch()" but then the websocket will be used. 
I can implement it, but I'm not competent in the discussion.


@Scheduled  was previously used in the [consul](https://github.com/spring-cloud/spring-cloud-consul/blob/8d43d78959a262efb6c0c68389760beddf62e9c0/spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulCatalogWatch.java#L56)


Watching use in [Zookeeper](https://github.com/spring-cloud/spring-cloud-zookeeper/blob/6ba78fd833dd63dfa6c7015edca42d66b27805ab/spring-cloud-zookeeper-discovery/src/main/java/org/springframework/cloud/zookeeper/discovery/ZookeeperServiceWatch.java#L67)


In [cloud-config](https://github.com/spring-cloud-incubator/spring-cloud-kubernetes/blob/3ad6fa14e14daa4d6b49d81e4662e5ccbf74a916/spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/reload/ConfigReloadAutoConfiguration.java#L79) there is a choice between implementations. 
Is it necessary to do the same here?

iocanel: Great point!

I've seen people struggling with websockets when they use the kube proxy. Making it configurable would be awesome. It might make sense to have a watch feature toggle that we enable disable globaly.

spencergibb: I removed @scheduled

PR #219, opened 2018-07-25
Refactor leader election to use watchers

gytis:
After discussions with @iocanel and @steven-sheehy I've decided to refactor the leader election implementation to have a more thorough checking on pod readiness and also make use of watchers provided by KubernetesClient.

First, change is that pod no longer can become a leader which its status is not ready.
Another change is that there were multiple actions added that can trigger leader information reload and change as opposed to only scheduled update as it was before:

- Periodic update scheduled by LeaderInitiator in case there was a missed event or other problem
- Add, update or delete action on configmap received by LeaderRecordWatcher
- Readiness status change received by any instance of PodReadinessWatcher. Normally there are one or two instances running at a time. Always an instance monitoring readiness of a host pod. And optionally an instance monitoring a leader pod, if host is not a leader.

I have tested this manually and also added unit tests for the smaller classes. I didn't want to waste too much time writing unit tests for the biggest class or integration tests before getting some feedback on the change.

@iocanel, @steven-sheehy, would you mind providing your feedback?


gytis:
FYI this PR makes https://github.com/spring-cloud-incubator/spring-cloud-kubernetes/pull/204 irrelevant.

steven-sheehy:
> First, change is that pod no longer can become a leader which its status is not ready.

What about when there are no ready pods? It should still choose a random leader so it doesn't wait a minute or two for a pod to become ready.

gytis:
I don't think it is correct to assign leadership to a random non-ready pod.
Firstly, because if the pod is not ready it means it is not yet functioning properly and is not ready to accept requests. If your pods are not booting up it must be an issue and just assigning a leader to a random one wouldn't solve the issue.
Secondly, it will increase a possibility that the leadership will be assigned and immediately revoked when another pod becomes ready faster.
And overall this behaviour would confusing and unpredictable.

spring-cloud-kubernetes-leader/src/main/java/org/springframework/cloud/kubernetes/leader/LeaderRecordWatcher.java
```diff
@@ -0,0 +1,86 @@
...
+
+	private static final Logger LOGGER = LoggerFactory.getLogger(LeaderRecordWatcher.class);
+
+	private final LeadershipController leadershipController;
+
+	private final LeaderProperties leaderProperties;
+
+	private final KubernetesClient kubernetesClient;
+
+	private Watch watch;
```
iocanel: i think that this is only written by `start` and `onClose`. With reasonable use we should be fine regarding thread safety. Still it would be nice to either comment or guard.

iocanel: yeah. thats what i meant as guard.

gytis: @iocanel I've added the locks around the status update blocks in watchers

PR #227, opened 2018-08-24
Initial Service Filtering using Spel Expression

Salaboy:
This PR includes an initial filter implementation using an SPEL expression.
You can add a filter for your services by setting the property:
spring.cloud.kubernetes.discovery.filter = metadata.additionalProperties['spring-boot']

### DIFF
--- a/spring-cloud-kubernetes-discovery/src/main/java/org/springframework/cloud/kubernetes/discovery/KubernetesDiscoveryClient.java
+++ b/spring-cloud-kubernetes-discovery/src/main/java/org/springframework/cloud/kubernetes/discovery/KubernetesDiscoveryClient.java
@@ -145,12 +145,12 @@ public class KubernetesDiscoveryClient implements DiscoveryClient {
 
     @Override
     public List<String> getServices() {
-        String spelExpresion = properties.getFilter();
+        String spelExpression = properties.getFilter();
         Predicate<Service> filteredServices;
-        if (spelExpresion == null || spelExpresion.isEmpty()) {
+        if (spelExpression == null || spelExpression.isEmpty()) {
             filteredServices = (Service instance) -> true;
         } else {
-            Expression filterExpr = parser.parseExpression(spelExpresion);
+            Expression filterExpr = parser.parseExpression(spelExpression);
             filteredServices = (Service instance) -> {
                 Boolean include = filterExpr.getValue(evalCtxt, instance, Boolean.class);
                 if (include == null) {
--- a/spring-cloud-kubernetes-discovery/src/main/java/org/springframework/cloud/kubernetes/discovery/KubernetesDiscoveryClient.java
+++ b/spring-cloud-kubernetes-discovery/src/main/java/org/springframework/cloud/kubernetes/discovery/KubernetesDiscoveryClient.java
@@ -29,12 +29,21 @@ import io.fabric8.kubernetes.api.model.Endpoints;
 import io.fabric8.kubernetes.api.model.Service;
 import io.fabric8.kubernetes.client.KubernetesClient;
 import io.fabric8.kubernetes.client.utils.Utils;
+import java.util.function.Predicate;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.cloud.client.DefaultServiceInstance;
 import org.springframework.cloud.client.ServiceInstance;
 import org.springframework.cloud.client.discovery.DiscoveryClient;
+import org.springframework.context.ApplicationContext;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.expression.BeanFactoryResolver;
 import org.springframework.util.Assert;
+import org.springframework.expression.Expression;
+import org.springframework.expression.spel.standard.SpelExpressionParser;
+import org.springframework.expression.spel.support.SimpleEvaluationContext;
+import org.springframework.expression.spel.support.StandardEvaluationContext;
 
 public class KubernetesDiscoveryClient implements DiscoveryClient {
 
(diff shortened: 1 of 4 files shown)


spring-cloud-kubernetes-discovery/src/main/java/org/springframework/cloud/kubernetes/discovery/KubernetesDiscoveryClient.java
```diff
@@ -134,12 +147,42 @@ public ServiceInstance getLocalServiceInstance() {
...
+        
 	@Override
 	public List<String> getServices() {
-		return client.services().list()
-			.getItems()
-			.stream().map(s -> s.getMetadata().getName())
-			.collect(Collectors.toList());
+                String spelExpresion = properties.getFilter();
+		if(spelExpresion == null || spelExpresion.isEmpty()){
+                    spelExpresion = "true";
```
spencergibb: I'd prefer to just have the original if there is no expression, no reason to add the overhead of spel if there is none.

```
return client.services().list()
	.getItems()
.stream().map(s -> s.getMetadata().getName())
```

spring-cloud-kubernetes-discovery/src/main/java/org/springframework/cloud/kubernetes/discovery/KubernetesDiscoveryClient.java
```diff
@@ -134,12 +147,42 @@ public ServiceInstance getLocalServiceInstance() {
...
+		return getServices(includeExpr);
 	}
+        
+        public List<String> getServices(String spelFilterExpression) {
+                Expression includeExpr = parser.parseExpression(spelFilterExpression);
+		return getServices(includeExpr);
+	}
+
+        private List<String> getServices(Expression filterExpr) {
+            SimpleEvaluationContext evalCtxt = SimpleEvaluationContext
```
spencergibb: I wonder if this can be a reused as a field?

spring-cloud-kubernetes-discovery/src/main/java/org/springframework/cloud/kubernetes/discovery/KubernetesDiscoveryClient.java
spencergibb: Or instead of passing a spel expression, pass the predicate, one that includes spel and one that just returns true.

Salaboy:
@spencergibb all comments done!, sorry about the formatting, let me know if there is someone else to change.

spring-cloud-kubernetes-discovery/src/main/java/org/springframework/cloud/kubernetes/discovery/KubernetesDiscoveryClient.java
```diff
@@ -14,132 +14,161 @@
...
+            }
+            return defaultInstance;
+
+        } catch (Throwable t) {
+            return defaultInstance;
+        }
+    }
+
+    @Override
+    public List<ServiceInstance> getInstances(String serviceId) {
```
paolocarta: Hi guys, having a look at this method and getLocalServiceInstance(), I think the logic is pretty similar. Maybe it would be nice reusing the common logic. 

If you want, I could do that in a follow up PR.

spring-cloud-kubernetes-discovery/src/main/java/org/springframework/cloud/kubernetes/discovery/KubernetesDiscoveryClient.java
```diff
@@ -14,132 +14,161 @@
...
+
+    @Override
+    public List<String> getServices() {
+        String spelExpresion = properties.getFilter();
+        Predicate<Service> filteredServices;
+        if (spelExpresion == null || spelExpresion.isEmpty()) {
+            filteredServices = (Service instance) -> true;
+        } else {
+            Expression filterExpr = parser.parseExpression(spelExpresion);
+            filteredServices = (Service instance) -> {
```
paolocarta: Maybe you could extract this lambda expression for testing it in isolation, what do you think?

spring-cloud-kubernetes-discovery/src/main/java/org/springframework/cloud/kubernetes/discovery/KubernetesDiscoveryClient.java
```diff
@@ -14,132 +14,161 @@
...
+        }
+
+        return instances;
+    }
+
+    @Override
+    public List<String> getServices() {
+        String spelExpresion = properties.getFilter();
+        Predicate<Service> filteredServices;
+        if (spelExpresion == null || spelExpresion.isEmpty()) {
```
paolocarta: typo: spelExpression

spring-cloud-kubernetes-discovery/src/main/java/org/springframework/cloud/kubernetes/discovery/KubernetesDiscoveryClient.java
Salaboy: @paolocarta I was thinking that we can remove it completely. I cannot understand why we need that besides some kind of edge case in minikube.

spring-cloud-kubernetes-discovery/src/main/java/org/springframework/cloud/kubernetes/discovery/KubernetesDiscoveryClient.java
Salaboy: @paolocarta typo fixed. Not sure about StringUtils, is that used all over spring? I wouldn't want to introduce something that is not widely used in the framework.

spring-cloud-kubernetes-discovery/src/main/java/org/springframework/cloud/kubernetes/discovery/KubernetesDiscoveryClient.java
Salaboy: @paolocarta it is a good improvement, but I would like to do it after this PR is merged if possible.

spencergibb:
Squashed and merged with some polish via a12eef989af9de300634bab0af1d5b9c926d2f2b

Issue #230, opened 2018-09-11
Cannot load multiple mounted configmap with multiple application.yaml


au-hao:
I would like to use multiple mounted config maps and according to the source code `KubernetesPropertySource.java`, same key `application.yaml` will be overwritten in the result `HashMap` which causes that multiple config maps cannot be loaded at the same time and only the last mounted configmap that has the same key can be loaded successfully.  Need a help since I am not sure if this is the design or it is a bug. Also I tried consuming configmap from API which works.

Configuration files:
```yaml
# configmap
apiVersion: v1
kind: ConfigMap
metadata:
  name: configmap-spring-1
  namespace: {{ .Release.Namespace }}
data:
  application.yaml: |-
    deployment:
      type: cloud
      name: beta
```
```yaml
# configmap
apiVersion: v1
kind: ConfigMap
metadata:
  name: configmap-spring-2
  namespace: {{ .Release.Namespace }}
data:
  application.yaml: |-
{{ .Files.Get "application.yaml" | indent 4 }}
    ---
    spring.profiles: {{ .Values.springProfile }}
{{ .Files.Get (printf "application-%s.yaml" .Values.springProfile) | indent 4 }}
```
```yaml
# values.yaml
springProperties:
  spring:
    cloud:
      kubernetes:
        config:
          enableApi: false
          paths:
            - "/home/configmaps"
        reload:
          enabled: true
```

geoand:
I am not absolutely sure, but I think this is a result of Kubernetes and there isn't anything Spring Cloud Kubernetes can do.
This can be confirmed if exec into the pod and do `ls -l /home/configmap`. I am willing to bet you will only see one of the files.

au-hao:
@goand Yeap, I did check that those configmap are in the pod and the problem is that those two sample configmap all have application.yaml files.

geoand:
So can you confirm that `ls -l /home/configmap` inside the the Pod only contains a single file?
If so, then this is not an issue that SCK can deal with

au-hao:
@geoand those configmap are there.

geoand:
Can you please paste the output of the 2 files? Thanks

geoand:
I mean as they show up in the file system of the pod, not the configmap resource itself :)

au-hao:
![image](https://user-images.githubusercontent.com/40636256/45365850-70b8e280-b5ab-11e8-8827-0d47346f1fd4.png)

[...]

PR #244, opened 2018-10-08
Rationalize ConfigMap usage when paths (mounted files) are involved

geoand:
This is done by removing the file walking and loading logic that existed
before and replacing it with an implementation that creates on
ConfigMapResource per specified path.
Reading and processing the content of the path is done in the exact
same way as the enableApi method does thus providing a consistent
experience.

The implementation is inspired by the following comment:
https://github.com/spring-cloud/spring-cloud-kubernetes/issues/230#issuecomment-420341743

Fixes: #230

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -224,6 +224,12 @@ spec:
 - check the security configuration section, to access config maps from inside a pod you need to have the correct
 Kubernetes service accounts, roles and role bindings.
 
+Another option for using ConfigMaps, is to mount them into the Pod running the Spring Cloud Kubernetes application
+and have Spring Cloud Kubernetes read them from the file system.
+This behavior is controlled by the `spring.cloud.kubernetes.config.paths` property and can be used in
+addition to or instead of the mechanism described earlier.
+Multiple (exact) file paths can be specified in `spring.cloud.kubernetes.config.paths` by using the `,` delimiter
+
 .Properties:
 [options="header,footer"]
 |===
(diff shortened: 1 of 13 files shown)


geoand:
@au-hao @william-tran @spencergibb @iocanel Could you please give this a review and let me know what you think?

I also believe that this same method should be extended the handling `Secrets` and I can fix them up as well if you think this approach is viable

geoand:
The more I look at how [Secrets](https://kubernetes.io/docs/concepts/configuration/secret/) are meant to be handled in Kubernetes, the more I think that Secret in handling in SCK should not be the same as that of ConfigMaps.

The reasoning is that ConfigMaps can contain entire `properties` or `yaml` files which should then be "flattened" into key/value properties.
On the other hand, although Secrets can contain values from files, those values are not meant to be  structured values, but instead they are encoded string values. In my mind, that means that the current handling of Secrets as a single PropertySource in SCK makes perfect sense.

So I think that this PR is the correct way to go for ConfigMaps while the current implementation of Secrets should be maintaned (perhaps with a little polish)

william-tran:
As long as the precedence of secrets loaded via `spring.cloud.kubernetes.secrets.paths` is consistent with `spring.cloud.kubernetes.config.paths`, where it seems last one wins, I'm ok with this. Because secrets are used as a flat key/value map much like each entry in a .properties file, its a lot simpler to deal with merging multiple ones with a predictable order of precedence.

geoand:
Can I get some feedback please on this approach?

If valid we can discuss then how to deal with secrets

salaboy:
@geoand @spencergibb this looks good to me but I would like to talk with @geoand to understand a little bit more the implementation.

geoand:
@salaboy I won't be able to make tomorrow's call since I will be at a local meetup. We can chat however whenever works for you :)

geoand:
@iocanel PR is rebased and should now be ready for review

PR #256, opened 2018-10-24
Introduce integration tests

geoand:
Opening PR so I can get feedback from CircleCI


geoand:
Launching a k8s cluster via microk8s is now working. So it's just a matter of creating the integration tests now.

One more thing we need to figure out is whether the spring applications _need_ to run in the cluster or not.
Having the applications run outside the cluster will make things a lot simpler and faster

geoand:
I have made progress with this. I have created a [prototype](https://github.com/geoand/sb-fmp-microk8s/blob/sck/pom.xml) application that uses the Fabric8 Maven Plugin in order to deploy an SCK application into a Kubernetes cluster setup using microk8s.
What remains now is to write integration tests (probably leveraging Arquillian Cube) and setup some scripts that will deploy each test application, test it and undeploy it

geoand:
The prototype specified above now includes an Arquillian Cube integration test. So what needs to be done now is for me to port the setup to this repo, add a couple integration tests, create a few bash scripts that can launch all the integration tests and finally update the CircleCI configuration

geoand:
PR is now complete

ryanjbaxter:
Its not clear to me what will happen if I dont have microk8s installed and running and I run the build.  I built your branch and the integration test modules built fine, so what happened there?

geoand:
@ryanjbaxter A typical Maven build wont result in the integration tests being executed. 

You need to run the build with the `it` and `fmp` profiles.

If you do that and microk8s isn't installed, the build should fail, the first error probably being the inability to push a docker image

geoand:
@ryanjbaxter I don't such a thing would be possible. 
microk8s launches a regular k8s cluster so I don't see how microk8s (or any other solution short from a VM - which is what minikube does by default) could achieve having everything ready made in a docker image

ryanjbaxter:
Just a shame that we wont be able to run them outside of using Linux

geoand:
Indeed. 

If you have a Kubernetes cluster, then the tests can be configured to run there. 
The tests themselves aren't constrained to microk8s, but the project has been setup in a why to make running them on microk8s the default option

salaboy:
@ryanjbaxter I do feel the same way being on Mac, but I do see the potential to have the tests running as part of the CI pipelines against a real instance of the Kubernetes APIs. You can always set up a personal travis CI account to work on changes on top of your forks.

iocanel:
@geoand: To address @ryanjbaxter concerns I would suggest to leverage arquillian cube features for setting up the environment and just use circleci for setting up mircok8s. What we could achieve this way is that developers could run integration tests against their target k8s/openshift environment without having to install microk8s.

geoand:
@iocanel Than can already be done easily, it's just not documented :)

All that needs to be done is to set a few maven properties and everything should work as expected

geoand:
I'll try spare some time to update the PR along the lines you specified @iocanel , but I can't promise fast turnover unfortunately :(

iocanel:
@geoand: You are not missing a thing. The only benefit you'll get if you give more control to arquillian-cube, is that you won't require external shell scripts for controlling the test environment.

iocanel:
And of course its clear that the shell scripts doesn't need to bind you to the underlying cluster and could just assume kubectl. Still if we eliminated them it would make it easier for a user to run the whole thing using something like: `mvn clean fabric8:build integration-test`. 

An alternative woulbe to separate setting up microk8s from orchestrating the test and move orchestrating scripts under the integration tests.

geoand:
@iocanel Thanks for the tips! I have some catching up to do, but I'll try to look into this soon

geoand:
@iocanel I have removed the custom bash scripts and the lifecycle of the tests is now handled by Arquillian Cube.
Also I have removed the microk8s defaults from `pom.xml` and moved them as cli args to the CircleCI configuration. 
It should also be clear from documentation which piece does what

geoand:
Thanks!!!

PR #257, opened 2018-10-26
Fix pvc template in README 

tanshanshan:
Fix pvc template in README


tanshanshan:
I think I have done it both in minikube and Minishift @spencergibb

Issue #260, opened 2018-10-30
Make the source of Service Metadata configureable


iocanel:
Currently service labels are used. There are people who would rather use annotations instead. See #166.

After a long discussion we have decided to make it configurable. 
Since there is not recent activity in the pull request, I will close it and keep this issue for tracking purposes.

geoand:
Hopefully I'll be able to pick this up this week

PR #270, opened 2018-11-15
Make the source of Service Metadata configurable

geoand:
Make service metadata configurable, so that users can tune it
however they see fit. Specifically we add:

* Add flag to enable/disable addition of labels to metadata
(default: true)
* Add optional prefix to label keys (default none)
* Add flag to enable/disable addition of annotation to metadata
(default: true)
* Add optional prefix to annotation keys (default none)

Fixes: #260

### DIFF
--- a/spring-cloud-kubernetes-discovery/src/main/java/org/springframework/cloud/kubernetes/discovery/KubernetesDiscoveryClient.java
+++ b/spring-cloud-kubernetes-discovery/src/main/java/org/springframework/cloud/kubernetes/discovery/KubernetesDiscoveryClient.java
@@ -23,6 +23,7 @@ import io.fabric8.kubernetes.api.model.Service;
 import io.fabric8.kubernetes.client.KubernetesClient;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.function.Predicate;
@@ -72,19 +73,35 @@ public class KubernetesDiscoveryClient implements DiscoveryClient {
 	public List<ServiceInstance> getInstances(String serviceId) {
 		Assert.notNull(serviceId,
 				"[Assertion failed] - the object argument must be null");
-		final Map<String, String> labels = getLabels(serviceId);
 
 		Endpoints endpoints = client.endpoints().withName(serviceId).get();
 		List<EndpointSubset> subsets = null != endpoints ? endpoints.getSubsets() : new ArrayList<>();
 		List<ServiceInstance> instances = new ArrayList<>();
 		if (!subsets.isEmpty()) {
+
+			final Service service = client.services().withName(serviceId).get();
+
+			final Map<String, String> metadata = new HashMap<>();
+			if(properties.isEnabledAdditionOfLabelsAsMetadata()) {
+				metadata.putAll(
+					getMapWithPrefixedKeys(
+						service.getMetadata().getLabels(), properties.getLabelKeysPrefix())
+				);
+			}
+			if(properties.isEnabledAdditionOfAnnotationsAsMetadata()) {
+				metadata.putAll(
+					getMapWithPrefixedKeys(
+						service.getMetadata().getAnnotations(), properties.getAnnotationKeysPrefix())
+				);
+			}
+
 			for (EndpointSubset s : subsets) {
 				List<EndpointAddress> addresses = s.getAddresses();
 				for (EndpointAddress a : addresses) {
 					instances.add(new KubernetesServiceInstance(serviceId,
 							a,
 							s.getPorts().stream().findFirst().orElseThrow(IllegalStateException::new),
-							labels,
+							metadata,
 							false));
 				}
 			}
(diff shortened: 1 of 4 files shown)


spring-cloud-kubernetes-discovery/src/main/java/org/springframework/cloud/kubernetes/discovery/KubernetesDiscoveryProperties.java
```diff
@@ -29,11 +29,36 @@
...
+	 * When set, the Kubernetes labels of the services will be included as metadata
+	 * of the returned ServiceInstance
+	 */
+	private boolean enabledAdditionOfLabelsAsMetadata = true;
+
+	/**
+	 * When enabledAdditionOfLabelsAsMetadata is set, then the value labelKeysPrefix
+	 * will be used as a prefix to the key names in the metadata map
+	 */
+	private String labelKeysPrefix = "";
```
spencergibb: I'd rather a null here.

spring-cloud-kubernetes-discovery/src/main/java/org/springframework/cloud/kubernetes/discovery/KubernetesDiscoveryProperties.java
```diff
@@ -29,11 +29,36 @@
...
+	 * When set, the Kubernetes annotations of the services will be included as metadata
+	 * of the returned ServiceInstance
+	 */
+	private boolean enabledAdditionOfAnnotationsAsMetadata = true;
+
+	/**
+	 * When enabledAdditionOfAnnotationsAsMetadata is set, then the value annotationKeysPrefix
+	 * will be used as a prefix to the key names in the metadata map
+	 */
+	private String annotationKeysPrefix = "";
```
spencergibb: I'd rather a null here.

spring-cloud-kubernetes-discovery/src/main/java/org/springframework/cloud/kubernetes/discovery/KubernetesDiscoveryClient.java
```diff
@@ -93,12 +110,18 @@ public String description() {
...
 
-	private Map<String, String> getLabels(String serviceName) {
-		final Service service = client.services().withName(serviceName).get();
-		if (service != null) {
-			return service.getMetadata().getLabels();
+	// always returns a new map that contain all the entries of the original map
+	// but with the keys prefixed
+	private Map<String, String> getMapWithPrefixedKeys(Map<String, String> map, String prefix) {
+		// when the prefix is empty just return an map with the same entries
+		if (prefix.isEmpty()) {
```
spencergibb: use `StringUtils.hasText(prefix)` (assuming prefix defaults to null: see other comments).

spring-cloud-kubernetes-discovery/src/main/java/org/springframework/cloud/kubernetes/discovery/KubernetesDiscoveryClient.java
```diff
@@ -93,12 +110,18 @@ public String description() {
...
+			return new HashMap<>(map);
```
spencergibb: Why a new map?

spring-cloud-kubernetes-discovery/src/main/java/org/springframework/cloud/kubernetes/discovery/KubernetesDiscoveryProperties.java
geoand: Fixed

spring-cloud-kubernetes-discovery/src/main/java/org/springframework/cloud/kubernetes/discovery/KubernetesDiscoveryClient.java
geoand: I return a new map to keep the method behavior consistent in all cases - it always returns a new map regardless of the input. That way if it's ever used in another invocation, it's crystal clear how it behaves and doesn't present any surprises.

That said, I have no problem changing it if you like :)

spencergibb: It's return value is input to `metadata.putAll()`. May as well not do extra work.

geoand: Fixed

spring-cloud-kubernetes-discovery/src/main/java/org/springframework/cloud/kubernetes/discovery/KubernetesDiscoveryClient.java
```diff
@@ -93,12 +111,19 @@ public String description() {
 		return instances;
 	}
 
-	private Map<String, String> getLabels(String serviceName) {
-		final Service service = client.services().withName(serviceName).get();
-		if (service != null) {
```
spencergibb: Last item: this null check is gone. Is that significant (I don't know the API)?

Issue #104, opened 2017-06-13
Support services with multiple ports in KubernetesDiscoveryClient


manderson23:
I have a zookeeper Service configured as follows:
```
apiVersion: v1
kind: Service
metadata:
  name: zoo1
  labels:
    app: zookeeper-1
spec:
  ports:
  - name: client
    port: 2181
    protocol: TCP
  - name: follower
    port: 2888
    protocol: TCP
  - name: leader
    port: 3888
    protocol: TCP
  selector:
    app: zookeeper-1
```
In this case the K8 client finds an Endpoint for `zoo1` where the `EndpointSubset` has one address and three ports.

A call to `discoveryClient.getInstances("zoo1")` returns the `leader` port only whereas I would like the `client` port.

Should `KubernetesDiscoveryClient` return three `ServiceInstance` in this case? One for each port?

spencergibb:
Are the services on each port distinct? We do something similar in s-c-consul and s-c-zookeeper if the management port is set.

manderson23:
ZK deployment and service is configured in a similar manner to the yaml within spring-cloud-dataflow-server-kubernetes.

There are then 3 instances of each ZK service/deployment pair.

spencergibb:
@iocanel thoughts?

iocanel:
In theory we could return all service instances and add the port name as part of the metadata, so that the user can filter ports if needed. 

The thing is that the discovery client is not just something used by the user (who can adapt to whatever convention we use), but is also used by other parts of the ecosystem, that will fail to properly handle the convention. 

So I really don't know what's best and would adopt whatever practice 'spring cloud' uses for similar cases.

Now, specifically for zookeeper in kubernetes, I find it really handy to define the client port separately from peer, and election ports. Why? Because for peer and election ports you need reference them by node id (and that's why you define one service per node), while for client port this is not the case. Here's an example: https://github.com/fabric8io/fabric8-ipaas/tree/master/zookeeper-ensemble/src/main/fabric8

Zetten:
I have a branch at https://github.com/Zetten/spring-cloud-kubernetes/tree/ports-as-metadata which simply adds each named port to the metadata map as `port.<port-name>`.

This meets my particular need, which was previously handled with Eureka instance metadata: my service exposes both HTTP and gRPC endpoints, and I'd like to be able to select between them.

[...]

PR #271, opened 2018-11-16
Add K8s Service ports as optional Service Metadata

Zetten:
* Add flag to enable/disable addition of K8s Service ports to the
ServiceInstance metadata map (default: true).
* Add optional prefix to port keys in the metadata map
(default: "port.").

Fixes: gh-104

### DIFF
--- a/spring-cloud-kubernetes-discovery/src/main/java/org/springframework/cloud/kubernetes/discovery/KubernetesDiscoveryClient.java
+++ b/spring-cloud-kubernetes-discovery/src/main/java/org/springframework/cloud/kubernetes/discovery/KubernetesDiscoveryClient.java
@@ -17,9 +17,11 @@
 package org.springframework.cloud.kubernetes.discovery;
 
 import io.fabric8.kubernetes.api.model.EndpointAddress;
+import io.fabric8.kubernetes.api.model.EndpointPort;
 import io.fabric8.kubernetes.api.model.EndpointSubset;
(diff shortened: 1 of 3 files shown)


geoand:
@Zetten You need to update the Groovy test as well: https://github.com/spring-cloud/spring-cloud-kubernetes/blob/42c263c6be1096a84d476efd973cdc838480c62c/spring-cloud-kubernetes-discovery/src/test/groovy/io/fabric8/spring/cloud/discovery/KubernetesDiscoveryClientTest.groovy

It's very easy to miss that one since the IDE probably won't run it by default, but Maven picks it up :)

spencergibb:
BTW, I updated the properties added here and in the previous PR #270.

I moved them to an inner class. a277bf1d1b47e6f01ac97d9596d6a66777ca8e05

This makes the properties grouped better (rather than in the name)
and easier to read.

```java
properties.getMetadata().isAddLabels()
// vs 
properties.isEnabledAdditionOfLabelsAsMetadata()
```

geoand:
Thanks @spencergibb!

Zetten:
Yep, thanks for the quick action :grinning:

I'm looking forward to seeing the new functionality in a release. Is `spring-cloud-kubernetes:master` on track for making it into Greenwich.RELEASE, or was that branch cut previously?

spencergibb:
See https://github.com/spring-cloud/spring-cloud-release/milestones?direction=asc&sort=due_date

We're hoping for today but ran into a technical snag.

Issue #105, opened 2017-06-22
The client configuration bean requires additional settings


iocanel:
Currently its out of sync with https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/Config.java

Some properties of interest are the proxy related properties which are definitely missing.

spencergibb:
@iocanel, @geoand or @salaboy anyone want to give it a go?

geoand:
I will assign it to me and take a look (hopefully within the week)

PR #273, opened 2018-11-22
Add proxy properties for the Kubernetes client configuration

geoand:
Fixes: #105


geoand:
@iocanel Are there any more properties that you think it makes sense to add?

PR #275, opened 2018-11-25
Doc improvements and fix 268

### LINKED ISSUE #268: Secrets can be read using API given proper RBAC is in place
This statement in the README seems misleading:

```
Please note that by default, consuming Secrets via API (points 2 and 3 above) is not enabled for security reasons and it is recommend that containers share secrets via mounted volumes.
```

As if we use proper RBAC where pod is restricted only to read its on secrets then reading through API has no security concerns.

salaboy:
Fixes #268 and add documentation examples and resources

### DIFF
--- a/docs/src/main/asciidoc/examples.adoc
+++ b/docs/src/main/asciidoc/examples.adoc
@@ -1,5 +1,20 @@
 = Examples
 
+Spring Cloud Kubernetes tries to make it transparent for your applications to consume Kubernetes Native Services
+following the Spring Cloud interfaces. For your applications, you will only need to change the implementation of these interfaces
+if you were using different DiscoveryClient or PropertySourceLocator implementations.
+
+The following projects highlight the usage of these dependencies and demonstrate how these libraries can be used from any Spring Boot application.
+
 List of examples using these projects:
 
-<TBD>
+- Spring Cloud Kubernetes Examples: the ones located inside this repository.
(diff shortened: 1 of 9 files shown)


docs/src/main/asciidoc/examples.adoc
```diff
@@ -1,5 +1,22 @@
 = Examples
 
+Spring Cloud Kubernetes tries to make it transparent for your applications to consume Kubernetes Native Services
+following the Spring Cloud interfaces. For your applications, you will only need to change the implementation of these interfaces
+if you were using different DiscoveryClient or PropertySourceLocator implementations.
```
ryanjbaxter: > For your applications, you will only need to change the implementation of these interfaces
if you were using different DiscoveryClient or PropertySourceLocator implementations.

What does this mean?

salaboy: @ryanjbaxter that means that if you have for example the Eureka implementation of DiscoveryClient in your classpath you can swap that dependency for spring-cloud-kubernetes-discovery and the application should work. 
If that make sense, please help me out to express that out correctly in English :)

ryanjbaxter: I would just say that exactly what you told me, you need to add the `spring-cloud-kubernetes-discovery` dependency to your classpath and remove any other dependency that contains a `DiscoveryClient` implementation.

PR #279, opened 2018-11-29
Added the instance id to the `KubernetesServiceInstance`

TYsewyn:
Implemented in favor of spring-cloud/spring-cloud-commons#419


geoand:
Hi @TYsewyn 

Could you please give some information about what the `instanceId` is used for and why in the case of Spring Cloud Kubernetes it's the pod name?

Thanks

TYsewyn:
I was having the same concerns an hour ago, but the problem is that you might not be able to reuse the predicate that is used to filter out the services. The other problem is that `spring.cloud.kubernetes.client.namespace` only allows one namespace, and the services are operating cross-namespace.
Something that could be useful is the label selector on the service, but not every service has this set and that would imply making an extra call to the master node per service if it has been set.

geoand:
Yes that is pretty much was I was thinking as well. I am not really sure how we should tackle this...

TYsewyn:
Since the discovery client has been greatly changed over time, I'd like to propose a new solution.
Although I didn't test this, I think the easiest solution would be to use the UID of the target ref of an `EndpointAddress`.
The change would be around [this line](https://github.com/spring-cloud/spring-cloud-kubernetes/blob/master/spring-cloud-kubernetes-discovery/src/main/java/org/springframework/cloud/kubernetes/discovery/KubernetesDiscoveryClient.java#L154), using `endpointAddress.getTargetRef().getUid()` as instance id.

WDYT @geoand @salaboy ?

geoand:
I like it @TYsewyn!

salaboy:
@TYsewyn same as @geoand great stuff!

TYsewyn:
According to the integration test this works like a charm! :)

TYsewyn:
Can we get this merged? 🙏 I want to make use of it already 😁

ryanjbaxter:
@TYsewyn can you update the pr so we can merge this?

TYsewyn:
@ryanjbaxter done 😁

PR #286, opened 2018-12-07
#233 Initial Istio Awareness implementation

### LINKED ISSUE #233: Create Istio awareness module
Create a module that help Spring Cloud components to be aware that they are running in an Istio enabled environment to avoid clashes.

salaboy:
fixes #233 initial implementation

This provides the Istio profile addition to your application that can be used to configure different components to behave differently when Istio is present.


spring-cloud-kubernetes-istio/src/main/java/org/springframework/cloud/kubernetes/istio/utils/StandardMeshUtils.java
```diff
@@ -0,0 +1,62 @@
...
+	public Boolean isIstioEnabled() {
+		return checkIstioServices();
+	}
+
+	private synchronized boolean checkIstioServices() {
+		try {
+			//Check if Istio proxy is installed.
+			// We can improve this initial detection if better methods are found.
+			WebClient client = WebClient.create("http://localhost:15090/");
+			ClientResponse clientResponse = client.get().uri("stats/prometheus").exchange().block();
```
geoand: I think the port and path should be configurable. I can envision some Istio installation were these defaults would not apply.

Furthermore, is there way to return immediately once the first byte of an  HTTP 200 is returned?

salaboy: @geoand yeah.. I was checking on how to get just the http code.. but I couldn't find anything for that (on my quick search)

geoand: @spencergibb any ideas?

salaboy: I will make the URL configurable.. but the problem is that we are relaying on this to be local, so I wouldn't recommend to change this default values.

geoand: The base for the URL should never change. However the port and the proxy path to be used should be configurable so we can cover all cases

spencergibb: `WebClient` isn't always available either since it is in webflux.

spring-cloud-kubernetes-istio/src/main/java/org/springframework/cloud/kubernetes/istio/profile/IstioProfileApplicationListener.java
```diff
@@ -0,0 +1,79 @@
...
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.springframework.boot.context.event.ApplicationEnvironmentPreparedEvent;
+import org.springframework.cloud.kubernetes.istio.utils.MeshUtils;
+import org.springframework.context.ApplicationListener;
+import org.springframework.core.Ordered;
+import org.springframework.core.env.ConfigurableEnvironment;
+import org.springframework.core.env.Environment;
+
+public class IstioProfileApplicationListener implements ApplicationListener<ApplicationEnvironmentPreparedEvent>, Ordered {
```
spencergibb: One place we've not been shy about going over the network is with `BootstrapConfiguration` (see config client, consul, and zookeeper). Let's move the contacting of the local envoy to a `BootstrapConfiguration`. You should be able to use `@ConfigurationProperties` to parameterize host/port etc... Have it add a boolean property to the environment (`spring.cloud.istio.enabled=true` or something), then code like this later can just check the value of that property and activate the profile.

spring-cloud-kubernetes-istio/src/main/java/org/springframework/cloud/kubernetes/istio/utils/StandardMeshUtils.java
salaboy: I will switch to rest tplate here

salaboy: @spencergibbs switching to a bootstrap config right now

spring-cloud-kubernetes-istio/src/main/java/org/springframework/cloud/kubernetes/istio/utils/StandardMeshUtils.java
```diff
@@ -0,0 +1,62 @@
...
+			if (clientResponse.statusCode().is2xxSuccessful()) {
+				return true;
+			}
+			return false;
+		} catch (Throwable t) {
+			LOG.warn("Failed to get Istio Resources. Are you missing serviceaccount permissions?", t);
```
geoand: This warning is no longer relevant since no call to the API server is being made.

spring-cloud-kubernetes-istio/src/main/java/org/springframework/cloud/kubernetes/istio/utils/StandardMeshUtils.java
geoand: Also I am not sure that `15090` is the proper default. I mean it is for openshift, but on the microk8s cluster I created the port was `15000` (which also seems to be the default according to istio [docs](https://istio.io/docs/reference/commands/pilot-agent/))

spring-cloud-kubernetes-integration-tests/istio/src/test/java/org/springframework/cloud/kubernetes/it/ProfilesIT.java
```diff
@@ -0,0 +1,28 @@
...
+	private static final Integer PORT = Integer.valueOf(System.getProperty("service.port"));
+
+	@Test
+	public void testProfileEndpoint() {
+		given()
+			.baseUri(String.format("http://%s:%d", HOST, PORT))
+			.get("profiles")
+			.then()
+			.statusCode(200)
+			.body(new StringContains("istio"));
```
salaboy: @geoand should we check that Kubernetes is also here?

geoand: We could, but we probably don't really need to :)

spring-cloud-kubernetes-integration-tests/istio/src/test/resources/logback-test.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<configuration>
+    <include resource="org/springframework/boot/logging/logback/base.xml" />
+    <logger name="org.hibernate.validator" level="info" /> <!-- Validator prints a lot of debug messages during integration tests -->
```
spencergibb: This can be done in application.properties

spring-cloud-kubernetes-istio/src/main/resources/META-INF/additional-spring-configuration-metadata.json
```diff
@@ -0,0 +1,16 @@
+{
```
spencergibb: No need to add this manually. Should be auto generated for configuration properties

spring-cloud-kubernetes-istio/src/main/java/org/springframework/cloud/kubernetes/istio/utils/MeshUtils.java
```diff
@@ -0,0 +1,26 @@
...
+ *     Unless required by applicable law or agreed to in writing, software
+ *     distributed under the License is distributed on an "AS IS" BASIS,
+ *     WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *     See the License for the specific language governing permissions and
+ *     limitations under the License.
+ */
+
+package org.springframework.cloud.kubernetes.istio.utils;
+
+public interface MeshUtils {
```
spencergibb: Why the interface?

spring-cloud-kubernetes-integration-tests/istio/src/test/resources/logback-test.xml
geoand: But would it apply to test as well (which keep in mind does not involve Spring)?

spring-cloud-kubernetes-istio/src/main/resources/META-INF/additional-spring-configuration-metadata.json
salaboy: @spencergibb really? who generates that ? because I needed to add that for core

spring-cloud-kubernetes-integration-tests/istio/src/test/resources/logback-test.xml
salaboy: @spencergibb I will remove this one then.

salaboy: @geoand are you ok with removing that one?

geoand: Can't do much harm either ways :)

spring-cloud-kubernetes-istio/src/main/java/org/springframework/cloud/kubernetes/istio/IstioAutoConfiguration.java
```diff
@@ -0,0 +1,20 @@
...
+import io.fabric8.kubernetes.client.Config;
+import me.snowdrop.istio.client.DefaultIstioClient;
+import me.snowdrop.istio.client.IstioClient;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+@ConditionalOnProperty(value = "spring.cloud.istio.enabled", matchIfMissing = true)
```
ryanjbaxter: should we specify that the value needs to be `true`?

spring-cloud-kubernetes-istio/src/main/java/org/springframework/cloud/kubernetes/istio/IstioBootstrapConfiguration.java
```diff
@@ -0,0 +1,79 @@
...
+					}
+				} else {
+					if (LOG.isDebugEnabled()) {
+						LOG.debug("Adding 'istio' to list of active profiles");
+					}
+					environment.addActiveProfile(ISTIO_PROFILE);
+				}
+			} else {
+				if (LOG.isDebugEnabled()) {
+					LOG.warn("Not running inside kubernetes with istio enabled. Skipping 'istio' profile activation.");
```
ryanjbaxter: why check debug and then log warn?

spring-cloud-kubernetes-istio/src/main/java/org/springframework/cloud/kubernetes/istio/IstioBootstrapConfiguration.java
```diff
@@ -0,0 +1,79 @@
...
+				}
+			}
+		}
+
+		private boolean hasIstioProfile(Environment environment) {
+			for (String activeProfile : environment.getActiveProfiles()) {
```
ryanjbaxter: just a suggestion to make this code a bit simpler you could do
`Arrays.stream(environment.getActiveProfiles()).anyMatch(ISTIO_PROFILE::equalsIgnoreCase);`

spring-cloud-kubernetes-istio/src/main/java/org/springframework/cloud/kubernetes/istio/utils/MeshUtils.java
```diff
@@ -0,0 +1,64 @@
...
+
+	public Boolean isIstioEnabled() {
+		return checkIstioServices();
+	}
+
+	private synchronized boolean checkIstioServices() {
+		try {
+			//Check if Istio Envoy proxy is installed. Notice that the check is done to localhost.
+			// TODO: We can improve this initial detection if better methods are found.
+			RestTemplate restTemplate = new RestTemplateBuilder().build();
```
ryanjbaxter: Could we move this outside of the method?  I dont see a need to create the `RestTemplate` each time.

spring-cloud-kubernetes-istio/src/main/java/org/springframework/cloud/kubernetes/istio/IstioAutoConfiguration.java
spencergibb: No need. It works without it in the simple case.

spring-cloud-kubernetes-istio/src/main/java/org/springframework/cloud/kubernetes/istio/IstioBootstrapConfiguration.java
salaboy: +1

spring-cloud-kubernetes-istio/src/main/java/org/springframework/cloud/kubernetes/istio/utils/MeshUtils.java
salaboy: totally

salaboy:
@ryanjbaxter @spencergibb @geoand all changes done.. I will switch to work on the RC2 issues now. let me know if there is more that needs to be added here.

Issue #289, opened 2018-12-12
Use Duration for updatePeriod in LeaderProperties


manderson23:
For consistency with Spring Boot (and other properties I have in my application) I'd like to specify this property using the flexible duration syntax (e.g. 5s).

Not sure if there are also other properties this would apply to in the project.

spencergibb:
PRs welcome

manderson23:
@geoand what is the schedule for final release?

geoand:
I think `1.0.0.RC2` is planned for next week and `1.0.0.RELEASE` for January

steven-sheehy:
@manderson23 Can this be closed?

manderson23:
Yes

spencergibb:
Closed via #294

PR #294, opened 2018-12-13
Use java.time.Duration for duration based properties.

manderson23:
Implements #289


spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/reload/ConfigReloadProperties.java
```diff
@@ -52,7 +54,7 @@
 	/**
 	 * Sets the polling period in milliseconds to use when the detection mode is POLLING.
 	 */
-	private Long period = 15000L;
+	private Duration period = Duration.ofMillis(15000L);
```
spencergibb: Need to change the javadoc, since it's not millis anymore

manderson23: Done

PR #295, opened 2018-12-13
Add InfoContributor reporting pod details

### LINKED ISSUE #128: Add InfoContributor to provide Pod details
It would be useful if the /info endpoint included a "kubernetes" key containing the same Pod details are returned by the health indicator.

manderson23:
Resolves #128.

This initial implementation is a clone of the Health indicator details. Happy to make changes or augment with anything else that might be useful.


spring-cloud-kubernetes-core/src/test/java/org/springframework/cloud/kubernetes/InfoContributorTest.java
```diff
@@ -0,0 +1,56 @@
+package org.springframework.cloud.kubernetes;
```
ryanjbaxter: Missing copyright header

PR #323, opened 2019-02-03
Add profile expression support

pukkaone:
Spring Boot 2.1 introduced the ability to match profiles by [profile expression](https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.1-Release-Notes#profile-expression). For example, if a YAML document specifies

    spring.profiles: production & us-east

then the properties from the document are loaded if both the `production` and `us-east` profiles are active.

### DIFF
--- a/spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/ConfigMapPropertySource.java
+++ b/spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/ConfigMapPropertySource.java
@@ -27,7 +27,9 @@ import io.fabric8.kubernetes.client.KubernetesClient;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
+import org.springframework.core.env.Environment;
 import org.springframework.core.env.MapPropertySource;
+import org.springframework.core.env.StandardEnvironment;
 import org.springframework.util.StringUtils;
 
 import static org.springframework.cloud.kubernetes.config.PropertySourceUtils.KEY_VALUE_TO_PROPERTIES;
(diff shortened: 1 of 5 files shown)


geoand:
Very nice @pukkaone, thanks for this!

I'll try and go over this in the next few days, but at a glance this looks great!
I would appreciate it if @ryanjbaxter and @spencergibb take a look as well

spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/ConfigMapPropertySource.java
```diff
@@ -42,14 +43,10 @@
 
 	private static final String PREFIX = "configmap";
 
-	public ConfigMapPropertySource(KubernetesClient client, String name) {
-		this(client, name, null, null);
-	}
-
 	public ConfigMapPropertySource(KubernetesClient client, String name, String namespace,
```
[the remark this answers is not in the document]
pukkaone: I assumed deleting this constructor is safe because [only the tests invoke it](https://github.com/spring-cloud/spring-cloud-kubernetes/blob/master/spring-cloud-kubernetes-config/src/test/java/org/springframework/cloud/kubernetes/config/ConfigMapsTest.java#L87). I will put it back.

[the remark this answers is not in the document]
spencergibb: It's a public API. We have to assume someone is using it and would break with this change.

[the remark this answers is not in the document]
pukkaone: I restored the constructor in the original public API.

spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/ConfigMapPropertySource.java
```diff
@@ -56,9 +57,9 @@ public ConfigMapPropertySource(KubernetesClient client, String name) {
 	}
 
 	public ConfigMapPropertySource(KubernetesClient client, String name, String namespace,
```
spencergibb: I don't see the change back.

pukkaone: You don't see the change back because I squashed the changes into a single commit and force pushed to this pull request branch.

spencergibb: The PR still shows a new one, so not sure what you did, but this PR doesn't have that change.

pukkaone: Got it. I put back all original constructors and added a new constructor.

Issue #327, opened 2019-02-08
ConfigMap is not loaded from mounted volume


Lex4hex:
Hello! 

spring-cloud-kubernetes-config version is 1.0.0. 
It wasn't an issue on 0.3.0.

I'm trying to load *.properties file from mounted volume: 

```volumeMounts:
            - mountPath: "/config-volume/"
              name: config-volume
              readOnly: true
      restartPolicy: Always
      volumes:
        - name: config-volume
          configMap:
            name: example-volume-file-configmap
            items:
              - key: example.application.properties
                path: example.application.properties
```

My bootstrap.yaml: 

``` spring:
  profiles: volume-config-profile
  cloud:
    kubernetes:
      config:
        paths: "/config-volume/" 
```
  
Kubernetes creates a symlink to `example.application.properties` in `/config-volume` directory. 
But in this commit  https://github.com/spring-cloud/spring-cloud-kubernetes/commit/a006296c780df21cc899690c1545747b9d934168#diff-61a318ccc3b341cb8a04969cbc5961db @geoand added a filter for regular files `.filter(Files::isRegularFile)` so it skips symlink to my properties file.

geoand:
Hey @Lex4hex 

The intention there was to skip directories, so I would consider this a bug.
If you have some time to spare and would like to work on a fix that would be great :)

If not, I can handle it a few days

Lex4hex:
@geoand I'll fix it

geoand:
Cool!

iscra:
Hello,
I have just tracked my issue to the same problem here and went through all the associated comments. Yes, I have seen the notice in README, but interpreted it differently, as in "recursive is off" would mean subdirectories are not parsed, just the top one.

The issue is "spring.cloud.kubernetes.config.paths" must now have only files, never directories.

This is very inconsistent with how secrets work and not very usable.
I would have to list all property keys for it to work, if I understand that now correctly.
This is simply not feasible and not possible in my use case as I don't know which config will be defined for different kubernetes namespaces / staging environment - this happens in deployment.
I suppose it works just for the use case, when the properties are embedded again as full file under one key, so it's kind a workaround then.

Please support reading configmaps from files same as secrets, as it was before. If necessary this could be configured as another config option. All the code is still there used by secrets, just need to enable it.

geoand:
Hello,

What you describe was intentionally removed as an option.
[...]

PR #328, opened 2019-02-08
Refs #327 ~Fix property files searching in directories defined by

Lex4hex:
https://github.com/spring-cloud/spring-cloud-kubernetes/issues/327 
Added recursive traversing of provided directory.

### DIFF
--- a/spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/ConfigMapPropertySourceLocator.java
+++ b/spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/ConfigMapPropertySourceLocator.java
@@ -22,6 +22,7 @@ import java.nio.file.Paths;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Stream;
 
 import io.fabric8.kubernetes.api.builder.Function;
 import io.fabric8.kubernetes.client.KubernetesClient;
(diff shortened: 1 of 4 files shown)


geoand:
Looking at your PR, I guess I misunderstood what the problem was... Sorry about that!

The intention of a006296c was to remove the ability to traverse directories and locate properties files.
What I though the problem you described in #327 was that symlinks were not being used correctly.

So I am against this, sorry. If you want to add a PR that simply changes `.filter(Files::isRegularFile)` to something that doesn't filter out symlinks, I'd be happy to give +1

Lex4hex:
@geoand The problem wasn't with filtering out symlinks. `.filter(Files::isRegularFile)` is actually working as it should. 
Paths property contains paths to directories with property files  e.g. `/config-volume`, right? 

``
this.properties.getPaths().stream().map(Paths::get)
``
That code returned only one path which is "/config-volume" and it's a directory so no property files are being imported from  `/config-volume`.

geoand:
The whole idea of the discussion in #230 was to rationalize the way the ConfigMaps were turned into PropertySources. 
And part of that was to remove the traversing logic - since there is no analogy in a typical Spring Boot application. 
You now need to specify the entire path of the propertes/yaml file you wish to be included.

Do that make sense?

Lex4hex:
Oh, I get it know, thanks. I thought you were talking only about traversing of nested directories. 
So, now I have to provide exact path to my mounted property files. 

Maybe it'll be good idea to directly point this out in docs?

geoand:
> Oh, I get it know, thanks. I thought you were talking only about traversing of nested directories.
> So, now I have to provide exact path to my mounted property files.
> 

Yes that's the idea since that's exactly what Spring Boot does as well

> Maybe it'll be good idea to directly point this out in docs?

This is probably a good idea. Perhaps mentioning that the old behavior has been changed. Would you like to take that up? As I said before, I can obviously do that but it will take me a few days to get around to

Lex4hex:
>This is probably a good idea. Perhaps mentioning that the old behavior has been changed. Would you like to take that up? As I said before, I can obviously do that but it will take me a few days to get around to 

Yep, I'll create a PR with update. 

I'm closing this PR then.

geoand:
> > This is probably a good idea. Perhaps mentioning that the old behavior has been changed. Would you like to take that up? As I said before, I can obviously do that but it will take me a few days to get around to
> 
> Yep, I'll create a PR with update.
> 
> I'm closing this PR then.

Thanks for your help @Lex4hex and sorry for the misunderstanding on my part about the original issue

Issue #293, opened 2018-12-13
Add InfoContributor to report leader election state


manderson23:
Add an InfoContributor to report leader election state. Something like

```
leaderElection: {
  leaderId: "id"
  role: "role"
  isLeader: true
}
```

manderson23:
I hope to be able to get a PR for this before final release.

PR #334, opened 2019-02-19
Add info contributor for leader election.

manderson23:
Implements #293

### DIFF
--- a/spring-cloud-kubernetes-leader/pom.xml
+++ b/spring-cloud-kubernetes-leader/pom.xml
@@ -39,7 +39,7 @@
 		</dependency>
 		<dependency>
 			<groupId>org.springframework.boot</groupId>
-			<artifactId>spring-boot-actuator</artifactId>
+			<artifactId>spring-boot-starter-actuator</artifactId>
 			<optional>true</optional>
 		</dependency>
 		<dependency>
@@ -55,11 +55,6 @@
 		<dependency>
 			<groupId>org.springframework.boot</groupId>
 			<artifactId>spring-boot-starter-test</artifactId>
-			<scope>test</scope>
-		</dependency>
-		<dependency>
-			<groupId>org.springframework.boot</groupId>
-			<artifactId>spring-boot-starter-actuator</artifactId>
 			<scope>test</scope>
 		</dependency>
 	</dependencies>
--- a/spring-cloud-kubernetes-leader/pom.xml
+++ b/spring-cloud-kubernetes-leader/pom.xml
@@ -39,12 +39,27 @@
 		</dependency>
 		<dependency>
 			<groupId>org.springframework.boot</groupId>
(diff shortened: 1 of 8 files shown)


spring-cloud-kubernetes-leader/pom.xml
```diff
@@ -37,6 +37,16 @@
...
 			<artifactId>spring-integration-core</artifactId>
 		</dependency>
+		<dependency>
+			<groupId>org.springframework.boot</groupId>
+			<artifactId>spring-boot-actuator</artifactId>
+			<optional>true</optional>
+		</dependency>
+		<dependency>
+			<groupId>io.rest-assured</groupId>
+			<artifactId>rest-assured</artifactId>
```
spencergibb: I'd rather not use rest assured here. `WebTestClient` should be sufficient.

manderson23: I also updated the other tests I'd based mine one.

spring-cloud-kubernetes-leader/pom.xml
```diff
@@ -37,16 +37,31 @@
...
+			<artifactId>spring-boot-starter-webflux</artifactId>
+			<scope>test</scope>
+		</dependency>
 		<dependency>
 			<groupId>org.springframework.boot</groupId>
 			<artifactId>spring-boot-starter-test</artifactId>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>org.springframework.boot</groupId>
```
ryanjbaxter: Why do we need to `actuator-starter` and `actuator`?

spencergibb: yup, just use `starter-actuator` and make it optional.

manderson23: This usage of optional `actuator` for the main code and `starter-actuator` for testing is the same approach as both the `core` and `config` `pom.xml`.

I assumed I should be consistent? My understanding was that `starter` is only required for the tests to auto-detect etc.

spencergibb: it's what we've done in other places.

manderson23:
Would you like it to be changed across all poms for consistency?

spencergibb:
not for this PR

spring-cloud-kubernetes-leader/src/test/java/org/springframework/cloud/kubernetes/leader/LeaderAutoConfigurationTests.java
```diff
@@ -31,10 +37,23 @@
 })
 public class LeaderAutoConfigurationTests {
 
+	@Value("${local.server.port}")
```
spencergibb: In the future this can be `@LocalServerPort`

Issue #326, opened 2019-02-06
Remove Version Elements From POM Samples In Documentation


ryanjbaxter:
```
<dependency>
    <groupId>org.springframework.cloud</groupId>
    <artifactId>spring-cloud-starter-kubernetes-ribbon</artifactId>
    <version>${latest.version}</version>
</dependency>
```

The version should be managed by the BOM

PR #337, opened 2019-02-26
Editing pass

Buzzardo:
I edited for spelling, punctuation, grammar, usage, and corporate voice. I also corrected some links and made sure everything points uses https rather than http.

### DIFF
--- a/docs/pom.xml
+++ b/docs/pom.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-		 xmlns="http://maven.apache.org/POM/4.0.0"
-		 xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns:xsi="https://www.w3.org/2001/XMLSchema-instance"
+		 xmlns="https://maven.apache.org/POM/4.0.0"
+		 xsi:schemaLocation="https://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
 	<modelVersion>4.0.0</modelVersion>
 	<groupId>org.springframework.cloud</groupId>
 	<artifactId>spring-cloud-kubernetes-docs</artifactId>
@@ -62,4 +62,3 @@
 		</profile>
 	</profiles>
 </project>
-
--- a/docs/src/main/asciidoc/discovery-client.adoc
+++ b/docs/src/main/asciidoc/discovery-client.adoc
@@ -15,7 +15,6 @@ This is something that you get for free by adding the following dependency insid
 <dependency>
     <groupId>org.springframework.cloud</groupId>
     <artifactId>spring-cloud-starter-kubernetes</artifactId>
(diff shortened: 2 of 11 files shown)


docs/pom.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-		 xmlns="http://maven.apache.org/POM/4.0.0"
-		 xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns:xsi="https://www.w3.org/2001/XMLSchema-instance"
```
pukkaone: The namespace URI is treated only has an identifier. It is [not a goal](https://www.w3.org/TR/2009/REC-xml-names-20091208/#ns-decl) that it be directly usable for retrieval of a schema (if any exists). The URI should be exactly as documented. Changing the scheme to `https` actually makes it a different namespace name.

spencergibb: Don't know that really matters here

Buzzardo: The guidance I've gotten is to replace them. The build works either way.

ryanjbaxter:
@Buzzardo mind looking at the conflicts?

docs/src/main/asciidoc/ribbon-integration.adoc
```diff
@@ -1,40 +1,54 @@
 == Ribbon discovery in Kubernetes
 
 
-Spring Cloud client applications calling a microservice should be interested on relying on a client load-balancing
+Spring Cloud client applications that call a microservice should be interested on relying on a client load-balancing
 feature in order to automatically discover at which endpoint(s) it can reach a given service. This mechanism has been
-implemented within the [spring-cloud-kubernetes-ribbon](spring-cloud-kubernetes-ribbon/pom.xml) project where a
-Kubernetes client will populate a https://github.com/Netflix/ribbon[Ribbon] `ServerList` containing information
+implemented within the https://github.com/fabric8io/spring-cloud-kubernetes/tree/master/spring-cloud-kubernetes-ribbon[spring-cloud-kubernetes-ribbon] project, where a
```
pukkaone: In the URL, the path segment `fabric8io` should be `spring-cloud`.

docs/src/main/asciidoc/discovery-client.adoc
```diff
@@ -1,36 +1,58 @@
...
+This is something that you get for free by adding the following dependency inside your project:
 
-```java
+====
+[source,xml]
+----
+<dependency>
+    <groupId>org.springframework.cloud</groupId>
+    <artifactId>spring-cloud-starter-kubernetes</artifactId>
+    <version>${latest.version}</version>
```
pukkaone: This reverses the change in issue https://github.com/spring-cloud/spring-cloud-kubernetes/issues/326 to remove version elements from the POM examples.

Buzzardo: I didn't add that line, but I have removed it now.

docs/src/main/asciidoc/ribbon-integration.adoc
Buzzardo: Fixed. Thanks.

PR #339, opened 2019-02-27
document reloader dependency on actuator-autoconfigure

paulczar:
Signed-off-by: Paul Czarkowski <username.taken@gmail.com>

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -476,6 +476,14 @@ This means, for example, that using polling on file mounted secret sources does 
 **Notes**:
 - Properties under *spring.cloud.kubernetes.reload.* should not be used in config maps or secrets: changing such properties at runtime may lead to unexpected results;
 - Deleting a property or the whole config map does not restore the original state of the beans when using the *refresh* level.
+- You need to have the reload actuator dependency in your pom file for reload to work:
+
+```xml
+<dependency>
+    <groupId>org.springframework.boot</groupId>
+	<artifactId>spring-boot-actuator-autoconfigure</artifactId>
+</dependency>
+```
 
 
 == Ribbon discovery in Kubernetes
--- a/README.adoc
+++ b/README.adoc
@@ -476,12 +476,12 @@ This means, for example, that using polling on file mounted secret sources does 
 **Notes**:
 - Properties under *spring.cloud.kubernetes.reload.* should not be used in config maps or secrets: changing such properties at runtime may lead to unexpected results;
 - Deleting a property or the whole config map does not restore the original state of the beans when using the *refresh* level.
-- You need to have the reload actuator dependency in your pom file for reload and the reload actuator enabled `management.endpoint.restart.enabled=true` in your `bootstrap.yaml` or `application.yaml`:
+- For reload to work you must have the `spring-boot-starter-actuator` dependency in your pom`.xml` and the reload actuator enabled `management.endpoint.restart.enabled=true` in your `bootstrap.yaml` or `application.yaml`:
 
 ```xml
 <dependency>
     <groupId>org.springframework.boot</groupId>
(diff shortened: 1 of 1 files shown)


paulczar:
The error in circleci appears to be unrelated to my change.

geoand:
I don't think this is actually needed...

spencergibb:
I think it is, the plain starter doesn't include config.

geoand:
I'm looking to see if the reload example works as is

paulczar:
It definitely needs the actuator and reload endpoint enabled, I was bashing my head against this all day yesterday.

README.adoc
```diff
@@ -476,6 +476,15 @@ This means, for example, that using polling on file mounted secret sources does
 **Notes**:
 - Properties under *spring.cloud.kubernetes.reload.* should not be used in config maps or secrets: changing such properties at runtime may lead to unexpected results;
 - Deleting a property or the whole config map does not restore the original state of the beans when using the *refresh* level.
+- You need to have the reload actuator dependency in your pom file for reload and the reload actuator enabled `management.endpoint.restart.enabled=true` in your `bootstrap.yaml` or `application.yaml`:
+
+```xml
+<dependency>
```
ryanjbaxter: If this is needed we should specify the spring boot starter here

geoand:
I just tried https://github.com/spring-cloud/spring-cloud-kubernetes/tree/master/spring-cloud-kubernetes-examples/kubernetes-reload-example which I just realized that is **does** have `spring-boot-actuator-autoconfigure`.

So yeah, this makes sense!

README.adoc
geoand: We probably should include the dependency in `spring-cloud-starter-kubernetes-config`

README.adoc
```diff
@@ -476,6 +476,15 @@ This means, for example, that using polling on file mounted secret sources does
 **Notes**:
 - Properties under *spring.cloud.kubernetes.reload.* should not be used in config maps or secrets: changing such properties at runtime may lead to unexpected results;
 - Deleting a property or the whole config map does not restore the original state of the beans when using the *refresh* level.
+- You need to have the reload actuator dependency in your pom file for reload and the reload actuator enabled `management.endpoint.restart.enabled=true` in your `bootstrap.yaml` or `application.yaml`:
```
ryanjbaxter: I dont think specifically we need to set `management.endpoint.restart.enabled=true`.  Maybe if you change to `spring.cloud.kubernetes.reload.strategy=restart_context` you might?  Is that what you did.  We should just need to enable the refresh endpoint by default.
```
management:
  endpoints:
    web:
      exposure:
        include: refresh
```

geoand:
@paulczar Would you also please update the reload example with the changes proposed by @ryanjbaxter?

README.adoc
paulczar: I've tried that.

```
spring.cloud.kubernetes.reload.strategy: refresh
#management.endpoint.restart.enabled: true
management.endpoints.web.exposure.include: refresh
```

gives me the following error during startup

```
Caused by: org.springframework.beans.factory.NoSuchBeanDefinitionException: No qualifying bean of type 'org.springframework.cloud.context.restart.RestartEndpoint' available: expected at least 1 bean which qualifies as autowire candidate. Dependency annotations: {}
```

spencergibb:
Actuator needs to remain optional.

README.adoc
paulczar: good catch, switched it across to use the starter

README.adoc
ryanjbaxter: @paulczar did you include `spring-boot-starter-actuator`?  Also can you try
```
management:
  endpoint:
    restart:
      enabled: true
```

paulczar: yeah both those are what the PR is adding to the docs :)

PR #401, opened 2019-05-31
README update for ConfigMap reload sample

anwarchk:
1. Fixed small typos
2. Removed notes about Spring Boot liveness and readiness check related
to Fabric8 since that issue has been resolved and is working as expected
3. Added sample configuration files for K8s RBAC set up required for the
example to work
4. Added section about how to setup RBAC config

### DIFF
--- a/spring-cloud-kubernetes-examples/kubernetes-reload-example/README.md
+++ b/spring-cloud-kubernetes-examples/kubernetes-reload-example/README.md
@@ -5,17 +5,17 @@ To play with these examples, you can install locally Kubernetes & Docker using `
 managed by a hypervisor (Xhyve, Virtualbox or KVM) if your machine is not a native Unix operating system.
 
   
-When the minikube  is installed on your machine, you can start kubernetes using this command:
+When the Minikube  is installed on your machine, you can start kubernetes using this command:
 ```
 minikube start
 ```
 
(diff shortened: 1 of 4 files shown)


geoand:
Thanks for the PR!

I am not sure that the ClusterServiceRole is actually the best way to go here (since it gives rights that might be too broad). @iocanel @salaboy WDYT?

iocanel:
I agree we should go with plain old view role.

anwarchk:
@geoand @iocanel We are still using the `view` role for the ServiceAccount. That is, we are binding the `config-reader` service account to the `view` ClusterRole. Is there a different example that you can please share that would bind to a less privileged role ? I would be happy to update the example config with that.

anwarchk:
@ryanjbaxter ^^^

geoand:
@anwarchk what I am concerned about is that it is a `ClusterRole`, meaning that it applies to all namespaces. If you could do the same but with a regular role, it would be enough I think

anwarchk:
@geoand Updated the sample RBAC configs with a regular role

geoand:
Thanks! Have you verified that this role works for your setup?

anwarchk:
Yes, just did :-)

geoand:
@ryanjbaxter would you like to take a look as well?

PR #404, opened 2019-06-04
Enhancment/improve config map property source logging

### LINKED ISSUE #400: Spring Cloud Kubernetes Config ConfigMapPropertySource class should provide better logging
Spring Cloud Kubernetes Config ConfigMapPropertySource class should provide better logging.  Currently it logs a WARN if there is a problem pulling in a configMap reference.  

`private static Map<String, String> getData(KubernetesClient client, String name,
			String namespace, Environment environment) {
		try {
			ConfigMap map = StringUtils.isEmpty(namespace)
					? client.configMaps().withName(name).get()
					: client.configMaps().inNamespace(namespace).withName(name).get();

			if (map != null) {
				return processAllEntries(map.getData(), environment);
			}
		}
		catch (Exception e) {
			LOG.warn("Can't read configMap with name: [" + name + "] in namespace:["
					+ namespace + "]. Ignoring");
		}

		return new HashMap<>();
	}`
The exception should either be logged or re-thrown.  Took a while to debug my issue where this would have been very helpful.

fico-jessecarroll:
Fixes #400 

Added exception to logging.  Let me know if message needs any changing.

Thanks

### DIFF
--- a/spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/ConfigMapPropertySource.java
+++ b/spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/ConfigMapPropertySource.java
@@ -98,7 +98,7 @@ public class ConfigMapPropertySource extends MapPropertySource {
 		}
 		catch (Exception e) {
 			LOG.warn("Can't read configMap with name: [" + name + "] in namespace:["
-					+ namespace + "]. Ignoring");
+					+ namespace + "]. Ignoring.  Reason for failure: ", e);
 		}
 
 		return new HashMap<>();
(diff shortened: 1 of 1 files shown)


geoand:
Thanks! The change itself looks good but I am not sure about the formatting thing... @ryanjbaxter ?

spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/ConfigMapPropertySource.java
```diff
@@ -98,7 +98,7 @@ private static String getName(KubernetesClient client, String name,
 		}
 		catch (Exception e) {
 			LOG.warn("Can't read configMap with name: [" + name + "] in namespace:["
-					+ namespace + "]. Ignoring");
```
ryanjbaxter: You don't need to add "Reason for failure" just pass the exception

fico-jessecarroll: Done

PR #408, opened 2019-06-10
Enhance docs about setting Spring profile at bootstrap via ENV Variable 

medmedchiheb:
i have updated the paragraph as discussed, but also i have modified the resource yaml file to be general I've used a POD, any suggestions !

### DIFF
--- a/docs/src/main/asciidoc/property-source-config.adoc
+++ b/docs/src/main/asciidoc/property-source-config.adoc
@@ -185,22 +185,33 @@ If both profiles are active, the property that appears last within the `ConfigMa
 
 To tell Spring Boot which `profile` should be enabled at bootstrap, you can pass `SPRING_PROFILES_ACTIVE` environment variable.
  To do so, you can launch your Spring Boot application with an environment variable that you can define it in the PodSpec at the container specification.
- Pod resource file, as follows:
+ Deployment resource file, as follows:
 
 ====
 [source,yaml]
 ----
-apiVersion: v1
-kind: Pod
+apiVersion: apps/v1
+kind: Deployment
 metadata:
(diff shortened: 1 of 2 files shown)


geoand:
Thanks for this!

Actually the README.adoc is generateds from the contents inside `docs`. Would you like to make the changes there (and revert the the README change) please?

medmedchiheb:
Alright,

docs/src/main/asciidoc/property-source-config.adoc
```diff
@@ -183,25 +183,24 @@ However, if the `production` profile is active, the configuration becomes:
...
+ Pod resource file, as follows:
 
 ====
 [source,yaml]
 ----
 apiVersion: v1
-kind: DeploymentConfig
+kind: Pod
+metadata:
+  name: pod-name
```
geoand: Actually I think that:

```
kind: Pod
metadata:
  name: pod-name
```

should be replaced by `kind: Deployment` since a user wouldn't be expected to configure the pod, but the Deployment

docs/src/main/asciidoc/property-source-config.adoc
```diff
@@ -183,25 +183,24 @@ However, if the `production` profile is active, the configuration becomes:
...
 spec:
-  replicas: 1
-  ...
-    spec:
```
geoand: Again, I think that was discussed was to change the name of the env var and replace `DeploymentConfig` with `Deployment`. Not show the pod configuration (which a user wouldn't specify directly, but via the `Deployment`)

medmedchiheb: done

PR #410, opened 2019-06-13
1.0.x update local leader before notifications

### LINKED ISSUE #405: Leadership election, local leader set to late?
Hi,

we migrated our Leader election code from Hazelcast to Kubernetes. The setup we have as of today is like:

```
@Override
    public void onApplicationEvent( AbstractLeaderEvent event )
    {
        boolean isLeader = event.getContext()
                                .isLeader();
        LOG.info( "Leadership changed, node is leader={}", isLeader );
        ...
     }
```

My expectation was that this will yield to the correct behaviour. Yet an `OnGrantedEvent`  where 
`event.getContext().isLeader()` is invoked will return `false`. On the contrary an `OnRevokedEvent` will yield for the same invocation `true`.

This is due the fact that LeadershipController is setting localLeader = newLeader after it notified on grant/revoke, which to me looks too late.

I can work around this issue by checking for the type of event received OnRevokedEvent/OnGrantedEvent rather than the isLeader() method. Yet I would assume that both should allow me to conclude the same result.

gytis:
Fixes #405.

@ryanjbaxter I had to pull-in other two commits in order to make the sample application work on the latest minikube.


ryanjbaxter:
@gytis there are test failures on circleci can you take a look?

gytis:
I'll have to revert the FMP upgrade. It expects health endopint at "/actuator/health", so won't work with Spring Boot 1.5.x.

gytis:
I dropped the FMP upgrade. But had to modify the example as a result.
FMP 3.5.x does not include Role and RoleBinding objects into a generated kubernetes.yml file, only openshift.yml. So I added leader-role.yml and leader-rolebinding.yml to the root directory of the example and included steps to apply them in the readme.

gytis:
Hmm, the same test failed again. It must be unrelated to the PR.

geoand:
I relaunched it, lets see what happens

ryanjbaxter:
Seems like it might be something with CircleCI.  I built the 1.0.x branch with and without your changes and it was fine on my machine.

PR #412, opened 2019-06-14
Highlight reference for info on consuming secrets via API

### LINKED ISSUE #252: Why is Consuming Secrets via API not secure ?
It is mentioned in the Readme: 

> Please note that by default, consuming Secrets via API (points 2 and 3 above) is not enabled for security reasons and it is recommend that containers share secrets via mounted volumes.

What are the security reasons due to which it is not enabled ? As I am thinking to consume secrets via API and I think it might be more secure as there won't be any file containing the secret mounted on the instance. 

Provided that we have correct RBAC and pod only reads its own secret, and that we have enabled encryption at REST, is it still not secure ? Why so ?

hazim1093:
Added more information to highlight the presence of risks and best practices when consuming secrets through the API

Fixes #252

### DIFF
--- a/docs/src/main/asciidoc/property-source-config.adoc
+++ b/docs/src/main/asciidoc/property-source-config.adoc
@@ -251,7 +251,7 @@ When enabled, the `SecretsPropertySource` looks up Kubernetes for `Secrets` from
 
 *Note:* 
 
-By default, consuming Secrets through the API (points 2 and 3 above) *is not enabled* for security reasons. The `list` permision on secrets allows the clients to inspect the values of all secrets that are in that namespace.
+By default, consuming Secrets through the API (points 2 and 3 above) *is not enabled* for security reasons. The permission 'list' on secrets allows clients to inspect secrets values in the specified namespace.
 Further, we recommend that containers share secrets through mounted volumes.
 
 If you enable consuming Secrets through the API, we recommend that you limit access to Secrets by using an authorization policy, such as RBAC. 
--- a/docs/src/main/asciidoc/property-source-config.adoc
+++ b/docs/src/main/asciidoc/property-source-config.adoc
@@ -253,7 +253,7 @@ Note that, by default, consuming Secrets through the API (points 2 and 3 above) 
 Further, we recommend that containers share secrets through mounted volumes.
 If you enable consuming Secrets through the API, we recommend that you limit access to Secrets by using an authorization policy, such as RBAC. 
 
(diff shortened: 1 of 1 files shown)


docs/src/main/asciidoc/property-source-config.adoc
```diff
@@ -251,8 +251,9 @@ When enabled, the `SecretsPropertySource` looks up Kubernetes for `Secrets` from
 
 Note that, by default, consuming Secrets through the API (points 2 and 3 above) *is not enabled* for security reasons.
 Further, we recommend that containers share secrets through mounted volumes.
-If you enable consuming Secrets through the API, we recommend that you limit access to Secrets by using an
-[authorization policy, such as RBAC](https://kubernetes.io/docs/concepts/configuration/secret/#best-practices).
+If you enable consuming Secrets through the API, we recommend that you limit access to Secrets by using an authorization policy, such as RBAC.
```
geoand: Perhaps we can also add a note that there is no way to limit access to specific secrets inside a Kubernetes namespace with RBAC - meaning that am app that can read a secret in a namespace can read all of them?

hazim1093: It is mentioned in the referenced link but we can also add it here

geoand: Since it's rather serious, I propose we mention it here as well

docs/src/main/asciidoc/property-source-config.adoc
```diff
@@ -251,8 +251,9 @@ When enabled, the `SecretsPropertySource` looks up Kubernetes for `Secrets` from
 
 Note that, by default, consuming Secrets through the API (points 2 and 3 above) *is not enabled* for security reasons.
 Further, we recommend that containers share secrets through mounted volumes.
-If you enable consuming Secrets through the API, we recommend that you limit access to Secrets by using an
-[authorization policy, such as RBAC](https://kubernetes.io/docs/concepts/configuration/secret/#best-practices).
+If you enable consuming Secrets through the API, we recommend that you limit access to Secrets by using an authorization policy, such as RBAC. 
+
+For more information about risks and best practices when consuiming Secrets through the API refer to https://kubernetes.io/docs/concepts/configuration/secret/#best-practices[this doc].
```
medmedchiheb: typo here! "consuiming" should be "consuming"

ryanjbaxter:
Can you submit this to the 1.0.x stream and we can merge it forward?

hazim1093:
updated, @ryanjbaxter should I open a new Pull Request to 1.0.x branch and close this one ?

ryanjbaxter:
Yes

ryanjbaxter:
Closed via 04f59cf

PR #413, opened 2019-06-19
Enhance Hystrix ThreadPool mode so that it can correctly pass Istio's call chain headers

idefav:
Enhance Hystrix ThreadPool mode so that it can correctly pass Istio's call chain headers, and set the RestTemplate interceptor to intercept ordinary Http requests, providing the ThreadPoolHolder operation class to manually synchronize custom thread pool variables.

Use the effect to string the entire call chain.
image


idefav:
image

PR #414, opened 2019-06-19
Enhance Hystrix ThreadPool mode so that it can correctly pass Istio's call chain headers

idefav:
Enhance Hystrix ThreadPool mode so that it can correctly pass Istio's call chain headers, and set the RestTemplate interceptor to intercept ordinary Http requests, providing the ThreadPoolHolder operation class to manually synchronize custom thread pool variables.

Use the effect to string the entire call chain.
image
image

### DIFF
--- a/spring-cloud-kubernetes-istio/src/main/java/org/springframework/cloud/kubernetes/istio/trace/FeignPropagationRequestInterceptor.java
+++ b/spring-cloud-kubernetes-istio/src/main/java/org/springframework/cloud/kubernetes/istio/trace/FeignPropagationRequestInterceptor.java
@@ -1,0 +1,60 @@
+/*
+ * Copyright 2013-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.kubernetes.istio.trace;
+
+import java.util.Arrays;
+import java.util.List;
+
+import feign.RequestInterceptor;
+import feign.RequestTemplate;
+import org.apache.commons.lang3.StringUtils;
+import org.slf4j.Logger;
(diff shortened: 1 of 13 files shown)


geoand:
Very interesting! @salaboy would you like to take a look?

spencergibb:
I think we're duplicating functionality from sleuth, no?

idefav:
> I think we're duplicating functionality from sleuth, no?

This PR is not only enhanced for Hystrix, but also for RestTemplate, Ribbon, Feign, custom thread pool and HTTPClient. The Hystrix thread pool mode does not support the call chain. Here you need to enhance it to support Istio's Headers. pass.

spencergibb:
Sleuth instruments all those things

spring-cloud-kubernetes-istio/src/main/java/org/springframework/cloud/kubernetes/istio/trace/FeignPropagationRequestInterceptor.java
```diff
@@ -0,0 +1,60 @@
...
+ */
+@ConditionalOnClass(RequestInterceptor.class)
+@ConditionalOnProperty(value = "spring.cloud.istio.enabled", matchIfMissing = true)
+public class FeignPropagationRequestInterceptor implements RequestInterceptor {
+
+	private static final Logger LOG = LoggerFactory
+			.getLogger(FeignPropagationRequestInterceptor.class);
+
+	private static List<String> HEADERS = Arrays.asList(IstioConstants.X_REQUEST_ID,
+			IstioConstants.X_B3_TRACEID, IstioConstants.X_B3_SPANID,
```
adriancole: similarly missing "b3"

spring-cloud-kubernetes-istio/src/main/java/org/springframework/cloud/kubernetes/istio/trace/IstioHeaderPropagatoinRequestFilter.java
```diff
@@ -0,0 +1,48 @@
...
+ * be passed.
+ *
+ * @author wuzishu
+ * @see org.springframework.cloud.kubernetes.istio.trace.HeaderPropagationRequestFilter
+ */
+@ConditionalOnProperty(value = "spring.cloud.istio.enabled", matchIfMissing = true)
+public class IstioHeaderPropagatoinRequestFilter extends HeaderPropagationRequestFilter {
+
+	private static List<String> HEADERS = Arrays.asList(IstioConstants.X_REQUEST_ID,
+			IstioConstants.X_B3_TRACEID, IstioConstants.X_B3_SPANID,
```
adriancole: "b3"

spencergibb:
I'm going to close this. If sleuth is missing anything, it should be added there.

Issue #402, opened 2019-06-02
ConfigMap change does not trigger bean refresh


martin-schaefer:
I am doing some training with spring-cloud-kubernetes on Red Hat minishift. I managed to define a ConfigMap which is correctly found and the configured message ${service.message} is shown..

However, when I change the value in the ConfigMap, the RestController in the @RefreshScope seems not to be refreshed and the message stays the same. Only a restart of the app updates the message.

According to the documentation I assumed this should work. Did I miss some additional step or is this not working yet?

```java
@Controller
@RefreshScope
public class RestController {

	private final String message;

	public RestController(@Value("${service.message}") String message) {
		this.message = message;
	}

	/**
	 * Say hello to earth
	 * 
	 * @return First word from the moon
	 * 
	 */
	@GetMapping("/hello")
	@ResponseBody
	public String hello() {
		return message + " (" + LocalDateTime.now().toString() + ")";
	}
}
```
The ConfigMap

```yaml
apiVersion: v1
data:
  application.yml: |-
    service:
      message: Say hello to the world  
kind: ConfigMap
metadata:
  creationTimestamp: '2019-06-02T17:35:24Z'
  name: openshift-spring-boot-service
  namespace: dsp-dev
  resourceVersion: '67065'
  selfLink: /api/v1/namespaces/dsp-dev/configmaps/openshift-spring-boot-service
  uid: cd7a8217-855c-11e9-8dc5-00155d010b07
```

geoand:
I would suggest you take a look at [this](https://github.com/spring-cloud/spring-cloud-kubernetes/tree/master/spring-cloud-kubernetes-integration-tests/simple-configmap) integration test. My guess is that you probably haven't enabled restart.

martin-schaefer:
Thanks for the hint. I activated the config reload like this:

```yaml
spring:
  cloud:
    kubernetes:      
      reload:
        enabled: true
```
But then I received the following error on startup:
```
Description:
Parameter 2 of method configurationUpdateStrategy in org.springframework.cloud.kubernetes.config.reload.ConfigReloadAutoConfiguration$ConfigReloadAutoConfigurationBeans required a bean of type 'org.springframework.cloud.context.restart.RestartEndpoint' that could not be found.
The following candidates were found but could not be injected:
[...]

PR #418, opened 2019-06-27
Restart endpoint is not needed with refresh strategy gh-402

Haybu:


ryanjbaxter:
How is this related to #422?

Issue #420, opened 2019-07-01
KubernetesProfileEnvironmentPostProcessor should be conditional on spring.cloud.kubernetes.enabled


stefanocke:
Several configuration classes like KubernetesAutoConfiguration and KubernetesDiscoveryClientAutoConfiguration are condtional on spring.cloud.kubernetes.enabled.
This allows to "switch off" kubernetes client completely in non-k8s environments.

But, KubernetesProfileEnvironmentPostProcessor is currently not conditional.
So, it always runs and for example creates a DefaultKubernetesClient which in turn can lead to k8s-related warnings,  for example in **io.fabric8.kubernetes.client.Config.tryServiceAccount**

In my case, I have seen the following:
```
Error reading service account token from: [/var/run/secrets/kubernetes.io/serviceaccount/token]. Ignoring.
```

It would be nice to have KubernetesProfileEnvironmentPostProcessor  conditional on spring.cloud.kubernetes.enabled, too,  to avoid that kind of warnings in non-k8s enironments.

geoand:
Thanks for opening the issue.

Can you please provide the entire log line? 
I am thinking perhaps we need to change the logging level since I don't think we can't really turn off the post processor

stefanocke:
@geoand , the complete message is the one above.  It is from  io.fabric8.kubernetes.client.Config.tryServiceAccount.

I think, the logging level there is quite appropriate.  (And btw, it is not in spring-cloud-kubernetes anyway.)  
Why not to turn off the PostProcessor in a situation where Kubernetes Client shall be switched off completely?

geoand:
@stefanocke yeah, makes sense. I opened #421

PR #421, opened 2019-07-01
Make Post Processor honor spring.cloud.kubernetes.enabled

geoand:
Fixes: #420


stefanocke:
@geoand , thank you. But shouldn't it be the other way around?

That is:
```
		final String enabledStr = environment.getProperty("spring.cloud.kubernetes.enabled", "true");
		if ("false".equals(enabledStr.toLowerCase())) {
			return;
		}
```

instead of 

```
		final String enabledStr = environment.getProperty("spring.cloud.kubernetes.enabled", "false");
		if ("true".equals(enabledStr.toLowerCase())) {
			return;
		}
```

The default for this property is true.
And the PostProcessor should return immediately,  if it is set to false.

geoand:
@stefanocke you are absolutely right! It slipped through the cracks... I will push a fix

spencergibb:
Already swapped in 599f96a7b3a85514033b08d18e1cf3f395c4a64a

geoand:
Thanks @spencergibb !

PR #425, opened 2019-07-10
restart endpoint does not need to be enabled when refresh is used by …

Haybu:
…default fixes gh-402

PR is resubmitted as the earlier one was polluted with other commits.

### DIFF
--- a/spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/reload/ConfigReloadAutoConfiguration.java
+++ b/spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/reload/ConfigReloadAutoConfiguration.java
@@ -40,6 +40,7 @@ import org.springframework.context.annotation.Configuration;
 import org.springframework.core.env.AbstractEnvironment;
 import org.springframework.scheduling.annotation.EnableAsync;
 import org.springframework.scheduling.annotation.EnableScheduling;
+import org.springframework.util.Assert;
 
 /**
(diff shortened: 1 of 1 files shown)


spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/reload/ConfigReloadAutoConfiguration.java
```diff
@@ -111,32 +112,34 @@ public ConfigurationChangeDetector propertyChangeWatcher(
...
 		@ConditionalOnMissingBean
 		public ConfigurationUpdateStrategy configurationUpdateStrategy(
-				ConfigReloadProperties properties, ConfigurableApplicationContext ctx,
-				RestartEndpoint restarter, ContextRefresher refresher) {
+			ConfigReloadProperties properties, ConfigurableApplicationContext ctx,
+			@Autowired(required = false) RestartEndpoint restarter,
+			ContextRefresher refresher) {
 			switch (properties.getStrategy()) {
 			case RESTART_CONTEXT:
+				Assert.notNull(restarter, "Restart endpoint is not enabled");
```
ryanjbaxter: this is the main change right?  it is hard to see with the formatting changes

Haybu: one change, an autowired injected param noted below that made optional.

ryanjbaxter:
@Haybu Can you submit the fix against the 1.0.x stream so we can get it in greenwich?

PR #426, opened 2019-07-10
Feature/multi secrets

Haybu:
adding ability to source in multiple secrets. This is made possible by the `spring.cloud.kubernetes.secrets.sources` list. For example one could define the following Secrets:

```
spring:
  application:
    name: some-app-name
  cloud:
    kubernetes:
      enabled: true
      secrets:
        enabled: true
        enable-api: true
        namespace: default-ns
        sources:
          # looks up secret named name1 in namespace default-ns with label env='foo'
          - name: name1
            labels:
              - env: 'foo'
        # looks up secret named name2 in namespace another-ns with label env='bar'
          - name: name2
            namespace: another-ns
            labels:
              - env: 'bar'
```


ryanjbaxter:
I am wondering about accessing secrets from another namespace, could there be permissions issues with this?  @geoand @iocanel @salaboy any thoughts?

geoand:
There certainly could be permission issues, but it would be up the Kubernetes administrator to enforce them.
Secret is just like any Kubernetes Resource, to read it you need to have the proper access.

salaboy:
@ryanjbaxter @geoand yeah.. and I mentioned something related to this in an issue before. I think that the spring cloud kubernetes libraries should be namespace constrained. In the sense that if the app is running in a namespace, that should be the default namespace set in the Kubernetes-clients for all the calls. Unless you are using directly the Kubernetes APIs you shouldn't be able to change the namespace where the information is being read.

salaboy:
For this case, where you can explicitly set which namespace you want to retrieve your secrets should be ok, but again, this will not solve the security configurations that you will need to set up into the kubernetes side to make that happen.

ryanjbaxter:
Thanks @geoand and @salaboy so today, there is nothing stopping an SCK8S app from reading secrets from another namespace as long as the admin has enabled it?  That may not be the case in the future though, depending on which direction we take it.

geoand:
@ryanjbaxter that is correct and from my POV, allowing that to happen is the proper way to go. Why should we limit something in SCK when k8s itself allows is?

geoand:
I agree completely with the observation @salaboy made, it's just that I have the opposite opinion :P. IMHO we shouldn't try and limit what can be done in this project

salaboy:
@geoand and that is a perfectly valid opinion as well.. now.. my point was.. as soon as the defaults are ok, we shouldn't restrict users to do more things. But I am worried of not having a consistent default option for such things. I believe that good "default" values and Configurations are what makes spring boot the best and easier framework out there.

geoand:
Yeah, I definitely see your point @salaboy.

But in my experience Spring makes the sensible default options without limiting the users if they want to use other options.
Also one could argue here that the default option is that the user leaves out the namespace - resulting in the current namespace being used :)

ryanjbaxter:
> But in my experience Spring makes the sensible default options without limiting the users if they want to use other options.

Yes that is right.

I guess this is a separate topic, currently we do not have these restrictions around accessing things like config maps or secrets etc.  So I see no problem with the PR the way it is.  Whether we do something different in the future is probably a topic that needs to be discussed separately.

Issue #432, opened 2019-07-22
Spring Cloud K8S Build Fails With JDK 11


ryanjbaxter:
```
14:25:21 [ERROR] Failed to execute goal on project spring-cloud-kubernetes-core: Could not resolve dependencies for project org.springframework.cloud:spring-cloud-kubernetes-core:jar:1.1.0.BUILD-SNAPSHOT: Could not find artifact com.sun:tools:jar:1.7 at specified path /opt/jenkins/data/tools/hudson.model.JDK/jdk11/jdk-11.0.2/../lib/tools.jar -> [Help 1]
```

geoand:
Ah yes, we'll need to exclude the that "dependency"

spencergibb:
@geoand which one it is and where is it pulled in? :-)

geoand:
I am pretty sure it's a transitive of the kubernetes-client.

Not at my machine now, but I can take a look tomorrow. Otherwise @iocanel would definitely know exactly where to look

PR #433, opened 2019-07-23
Exclude tools java from dependencies

geoand:
Needed in order to build the project in Java 11

Fixes: #432


ryanjbaxter:
@geoand I just merged this but now i occurs to me that we might not need to add these exclusions when only building with java 11, did you try this with java 8 as well?

geoand:
@ryanjbaxter Yeah I tried it with Java 8 as well and it's all good :)

Issue #385, opened 2019-03-30
KubernetesDiscoveryClientAutoConfiguration should enable Scheduling


germm:
The scheduled Method `catalogServicesWatch()` in `KubernetesCatalogWatch` is not executed if the Annotation `@EnableScheduling` is not explicitly set by the user. This was alerady part of the discussion in #254 

IMHO, the `KubernetesDiscoveryClientAutoConfiguration` should enable Scheduling unless kubernetesCatalogWatch is disabled by the user.

ryanjbaxter:
We don't have any precedence for adding `@EnableScheduling` as part of auto configuration in other Spring Cloud projects or in Spring Boot itself as far as we can see.  If there are other `@Scheduled` annotations in the app I wouldn't expect those to automatically be scheduled just because Spring Cloud Kubernetes is on the classpath.  We should however at the very least document the fact that it is needed in order to make the functionality work.

germm:
Many thanks for your answer. Mentioning the fact in the documentation makes the situation better.

I understand your reason for not automatically enable scheduling but a search for `EnableScheduling` shows that this is not consequently handled this way. The following auto configurations do enable scheuduling:
- [ConfigReloadAutoConfiguration](https://github.com/spring-cloud/spring-cloud-kubernetes/blob/master/spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/reload/ConfigReloadAutoConfiguration.java)
- [HystrixStreamAutoConfiguration](https://github.com/spring-cloud/spring-cloud-netflix/blob/master/spring-cloud-netflix-hystrix-stream/src/main/java/org/springframework/cloud/netflix/hystrix/stream/HystrixStreamAutoConfiguration.java)

spencergibb:
I'll argue those are the exceptions and should probably be removed.

ryanjbaxter:
`ConfigReloadAutoConfiguration` is only enabled if `spring.cloud.kubernetes.reload.enabled` is set to true so it is not on by default.  We should probably remove `@EnableScheduling` from that class as well.

joshiste:
Did you thought about not relying on the `@Scheduled` at all? Imho a single threaded task scheduler for the `KubernetesCatalogWatch` would do the job and doesn't require any other condfigurations, or further steps from the developer... and no update to the documentation...

PR #434, opened 2019-07-23
Extracted catalog watch auto-configuration.

TYsewyn:
Needed in order to have that bean running in the correct application context when SC Config discovery is enabled.
Also fixes a bug where Kubernetes discovery was enabled when discovery was disabled with `spring.cloud.kubernetes.enabled=false`.
Relates to spring-cloud/spring-cloud-config#1229

### DIFF
--- a/docs/src/main/asciidoc/discovery-client.adoc
+++ b/docs/src/main/asciidoc/discovery-client.adoc
@@ -55,7 +55,3 @@ spring.cloud.kubernetes.discovery.enabled=false
 
 Some Spring Cloud components use the `DiscoveryClient` in order to obtain information about the local service instance. For
 this to work, you need to align the Kubernetes service name with the `spring.application.name` property.
-
-Spring Cloud Kubernetes can also watch the Kubernetes service catalog for changes and update the
-`DiscoveryClient` implementation accordingly.  In order to enable this functionality you need to add
-`@EnableScheduling` on a configuration class in your application.
--- a/docs/src/main/asciidoc/discovery-client.adoc
+++ b/docs/src/main/asciidoc/discovery-client.adoc
@@ -55,3 +55,7 @@ spring.cloud.kubernetes.discovery.enabled=false
 
 Some Spring Cloud components use the `DiscoveryClient` in order to obtain information about the local service instance. For
 this to work, you need to align the Kubernetes service name with the `spring.application.name` property.
+
+Spring Cloud Kubernetes can also watch the Kubernetes service catalog for changes and update the
+`DiscoveryClient` implementation accordingly.  In order to enable this functionality you need to add
+`@EnableScheduling` on a configuration class in your application.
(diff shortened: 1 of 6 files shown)


docs/src/main/asciidoc/discovery-client.adoc
```diff
@@ -55,7 +55,3 @@ spring.cloud.kubernetes.discovery.enabled=false
 
 Some Spring Cloud components use the `DiscoveryClient` in order to obtain information about the local service instance. For
 this to work, you need to align the Kubernetes service name with the `spring.application.name` property.
-
-Spring Cloud Kubernetes can also watch the Kubernetes service catalog for changes and update the
```
spencergibb: Why is this removed?

spring-cloud-kubernetes-discovery/src/main/java/org/springframework/cloud/kubernetes/discovery/KubernetesCatalogWatchAutoConfiguration.java
```diff
@@ -0,0 +1,49 @@
...
+import org.springframework.context.annotation.Configuration;
+import org.springframework.scheduling.annotation.EnableScheduling;
+
+/**
+ * Auto configuration for catalog watcher.
+ *
+ * @author Tim Ysewyn
+ */
+@Configuration
+@EnableScheduling
```
spencergibb: We decided not to do this.

TYsewyn: Do you remember why? This is the only service registry implementation that needs that annotation explicitly.

spencergibb: We chose to document it instead.

ryanjbaxter: See https://github.com/spring-cloud/spring-cloud-kubernetes/issues/385

TYsewyn: Thanks for the input! So what if we do it like it's done in the [ConsulCatalogWatchAutoConfiguration](https://github.com/spring-cloud/spring-cloud-consul/blob/master/spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulCatalogWatchAutoConfiguration.java)?
Then we don't need to explicitly add `@EnableScheduling`, and it will only be created when the watcher is enabled with `spring.cloud.kubernetes.discovery.catalog-services-watch.enabled`.

Maybe not for this PR, but could be a nice improvement for the next minor release.

spencergibb: Yup, not for this PR.

PR #456, opened 2019-08-27
spring.application.name has no effect on service registry name

### LINKED ISSUE #354: spring.application.name has no effect on service registry name
With other service discovery implementations `spring.application.name` is used as the name registered with the service discovery service.  This is not true in sc-kubernetes, we should document this.

medmedchiheb:
adding a note about spring.application.name in sc-k8s 

Fixes #354


medmedchiheb:
ping @ryanjbaxter for review

Issue #473, opened 2019-10-05
The problem occurs when you have two or more namespaces, where the service name is the same


claitonp:
The problem occurs when you have two or more namespaces, where the service name is the same.
Example: service 'price' is in namespaces ns1 and ns2 in the same cluster kubernetes.

xxx-service.yaml
-------------
```
metadata:
   namespace: ns1
   name: price
   labels:
       app: price
       group.department: dept1
```

yyy-service.yaml
-----------------
```
metadata:
  namespace: ns2
  name: price
  labels:
      app: price
      group.department: dept2
	  
```
app-discovery.yaml
------------	  
```
spring:
  cloud:
    kubernetes:
	discovery:
	  service-labels:
	  - group.department: dept1
```
In this case, the current code returns the endpoints of both namespaces, which is not expected.
To resolve this issue you have to apply the service-labels filter, **.withLabels(serviceLabels).**

spring-cloud-kubernetes, version:1.0.4.BUILD-SNAPSHOT	

```
package org.springframework.cloud.kubernetes.discovery;
public class KubernetesDiscoveryClient implements DiscoveryClient {
  ...
	@Override
	public List<ServiceInstance> getInstances(String serviceId) {
		...
		// bug
		// List<Endpoints> endpointsList = this.properties.isAllNamespaces()
		//	? this.client.endpoints().inAnyNamespace().withField("metadata.name", serviceId).list().getItems()
		//	: Collections.singletonList(this.client.endpoints().withName(serviceId).get());

		//adjusted
		Map<String, String> serviceLabels = properties.getServiceLabels();
		List<Endpoints> endpointsList = properties.isAllNamespaces() ?
				client.endpoints().inAnyNamespace().withField("metadata.name", serviceId).withLabels(serviceLabels).list().getItems()
				: Collections.singletonList(client.endpoints().withName(serviceId).get());
	}			
}
```

spencergibb:
Please say what is happening. "Does not work correctly" isn't helpful at all.

claitonp:
Hi,

When using serviceLabels, it is not filtering correctly:
```
  spring:
    cloud:
      kubernetes:
          discovery:
             service-labels:
```
To work properly, you must add: .withLabels(serviceLabels)

claitonp:
In the class KubernetesDiscoveryClient.java , method getInstance(String serviceId).
To work properly you must add the code in bold.

**Map<String, String> serviceLabels = properties.getServiceLabels();**
[...]

PR #503, opened 2019-12-18
Add namespace information to service instance metadata

m-kay:
As mentioned in issue #473 the namespace should be reflected in the serviceId when using discovery with all-namespaces set to true.

### DIFF
--- a/spring-cloud-kubernetes-discovery/src/main/java/org/springframework/cloud/kubernetes/discovery/KubernetesServiceInstance.java
+++ b/spring-cloud-kubernetes-discovery/src/main/java/org/springframework/cloud/kubernetes/discovery/KubernetesServiceInstance.java
@@ -31,6 +31,11 @@ import org.springframework.cloud.client.ServiceInstance;
  */
 public class KubernetesServiceInstance implements ServiceInstance {
 
+	/**
+	 * Key of the namespace metadata.
+	 */
+	public static final String NAMESPACE_METADAT_KEY = "k8s_namespace";
+
 	private static final String HTTP_PREFIX = "http";
 
 	private static final String HTTPS_PREFIX = "https";
@@ -116,6 +121,10 @@ public class KubernetesServiceInstance implements ServiceInstance {
 	@Override
 	public String getScheme() {
 		return isSecure() ? HTTPS_PREFIX : HTTP_PREFIX;
+	}
+
+	public String getNamespace() {
+		return this.metadata != null ? this.metadata.get(NAMESPACE_METADAT_KEY) : null;
 	}
 
 }
--- a/spring-cloud-kubernetes-discovery/src/main/java/org/springframework/cloud/kubernetes/discovery/KubernetesServiceInstance.java
+++ b/spring-cloud-kubernetes-discovery/src/main/java/org/springframework/cloud/kubernetes/discovery/KubernetesServiceInstance.java
@@ -34,7 +34,7 @@ public class KubernetesServiceInstance implements ServiceInstance {
 	/**
 	 * Key of the namespace metadata.
 	 */
-	public static final String NAMESPACE_METADAT_KEY = "k8s_namespace";
+	public static final String NAMESPACE_METADATA_KEY = "k8s_namespace";
 
 	private static final String HTTP_PREFIX = "http";
 
@@ -124,7 +124,7 @@ public class KubernetesServiceInstance implements ServiceInstance {
 	}
 
 	public String getNamespace() {
-		return this.metadata != null ? this.metadata.get(NAMESPACE_METADAT_KEY) : null;
+		return this.metadata != null ? this.metadata.get(NAMESPACE_METADATA_KEY) : null;
 	}
 
 }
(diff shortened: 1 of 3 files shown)


m-kay:
Correct me if I'm wrong, but I think is not possible to access the service by only its service id from outside its own namespace. See: https://kubernetes.io/docs/concepts/services-networking/service/#dns

Haybu:
@m-kay - that is true when a service name/id is used directly and you depend on k8s platform to resolve. But the idea here is that the discovery client uses a service id to interrogate the platform server for the service endpoints, hence providing a list of pods ip:port to pick from and hit.

m-kay:
But how can you be sure that two services are really the same? It's totally valid to have two different services with the same name in different namespaces. In my case I have different applications which are structured the same way in different namespaces. Each application has the same service for example "gateway". However each gateway does something different because it's specific for the application in that namespace.

spencergibb:
We'll need to have a discussion. What you've described and what Haytham and Ryan have said are both valid, but at odds with each other. We'll need to see if we can reconcile the two. I don't know if it's in discovery client or in loadbalancer or both.

GianniGiglio:
Hi am looking for a similar solution where services can be deployed on multiple namespaces (1 namespace eq customer instance). 

I still have a multitenant application using the kuberentesDiscoveryClient and I would need to route the request to multiple namespaces.

I was wondering why not add the namespace to kubernetesServiceInstance instead of in the serviceId?

ryanjbaxter:
After discussing this with the team today we think the best solution at this time is to only take the portion of this PR which puts the namespace in the service metadata
https://github.com/spring-cloud/spring-cloud-kubernetes/pull/503/files#diff-bf6bb75a365ae3173c5d845358dfdc85R116

From there clients should be able to determine which instance of the service they want to target.

@m-kay if you are interested in modifying the PR to just include those changes please let me know

GianniGiglio:
> After discussing this with the team today we think the best solution at this time is to only take the portion of this PR which puts the namespace in the service metadata
> https://github.com/spring-cloud/spring-cloud-kubernetes/pull/503/files#diff-bf6bb75a365ae3173c5d845358dfdc85R116
> 
> From there clients should be able to determine which instance of the service they want to target.
> 
> @m-kay if you are interested in modifying the PR to just include those changes please let me know

If the service instance would contain the namespace as a field, Spring cloud gateway could for instance forward to the correct namespace when using kubernetes services dns.

example config:
spring:
  cloud:
    gateway:
      discovery:
        locator:
          url-expression: "'http://'+serviceId+' + 'namespace'+''.svc.cluster-domain.example'

m-kay:
@ryanjbaxter yes I would like to change the pull requests. However the code you mentioned in the diff does not include the namespace in the service metadata but was to filter service instances in the k8s client. As I understood the idea is to include the namespace information in the ServiceInstance, right?

@GianniGiglio I could include a getter method in the KubernetesServiceInstance which returns the information from the metadata, that should also do the trick. However you will still have the problem that the service used can be from any namespace because the discovery client will return an unfiltered list of all instances of the service from all namespaces.

GianniGiglio:
> @ryanjbaxter yes I would like to change the pull requests. However the code you mentioned in the diff does not include the namespace in the service metadata but was to filter service instances in the k8s client. As I understood the idea is to include the namespace information in the ServiceInstance, right?
> 
> @GianniGiglio I could include a getter method in the KubernetesServiceInstance which returns the information from the metadata, that should also do the trick. However, you will still have the problem that the service which is used can be from any namespace because the discovery client will return an unfiltered list of all instances of the service from all namespaces.

That depends on how the Kubernetes client is configured I think. At least your proposed solution is how I worked around it by over-writing the kubernetesServiceInstance and populate the namespace based on the metadata.

this.kubernetesClient.endpoints().inNamespace("default") --> will only return the endpoints for the default namespace
this.kubernetesClient.endpoints().inAnyNamespace() --> will return all enpoints in all namespaces.

endpoints.getMetadata().getNamespace() should give you the namespace.

spring-cloud-kubernetes-discovery/src/main/java/org/springframework/cloud/kubernetes/discovery/KubernetesServiceInstance.java
```diff
@@ -31,6 +31,11 @@
  */
 public class KubernetesServiceInstance implements ServiceInstance {
 
+	/**
+	 * Key of the namespace metadata.
+	 */
+	public static final String NAMESPACE_METADAT_KEY = "k8s_namespace";
```
ryanjbaxter, line 37: I dont think we need to prefix the key with `k8s` since this is Kubernetes specific

m-kay: Ah right, the whole metadata part is Kubernetes specific. I removed the prefix now.

m-kay:
Can we get this in the next 1.1.x release by any chance or will it only make it in the next major release?

spring-cloud-kubernetes-discovery/src/main/java/org/springframework/cloud/kubernetes/discovery/KubernetesServiceInstance.java
```diff
@@ -31,6 +31,11 @@
  */
 public class KubernetesServiceInstance implements ServiceInstance {
 
+	/**
+	 * Key of the namespace metadata.
+	 */
+	public static final String NAMESPACE_METADAT_KEY = "namespace";
```
spencergibb, line 37: `NAMESPACE_METADAT_KEY` -> `NAMESPACE_METADATA_KEY` (metadata is missspelled).

Issue #489, opened 2019-11-21
Using ExternalName service type with KubernetesDiscoveryClient


ArtyomGabeev:
Imagine you have a setup when services running in K8S need to access services running outside.

I want to create an ExternalName service:
```yaml
apiVersion: v1
kind: Service
metadata:
  name: my-service
  namespace: prod
spec:
  type: ExternalName
  externalName: my.service.example.com
```
After that I should be able to resolve service call http://my-service/doCall into http://my.service.example.com/doCall

So instead of relying only on endpoints in KubernetesDiscoveryClient, check if service type is ExternalName, then return single service instance with hostname: `my.service.example.com`.

Does it make sense to contribute this enhancement?

guilhermecgs:
very much needed feature

is this related?
https://github.com/spring-cloud/spring-cloud-kubernetes/issues/344

michaelaplante:
I am using Spring Cloud Hoxton.RELEASE (i.e., with Spring Cloud Kubernetes 1.1.0.RELEASE) and the following configuration is not working, so I don't think what you're asking for is covered by  #344

ExternalName service:
```apiVersion: v1
kind: Service
metadata:
  name: my-service
  namespace: my-namespace
spec:
  type: ExternalName
  externalName: my.service.example.com
```
My client application is using ribbon with : `@FeignClient(name = my-service)`

Spring Sloud Kubernetes Ribbon config is:  `spring.cloud.kubernetes.ribbon.mode : SERVICE`

KubernetesServicesServerList   reports that it finds the service, but no ribbons: 
```o.s.b.f.s.DefaultListableBeanFactory     : Creating shared instance of singleton bean 'ribbonRule'
o.s.b.f.s.DefaultListableBeanFactory     : Autowiring by type from bean name 'ribbonRule' via factory method to bean named 'ribbonClientConfig'
c.netflix.config.ChainedDynamicProperty  : Flipping property: my-service.ribbon.ActiveConnectionsLimit to use NEXT property: niws.loadbalancer.availabilityFilteringRule.activeConnectionsLimit = 2147483647
o.s.b.f.s.DefaultListableBeanFactory     : Creating shared instance of singleton bean 'ribbonPing'
o.s.b.f.s.DefaultListableBeanFactory     : Autowiring by type from bean name 'ribbonPing' via factory method to bean named 'ribbonClientConfig'
[...]

PR #512, opened 2020-01-08
Add support for ExternalName service type

ArtyomGabeev:
https://github.com/spring-cloud/spring-cloud-kubernetes/issues/489

### DIFF
--- a/spring-cloud-kubernetes-discovery/src/main/java/org/springframework/cloud/kubernetes/discovery/KubernetesDiscoveryClient.java
+++ b/spring-cloud-kubernetes-discovery/src/main/java/org/springframework/cloud/kubernetes/discovery/KubernetesDiscoveryClient.java
@@ -33,6 +33,7 @@ import io.fabric8.kubernetes.client.KubernetesClient;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
+import org.springframework.cloud.client.DefaultServiceInstance;
 import org.springframework.cloud.client.ServiceInstance;
 import org.springframework.cloud.client.discovery.DiscoveryClient;
 import org.springframework.expression.Expression;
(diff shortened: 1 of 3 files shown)


ArtyomGabeev:
Having a service:
```yaml
apiVersion: v1
kind: Service
metadata:
  name: my-service
  namespace: prod
spec:
  type: ExternalName
  externalName: my.service.example.com
```
We should be able to resolve service call http://my-service/doCall into http://my.service.example.com/doCall

Notes:
Because there is not information about secure/port, I left them false/0.

So it's up to client to specify schema http vs https, e.g: 
http**s**://my-service/doCall should become http**s**://my.service.example.com/doCall

Regrading port, I've checked LB Utils and ports will be 80/443.

digows:
I've achieved that by changing the KubernetesServicesServerList.

`@Override
	public List<Server> getUpdatedListOfServers()
	{
		final List<Server> result = new ArrayList<>();

		final Service service = StringUtils.isNotBlank(this.namespace) ?
				this.client.services().inNamespace(this.namespace).withName(this.serviceId).get() :
				this.client.services().withName(this.serviceId).get();

		if ( service != null )
		{
			if ( LOG.isDebugEnabled() )
			{
				LOG.debug("Found Service[" + service.getMetadata().getName() + "]");
			}

			final String host;
			//is it a external service?
			if ( service.getSpec().getExternalName() != null )
			{
				host = service.getSpec().getExternalName();
			}
			//else, it will use the internal domain
			else
			{
				host = this.concatServiceFQDN(service);
			}

			if ( service.getSpec().getPorts().isEmpty() )
			{
				result.add(new Server(host, 80));
			}
			else if ( service.getSpec().getPorts().size() == 1 )
			{
				result.add(new Server(host, service.getSpec().getPorts().get(0).getPort()));
			}
			else
			{
				for ( ServicePort servicePort : service.getSpec().getPorts() )
				{
					if ( Utils.isNotNullOrEmpty(this.portName) || this.portName.endsWith(servicePort.getName()) )
					{
						result.add(new Server(host, servicePort.getPort()));
					}
				}
			}
		}

		if ( result.isEmpty() )
		{
			LOG.warn(String.format("Did not find any service in ribbon in namespace [%s] for name [%s] and portName [%s]", this.namespace, this.serviceId, this.portName));
		}
		return result;
	}`

ArtyomGabeev:
KubernetesServicesServerList is used by Ribbon, which is in maintenance mode.

Proposed changes should work both for spring cloud load balancer as well.

mindcrime:
Hey, just wondering if anybody could/would comment on if there is intention to merge this at some point? Just curious as we have a scenario where we really need ExternalName services to work with Ribbon as well.

wind57:
this is my next thing to take care of, but for the initial phase, I'll add such support for the fabric8 client only (there are other features that will be ported to k8s client in the nearest future, this one included).

imo, such support for `ExternalName` services should be allowed only by opting into it via a property. The first reason is compatibility: we were not returning such services until now, so we should keep it that way; unless you want this explicitly. The second reason is that this will introduce more calls to the api-server, so having it optional is only reasonable to me.

Otherwise I will present a proposal PR in the upcoming days.

wind57:
@ryanjbaxter this can be closed now. thank you.

Issue #465, opened 2019-09-10
DiscoveryClient allNamespaces Property Not Used When Querying Services


michaelmcfadyen:
This relates to #428 .

There was a change made to configure the discovery client to search across all namespaces based on a property. Inside `KubernetesDiscoveryClient,getInstances` the method call which queries the endpoints was updated to use this new property
```
this.properties.isAllNamespaces()
				? this.client.endpoints().isAllNamespaces()
						.withField("metadata.name", serviceId).list().getItems()
				: Collections
						.singletonList(this.client.endpoints().withName(serviceId).get());
```

However, the same change was not applied later in the same method when the services are queried.
The current code is:
```
this.client.services().withName(serviceId).get();
```

As a result, when trying to use the `isAllNamespaces` property we are consistently getting null pointer exceptions.

It seems like the call to query the services should also use the `isAllNamespaces` method when the property is set.

geoand:
Hi,

Would you like to a submit a pull request with your proposed fix?

michaelmcfadyen:
Tested out 1.0.4-BUILD-SNAPSHOT and everything is working as expected. Good work!

ryanjbaxter:
Fixed via d9bad61

PR #448, opened 2019-08-08
Feature/all ns svc

Haybu:
Enabling `DiscoveryClient` from all namespaces based on an opt-in property: `spring.cloud.kubernetes.discovery.all-namespaces=true` fixes gh-428 and also relates to gh-163

Issue #428, opened 2019-07-19
Enable DiscoveryClient from all namespaces based on property opt in


Incarnation-p-lee:
As we know, the package
```pom
<dependency>
    <groupId>org.springframework.cloud</groupId>
    <artifactId>spring-cloud-starter-kubernetes</artifactId>
</dependency>
```
will discovery all services from k8s, and we can configure the namespace from property
```
spring:
  cloud:
    kuberenetes:
      client:
        namespace: ${my-namespace}
```

But how can I discovery services from all namespace ?

geoand:
Hi,

You could define a custom `KubernetesClientServicesFunction` bean that would look something like this:

```
@Configuration
public class KubernetesClientServicesFunctionConfiguration {
 
   @Bean
   public KubernetesClientServicesFunction servicesFunction() {
      return client.services().inAnyNamespace()
   }
}
```

Incarnation-p-lee:
Thanks a lot. but what is this `client` here?

geoand:
Good catch! I updated the answer

Incarnation-p-lee:
seems need add `@Primary` for this bean, let me have a try again,

Incarnation-p-lee:
seems not working, I cannot see any service now.

Incarnation-p-lee:
looks some error like
`The bean 'servicesFunction', defined in class path resource [org/springframework/cloud/kubernetes/discovery/KubernetesDiscoveryClientAutoConfiguration.class], could not be registered. A bean with that name has already been defined in class path resource [com/microsoft/azure/spring/service/config/CloudGatewayConfiguration.class] and overriding is disabled.`

geoand:
Can you then add `@Primary`? 

The reason I ask to check the above configuration is because that is essentially what happens by default. If you add it and continue to see no services then there is something else going on

Incarnation-p-lee:
Yes, I added `@Primary`, it looks that bean added `@Primary` too.

ryanjbaxter:
@Incarnation-p-lee can you open an issue so we can update the `servicesFunction` bean name to be more specific, it is too generic (obviously).

I think this would be a nice enhancement that we could add so based on a property we create the bean for you.

Incarnation-p-lee:
how about `Support kubernetes discovery client discovery services from all namespace` as feature request ?

spencergibb:
> I think this would be a nice enhancement that we could add so based on a property we create the bean for you.

[...]

PR #544, opened 2020-04-06
Fix scheduled discovery when allNamespaces property is true

tzoratto:
Relates to #428 #448 #465 

**Type of request**

Bug fix

**Behavior**

When 

- the property `spring.cloud.kubernetes.discovery.all-namespaces` is true 
- and `@EnableScheduling` has been set

the service/pod discovery never see any addition/deletion of pods for a given service in a different namespace that the one the spring boot admin server is in.

**How to reproduce**

1. In a namespace `A`, deploy a spring boot application (deployment and service) with one replica
2. In a namespace `B`, deploy a spring boot admin server with the following annotations (as per the documentation of this project) :
`@SpringBootApplication
@EnableAdminServer
@EnableDiscoveryClient
@EnableScheduling`, the property `spring.cloud.kubernetes.discovery.all-namespaces` set to true (and the required rbac configuration) 
3. You should see your spring boot application properly reported in spring boot admin server, with 1 instance
4. Scale up your spring boot application (let's say 2 replicas)
5. Even after 30 secondes (the default scheduled interval), spring boot admin server still shows 1 spring boot application with 1 instance

**Expected behavior**

When these conditions are met we expect to see the dynamically updated list of instances in spring boot admin server without restarting it.

Please let me know if there is anything wrong with this patch, I had never used this lib before this morning so I might be missing something obvious. Adding `inAnyNamespace()` definitely fixes the issue though

Thanks

### DIFF
--- a/spring-cloud-kubernetes-discovery/src/main/java/org/springframework/cloud/kubernetes/discovery/KubernetesCatalogWatch.java
+++ b/spring-cloud-kubernetes-discovery/src/main/java/org/springframework/cloud/kubernetes/discovery/KubernetesCatalogWatch.java
@@ -45,12 +45,16 @@ public class KubernetesCatalogWatch implements ApplicationEventPublisherAware {
 
 	private final KubernetesClient kubernetesClient;
 
+	private final KubernetesDiscoveryProperties properties;
+
 	private final AtomicReference<List<String>> catalogEndpointsState = new AtomicReference<>();
 
 	private ApplicationEventPublisher publisher;
 
-	public KubernetesCatalogWatch(KubernetesClient kubernetesClient) {
+	public KubernetesCatalogWatch(KubernetesClient kubernetesClient,
+			KubernetesDiscoveryProperties properties) {
 		this.kubernetesClient = kubernetesClient;
+		this.properties = properties;
 	}
 
 	@Override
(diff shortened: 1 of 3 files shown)


ryanjbaxter:
Is Spring Boot Admin using `HeartBeatEvent` in some way to update the list of services?

cc: @Haybu

tzoratto:
apparently yes as adding `inAnyNamespace()` here (which only causes a `HeartBeatEvent` to be published) successfully fixes the issue and update the data within spring boot admin server

that's also how it works with discovery based on consul as far as I understand it (https://cloud.spring.io/spring-cloud-consul/2.1.x/multi/multi_spring-cloud-consul-discovery.html)

tzoratto:
any more feedback ? Until this fix is released I have to maintain a custom implementation of KubernetesCatalogWatch to make it work

spencergibb:
Please me patient. The next release isn't scheduled until May.

tzoratto:
of course, I just wanted to know if the fix looks good for you or if I have to change something before it can be merged

tzoratto:
Hello there, I'm still waiting for feedback :)

weliff:
I have the same problem in the description :/

tzoratto:
hi @weliff unfortunately as you can see I have had no feedback at all on this 20 LOC PR for more than 2 months now (I still don't understand why). The only solution for now is to create a custom bean implementing KubernetesCatalogWatch doing what this PR does, I hope you managed to do so 👍

llaszkie:
Hi! We can see it is merged - great! :-) But do you @ryanjbaxter plan also to release it - like soon - in scope of **1.1**?

ryanjbaxter:
Yes our next Hoxton service release will contain this change.  You can find dates here https://github.com/spring-cloud/spring-cloud-release/milestones

PR #554, opened 2020-05-14
Fix value data-type & entry-set order issues of properties bound to Map

moudatsos:
Issue: properties bound to a java.util.Map object do not maintain their
definition order and their values lose their data-type, since they 're
converted to String values.
This fix mainly targets org.springframework.cloud.kubernetes.config
.PropertySourceUtils.PROPERTIES_TO_MAP implementation: Configuration
properties that are bound to a java.util.Map had their value's type
affected by the use of .toString() on each entry's value during Map
entry processing. Moreover, the use of java.util.stream
.Collectors.toMap(java.util.function.Function<? super T,? extends K>,
java.util.function.Function<? super T,? extends U>) when processing
Map-bound properties, resulted in losing the entry insertion order of
the original Map. The fix entails usage of the original entry value and
a Collectors.toMap() variant that maintains the behavior of the method
previously used, yet uses a java.util.LinkedHashMap as the target data
structure for the processed entries. Ramification changes on method
signature and return type have been applied where needed.

### DIFF
--- a/spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/ConfigMapPropertySource.java
+++ b/spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/ConfigMapPropertySource.java
@@ -16,7 +16,7 @@
 
 package org.springframework.cloud.kubernetes.config;
 
-import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
@@ -35,12 +35,14 @@ import org.springframework.util.StringUtils;
 import static org.springframework.cloud.kubernetes.config.PropertySourceUtils.KEY_VALUE_TO_PROPERTIES;
 import static org.springframework.cloud.kubernetes.config.PropertySourceUtils.PROPERTIES_TO_MAP;
 import static org.springframework.cloud.kubernetes.config.PropertySourceUtils.yamlParserGenerator;
+import static org.springframework.cloud.kubernetes.config.PropertySourceUtils.throwingMerger;
 
 /**
  * A {@link MapPropertySource} that uses Kubernetes config maps.
  *
  * @author Ioannis Canellos
  * @author Ali Shahbour
+ * @author Michael Moudatsos
  */
 public class ConfigMapPropertySource extends MapPropertySource {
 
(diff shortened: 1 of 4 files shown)


moudatsos:
The provided fix's patch is compatible with all of master, 1.1.x and 1.0.x branches, with no conflicts. I have verified the correctness of the fix on 1.0.x branch.

geoand:
Hi,

Thanks for this.

It seems like CI is failing:

```
[ERROR] /home/circleci/project/spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/ConfigMapPropertySource.java:38: Wrong order for 'org.springframework.cloud.kubernetes.config.PropertySourceUtils.throwingMerger' import. [ImportOrder]
```

moudatsos:
I will look into it, though checkstyle was failing right from the first build I attempted when I downloaded the project, without any change, whatsoever. I had to disabe it from parent pom, in order to be able to work on the real issue. If this is more like a regulatory matter, I could refactor it to a full package reference just to move on a bit quicker. It's a pity to hold back the fix for checkstye reasons. That said, I fully understand the reasons behind checkstyle practices trying to help us all move this forward!

geoand:
Well no one is saying that the fix will be held back, I'm just saying that we need to make Maven (and therefore CI) happy :)

geoand:
Can you please squash the commits?

Thanks

geoand:
LGTM, @spencergibb ?

moudatsos:
A couple of questions:
1. Should I've been coding on master, in the first place or did I do well on creating a feature branch? General GitHub guides instructed the latter.

2. As about "squashing my commits". If create a patch of all changes and apply it on my branch after I reset it to the latest master revision (and then force push), would that do? Or is there some other way that is it's preferred to be done? 

3. This patch is applicable on all versions of spring-cloud-kubernetes-config . Should I implement it separately for each one (e.g. in the project I'm currently working on, we 're using 1.0.x version) and create a subsequent pull-request? Or is there another way? (e.g. maybe you cascade merges between 1.x branches)

Thank you in advance!

spencergibb:
@ryanjbaxter or @Haybu?

ryanjbaxter:
Thanks @moudatsos!

Can you create a PR against the 1.1.x branch and we will merge it forward to master?  The 1.0.x branch is no longer being actively maintained so no need to put the fix there.

PR #556, opened 2020-05-18
Fix value data-type & entry-set order of yml properties bound to Map

moudatsos:
implementing PR #554 on 1.1.x branch

Issue: properties bound to a java.util.Map object do not maintain their
definition order and their values lose their data-type, since they 're
converted to String values.
This fix mainly targets org.springframework.cloud.kubernetes.config
.PropertySourceUtils.PROPERTIES_TO_MAP implementation: Configuration
properties that are bound to a java.util.Map had their value's type
affected by the use of .toString() on each entry's value during Map
entry processing. Moreover, the use of java.util.stream
.Collectors.toMap(java.util.function.Function<? super T,? extends K>,
java.util.function.Function<? super T,? extends U>) when processing
Map-bound properties, resulted in losing the entry insertion order of
the original Map. The fix entails usage of the original entry value and
a Collectors.toMap() variant that maintains the behavior of the method
previously used, yet uses a java.util.LinkedHashMap as the target data
structure for the processed entries. Ramification changes on method
signature and return type have been applied where needed.

### DIFF
--- a/spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/ConfigMapPropertySource.java
+++ b/spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/ConfigMapPropertySource.java
@@ -16,7 +16,7 @@
 
 package org.springframework.cloud.kubernetes.config;
 
-import java.util.HashMap;
+import java.util.LinkedHashMap;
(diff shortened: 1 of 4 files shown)


moudatsos:
Created new PR on branch 1.1.x as requested in PR #554 
How do we move from that point on?

PR #561, opened 2020-06-07
upgrade to junit 5 

piomin:
In  the newest version some tests based on JUnit4 does not work. I fixed it by adding JUnit5 support for spring-cloud-kubernetes-leader module, and by adding junit-vintage-engine to spring-cloud-kubernetes-examples module since Arquillian Cube does not support JUnit5

### DIFF
--- a/spring-cloud-kubernetes-examples/kubernetes-leader-election-example/pom.xml
+++ b/spring-cloud-kubernetes-examples/kubernetes-leader-election-example/pom.xml
@@ -78,6 +78,14 @@
 					<skip>true</skip>
 				</configuration>
 			</plugin>
+			<plugin>
+				<artifactId>maven-surefire-plugin</artifactId>
+				<version>2.22.2</version>
+			</plugin>
+			<plugin>
+				<artifactId>maven-failsafe-plugin</artifactId>
+				<version>2.22.2</version>
+			</plugin>
 		</plugins>
 	</build>
 
--- a/spring-cloud-kubernetes-examples/kubernetes-leader-election-example/pom.xml
+++ b/spring-cloud-kubernetes-examples/kubernetes-leader-election-example/pom.xml
@@ -49,6 +49,11 @@
 			<artifactId>spring-boot-starter-test</artifactId>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>org.junit.vintage</groupId>
+			<artifactId>junit-vintage-engine</artifactId>
+			<scope>test</scope>
+		</dependency>
 	</dependencies>
 
 	<build>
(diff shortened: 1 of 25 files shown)


spring-cloud-kubernetes-examples/kubernetes-leader-election-example/pom.xml
```diff
@@ -78,6 +78,14 @@
 					<skip>true</skip>
 				</configuration>
 			</plugin>
+			<plugin>
+				<artifactId>maven-surefire-plugin</artifactId>
+				<version>2.22.2</version>
```
spencergibb, line 83: This should happen in spring-cloud-build

piomin: I see that currently these version is taken from parent pom.xml (branch master). I changed both versions for maven-failsafe-plugin and maven-surefire-plugin in the parent pom.xml.

		<plugins>
			<plugin>
				<groupId>org.apache.maven.plugins</groupId>
				<artifactId>maven-failsafe-plugin</artifactId>
				<version>${maven-failsafe-plugin.version}</version>
			</plugin>
		</plugins>

I don't know - maybe it should be taken from spring-cloud-build, but I'm guessing that it is the change for the whole Spring Cloud project. I had to set version of maven-surefire-plugin to 2.22.2 to make test works after changing enabling junit-jupiter for spring-cloud-kubernetes-leader module.

ryanjbaxter:
All tests should be passing and have been passing on CircleCI https://app.circleci.com/pipelines/github/spring-cloud/spring-cloud-kubernetes

piomin:
Well, I just replaced junit-vintage with junit-jupiter here. The tests were passing...

spencergibb:
yeah, this should be upgrade to junit 5 now instead of fixing errors.

spencergibb:
you'll need to rebase

piomin:
@spencergibb  Is it ok now?

spring-cloud-kubernetes-leader/src/test/java/org/springframework/cloud/kubernetes/leader/LeaderAutoConfigurationTests.java
```diff
@@ -16,21 +16,21 @@
...
 import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
 import org.springframework.boot.test.context.SpringBootTest;
 import org.springframework.http.MediaType;
-import org.springframework.test.context.junit4.SpringRunner;
 import org.springframework.test.web.reactive.server.WebTestClient;
 
 import static org.hamcrest.Matchers.containsString;
 
-@RunWith(SpringRunner.class)
+@ExtendWith(MockitoExtension.class)
```
Haybu, line 33: @piomin: won't this be a good case to use Spring TextContext leveraging @ExtendWith(SpringExtension)?

Haybu: could be changed later, not breaking the test for now

PR #564, opened 2020-06-15
Fix for KubernetesAutoConfiguration erroneously overwriting user password with proxy password

vijaykonnackal:
Issue: `KubernetesAutoConfiguration.kubernetesClientConfig` calls `withPassword` twice - the second time with proxy password resulting in user password replaced with proxy password.

Fix: the second `withPassword` method call is replaced with `withProxyPassword`.


ryanjbaxter:
@vijaykonnackal thanks for the PR, could you submit this to the 1.1.x branch and then we can merge it forward?

vijaykonnackal:
> 
> 
> @vijaykonnackal thanks for the PR, could you submit this to the 1.1.x branch and then we can merge it forward?

I seem to be github challenged here! You mean close this PR and open new one from forked 1.1.x ->spring-cloud 1.1.x?

PR #562, opened 2020-06-09
Enabling Spring Cloud Loadbalancer + integration tests

piomin:
This PR includes spring-cloud-starter-loadbalancer to spring-cloud-starter-kubernetes-all module. It also add some integration tests in spring-cloud-kubernetes-integration-tests.

PR #569, opened 2020-06-22
ConfigMap versioning

piomin:
It enables ConfigMap versioning similar to that in Spinnaker. More details here: https://piotrminkowski.com/2020/03/23/kubernetes-configmap-versioning-for-spring-boot-apps

### DIFF
--- a/spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/ConfigMapPropertySource.java
+++ b/spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/ConfigMapPropertySource.java
@@ -19,6 +19,7 @@ package org.springframework.cloud.kubernetes.config;
 import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 
@@ -56,13 +57,18 @@ public class ConfigMapPropertySource extends MapPropertySource {
 
 	private static final String PREFIX = "configmap";
 
+	private static final String LABEL_VERSION = "version";
+
+	private static final String LABEL_APP = "app";
+
 	public ConfigMapPropertySource(KubernetesClient client, String name) {
-		this(client, name, null, (Environment) null);
+		this(client, name, null, (Environment) null, false);
 	}
 
 	public ConfigMapPropertySource(KubernetesClient client, String name, String namespace,
-			String[] profiles) {
-		this(client, name, namespace, createEnvironmentWithActiveProfiles(profiles));
+			String[] profiles, boolean enableVersioning) {
+		this(client, name, namespace, createEnvironmentWithActiveProfiles(profiles),
+				enableVersioning);
 	}
 
 	private static Environment createEnvironmentWithActiveProfiles(
@@ -73,9 +79,9 @@ public class ConfigMapPropertySource extends MapPropertySource {
 	}
 
 	public ConfigMapPropertySource(KubernetesClient client, String name, String namespace,
-			Environment environment) {
-		super(getName(client, name, namespace),
-				asObjectMap(getData(client, name, namespace, environment)));
+			Environment environment, boolean enableVersioning) {
+		super(getName(client, name, namespace), asObjectMap(
+				getData(client, name, namespace, environment, enableVersioning)));
 	}
 
 	private static String getName(KubernetesClient client, String name,
@@ -89,12 +95,43 @@ public class ConfigMapPropertySource extends MapPropertySource {
 	}
 
 	private static Map<String, Object> getData(KubernetesClient client, String name,
-			String namespace, Environment environment) {
+			String namespace, Environment environment, boolean enableVersioning) {
 		try {
 			Map<String, Object> result = new LinkedHashMap<>();
-			ConfigMap map = StringUtils.isEmpty(namespace)
-					? client.configMaps().withName(name).get()
-					: client.configMaps().inNamespace(namespace).withName(name).get();
+			ConfigMap map = null;
+			if (enableVersioning) {
+				String version = environment.getProperty("info.app.version");
+				Optional<ConfigMap> optMap = StringUtils.isEmpty(namespace)
+						? client.configMaps().list().getItems().stream()
+								.filter(configMap -> configMap.getMetadata().getLabels()
+										.containsKey(LABEL_APP)
+										&& configMap.getMetadata().getLabels()
+												.get(LABEL_APP).equals(name)
+										&& configMap.getMetadata().getLabels()
+												.containsKey(LABEL_VERSION)
+										&& configMap.getMetadata().getLabels()
+												.get(LABEL_VERSION).equals(version))
+								.findFirst()
+						: client.configMaps().inNamespace(namespace).list().getItems()
+								.stream()
+								.filter(configMap -> configMap.getMetadata().getLabels()
+										.containsKey(LABEL_APP)
+										&& configMap.getMetadata().getLabels()
+												.get(LABEL_APP).equals(name)
+										&& configMap.getMetadata().getLabels()
+												.containsKey(LABEL_VERSION)
+										&& configMap.getMetadata().getLabels()
+												.get(LABEL_VERSION).equals(version))
+								.findFirst();
+				if (optMap.isPresent()) {
+					map = optMap.get();
+				}
+			}
+			else {
+				map = StringUtils.isEmpty(namespace)
+						? client.configMaps().withName(name).get()
+						: client.configMaps().inNamespace(namespace).withName(name).get();
+			}
 
 			if (map != null) {
 				result.putAll(processAllEntries(map.getData(), environment));
--- a/spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/ConfigMapConfigProperties.java
+++ b/spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/ConfigMapConfigProperties.java
@@ -36,6 +36,8 @@ public class ConfigMapConfigProperties extends AbstractConfigProperties {
 
 	private boolean enableApi = true;
 
+	private boolean enableVersioning = false;
+
 	private List<String> paths = new LinkedList<>();
 
 	private List<Source> sources = new LinkedList<>();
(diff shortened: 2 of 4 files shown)


spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/ConfigMapPropertySource.java
```diff
@@ -89,12 +95,43 @@ private static String getName(KubernetesClient client, String name,
...
-			String namespace, Environment environment) {
+			String namespace, Environment environment, boolean enableVersioning) {
 		try {
 			Map<String, Object> result = new LinkedHashMap<>();
-			ConfigMap map = StringUtils.isEmpty(namespace)
-					? client.configMaps().withName(name).get()
-					: client.configMaps().inNamespace(namespace).withName(name).get();
+			ConfigMap map = null;
+			if (enableVersioning) {
+				String version = environment.getProperty("info.app.version");
```
ryanjbaxter, line 103: Where is this coming from?

spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/ConfigMapPropertySource.java
```diff
@@ -89,12 +95,43 @@ private static String getName(KubernetesClient client, String name,
...
+								.filter(configMap -> configMap.getMetadata().getLabels()
+										.containsKey(LABEL_APP)
+										&& configMap.getMetadata().getLabels()
+												.get(LABEL_APP).equals(name)
+										&& configMap.getMetadata().getLabels()
+												.containsKey(LABEL_VERSION)
+										&& configMap.getMetadata().getLabels()
+												.get(LABEL_VERSION).equals(version))
+								.findFirst();
+				if (optMap.isPresent()) {
```
ryanjbaxter, line 126: what happens if it is not present?

piomin: it returns null, the same as for example here `client.configMaps().withName(name).get()` if it not found matching configmap

spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/ConfigMapPropertySource.java
piomin: From build-info.properties. So it would require to add goal build-info to spring-boot-maven-plugin:

```
<plugin>
	<artifactId>spring-boot-maven-plugin</artifactId>
	<executions>
		<execution>
			<goals>
				<goal>build-info</goal>
			</goals>
		</execution>
	</executions>
</plugin>
```

ryanjbaxter:
@piomin I talked about this today with the broader team and we wanted to know more about what about Spinnaker does that you think is valuable, can you elaborate?

Also our initial impression was that is might be better to take a more general approach of having the ability to filter ConfigMaps and Secrets based on labels, annotations, etc.  One filter may implement the versioning functionality you are proposing here

piomin:
@ryanjbaxter I think it could be also interesting https://spinnaker.io/reference/providers/kubernetes-v2/#traffic, or any more advanced form of traffic management.

Ok, I follow your idea. I'll try te prepare such change in this PR with more generic approach basing labels.

ryanjbaxter:
@piomin in what way do you think that the model from spinnakers traffic management could be applied here?

piomin:
@ryanjbaxter For example, we are talking about loadbalancer in SERVICE mode in another issue here: https://github.com/spring-cloud/spring-cloud-kubernetes/pull/562. Currently, it does not do any load balancing since it just call a single service always (or I'm missing something). We could give a possibility to use some additional configuration settings to provide not only a single service name, and load balancer would call addtional services like in this example from Spinnaker documentation.

And also one thing. I pushed some changes, that enable labellng mechanism for ConfigMap you have already proposed. But I have error during build on your CicrleCI: 
Could not resolve dependencies for project org.springframework.cloud:spring-cloud-kubernetes-core:jar:2.0.0-SNAPSHOT: The following artifacts could not be resolved: org.springframework.boot:spring-boot:jar:2.4.0-SNAPSHOT, org.springframework.boot:spring-boot-autoconfigure:jar:2.4.0-SNAPSHOT, org.springframework.boot:spring-boot-test-autoconfigure:jar:2.4.0-SNAPSHOT, org.springframework.boot:spring-boot-starter-web:jar:2.4.0-SNAPSHOT, org.springframework.boot:spring-boot-starter-json:jar:2.4.0-SNAPSHOT, org.springframework.boot:spring-boot-starter-webflux:jar:2.4.0-SNAPSHOT, org.springframework.boot:spring-boot-configuration-processor:jar:2.4.0-SNAPSHOT: Failure to find org.springframework.boot:spring-boot:jar:2.4.0-20200629.135853-137 in https://repo.spring.io/libs-snapshot-local was cached in the local repository, resolution will not be reattempted until the update interval of spring-snapshots has elapsed or updates are forced.

When I'm trying to builod it locally I have a following error:
Could not find artifact org.springframework.cloud:spring-cloud-dependencies-parent:pom:3.0.0-20200629.151922-184 in spring-snapshots (https://repo.spring.io/libs-snapshot-local)

Did you change something in dependencies? In build 108 everything worked properly.

spencergibb:
repo.spring.io is having issues today

spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/ConfigMapPropertySource.java
```diff
@@ -89,12 +97,34 @@ private static String getName(KubernetesClient client, String name,
 	}
 
 	private static Map<String, Object> getData(KubernetesClient client, String name,
-			String namespace, Environment environment) {
+			String namespace, Environment environment, boolean enableVersioning,
+			Map<String, String> labels) {
 		try {
 			Map<String, Object> result = new LinkedHashMap<>();
-			ConfigMap map = StringUtils.isEmpty(namespace)
```
ryanjbaxter, line 95: This is heading in the right direction but not exactly what I was envisioning.

We should allow the app to provide filters for both configmaps and secrets.  These filters can be just simple lambdas which take in a configmap or secret and return a boolean if it matches or doesn't.  This would eliminate all the versioning code here.  Maybe we could provide a versioning filter if we wanted to that apps could reuse, but it would not be part of the ConfigMapPropertySource code.  Does that make sense?

ryanjbaxter:
@piomin I dont really see what the ability to call other services via a load balancer has to do with versioning configmaps.

piomin:
@ryanjbaxter ok, I thought you were asking about this traffic management and generally about other interesting features of Spinnaker that my be applied here. Let's move back to the config versioning only - sorry for missunderstanding. Typically you are able to rollback deployment on Kubernetes, but Kubernetes does not allow you to rollback ConfigMap or Secret. So if your configuration is changing together the application, and you need to perform rollback, you have to change configuration in Secrets or ConfigMaps manually (for exaple you may have there feature toggle or smth like that). With this approach you may have multiple versions of your ConfigMaps and in case of deployment rollback everything works out-of-the-box.

spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/ConfigMapPropertySource.java
piomin: Yes, definitely it has sense. I'll think about it.

ryanjbaxter:
@piomin but what happens if the configmap from the previous version of your app gets deleted?  You have nothing to roll back to.  When we did versioning in Spring Cloud Config we relied on the versioning provided by the backing implementation, if there was one.  If you needed to restore some configuration you went to your git repo (or svn) and went to the version of the file you wanted and restored it.

The solution you are proposing forces people to keep their configmaps around infinitely if they want to ave a chance or rolling back their config.

That said I think if this is the solution you want to implement for versioning configmaps that is fine with us.  However we don't want it baked into Spring Cloud Kubernetes.  This is why we are proposing a more abstract way of providing a way of filtering configmaps.

piomin:
@ryanjbaxter well, I could also ask what happens if there is no configmap with metadata.name corresponding to the application name :) We are forcing users to have such configmap if they want that the current mechanism works for them, right? :)  
About that versioning concept - maybe we are not forcing them to keep their configmaps infinetely, but only such many versions back as the numbers of replication controllers history that allows to rollback deployment. 
But of course it shouldn't be a part of ConfigMapPropertySource code.

spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/ConfigMapPropertySource.java
piomin: I removed the whole code for versioning, and just leave that for labelling support. If we want provide a filter we should probably use AOP (?), which is currently not used in that project.

piomin:
@ryanjbaxter I removed code responsible for versioning support, and just left the code for generic labelling support. I see that such support has been currently implemented for Secret, so I just do it in similar way  for ConfigMap sources.

ryanjbaxter:
I am fine with taking the changes as is, but I was actually envisioning something different.  Let me describe

You could have and interface called `PropertySourceFilter` with a method called `filter` that would take in either a `ConfigMap` or `Secret` and return a boolean value if it was something the app was interested in using as a property source.

Apps using Spring Cloud Kubernetes could provide a bean like the following

```
@Bean
public List<PropertySourceFilter> myFilters() {

  List<PropertySourceFilter> filters = ....
  filters.add((configMap) -> {
    //implement filter logic such as versioning
  });
  return filters;
```

Then we could get this list filter out the ConfigMaps and Secrets the app wants as a property source.

This type of functionality would eliminate the need for name matching as well and allow the app to use whatever name it wants and pick ConfigMaps and Secretes based on other parameters, like labels, and annotations.

Maybe the default implementation just matches name so we retain the existing functionality.

piomin:
@ryanjbaxter Maybe let's merge these changes and I create another PR with the proposal of changes you described ?

piomin:
@ryanjbaxter any feedback on it?

ryanjbaxter:
I think I would rather implement the proposed solution I had above, rather than have some in between solution in the code for a while.

wind57:
imo, if we ever implement this, we should drop explicit support for labels in secrets implementation and provide this filtering idea instead. It's just a bit un-uniform now, config maps don't have that, while secrets do. 

I was thinking that we could have a few beans conditionally enabled like : `VersioningSupport` or `LabelsSupport`, etc. At some point in time, if there is no PR for this, I might pick it up.

wind57:
@ryanjbaxter the original issue has been fixed - adding labels support for config maps; but obviously not the suggestion you had with `PropertySourceFilter`, which is a big change...

what do you want to do with this defect? thank you

Issue #558, opened 2020-05-21
Reload mode stops working after a while


Snazzythat:
Hi,

I followed the tutorials to setup spring-cloud-starter-kubernetes-config with my Spring application. I got it to read the properties from the ConfigMap and update on event when I edit my ConfigMap:

`2020-05-21 05:04:19.606 +0000 [OkHttp https://10.96.0.1/...]  INFO  o.s.b.SpringApplication - Started application in 1.397 seconds (JVM running for 1967.199)
2020-05-21 05:08:59.215 +0000 [OkHttp https://10.96.0.1/...]  INFO  o.s.c.k.c.r.EventBasedConfigurationChangeDetector - Detected change in config maps
2020-05-21 05:08:59.215 +0000 [OkHttp https://10.96.0.1/...]  INFO  o.s.c.k.c.r.EventBasedConfigurationChangeDetector - Reloading using strategy: REFRESH
`

However, if I leave the application for around 45-60 minutes and try to issue a ConfigMap edit again, the event is not fired, no logs can be seen. I have to restart the pod in order to be able to issue ConfigMap changes again. I tired to specify newer version of kubernetes-client (as per compatibility matrix: https://github.com/fabric8io/kubernetes-client#compatibility-matrix) but it did not help. Any ideas, could it be websocket issue with Kubernetes API? Thanks!

ENV:
kubernetes version: 1.15.3
org.springframework.boot: 2.2.4.RELEASE
spring-cloud-starter-kubernetes-config: 1.1.2.RELEASE
kubernetes-client: 4.10.1

Snazzythat:
Fromt he logs, looks like web socket connection is forcefully closed after some time:

`2020-05-22 14:38:43.279 +0000 [OkHttp WebSocket https://10.96.0.1/...]  DEBUG i.f.k.c.d.i.WatchConnectionManager - WebSocket close received. code: 1000, reason: 

2020-05-22 14:38:43.279 +0000 [OkHttp WebSocket https://10.96.0.1/...]  DEBUG i.f.k.c.d.i.WatchConnectionManager - Submitting reconnect task to the executor

2020-05-22 14:38:43.279 +0000 [scheduleReconnect|Executor for Watch 2065320366]  DEBUG i.f.k.c.d.i.WatchConnectionManager - Scheduling reconnect task

[...]

Issue #557, opened 2020-05-19
Reload mode - event not working for k8 config - 1.1.2.RELEASE


dintualex:
I'm able to fetch the K8 config maps on startup but on subsequent update to the config map, its not resulting in consuming the event and updating the values in my spring boot application.

Here are the versions:
spring boot -     2.2.7.RELEASE    
spring-cloud-gcp - 1.2.2.RELEASE
spring-cloud - Hoxton.SR4

Its working  fine in 
spring boot -     2.1.11.RELEASE    
spring-cloud-gcp - 1.1.3.RELEASE
spring-cloud - Greenwich.SR4
Able to fetch the config map during startup as well as consuming the event when there is a change in the config map

Snazzythat:
I have an issue with events when updating the configmap as well. At first, the application picks up the configmap values and changes to configmap gets do get triggered:

2020-05-20 17:41:28.493 +0000 [OkHttp https://10.96.0.1/...]  INFO  o.s.c.k.c.r.EventBasedConfigurationChangeDetector - Detected change in config maps
2020-05-20 17:41:28.493 +0000 [OkHttp https://10.96.0.1/...]  INFO  o.s.c.k.c.r.EventBasedConfigurationChangeDetector - Reloading using strategy: REFRESH

However, if I leave the app running in my k8s cluster for about 15-20 minutes, the events stop firing when I submit changes to the config map.

spring boot - 2.2.4.RELEASE
spring-cloud-starter-kubernetes-config -1.1.2.RELEASE

ChrisIgel:
Same problem. Here are the relevant logs when the problem occurs:
```
{"time":"2020-07-16T14:49:39.695Z","msg":"WebSocket close received. code: 1000, reason: ","logger":"io.fabric8.kubernetes.client.dsl.internal.WatchConnectionManager","thread":"OkHttp WebSocket https://10.253.0.1/...","level":"DEBUG"}
{"time":"2020-07-16T14:49:39.695Z","msg":"Submitting reconnect task to the executor","logger":"io.fabric8.kubernetes.client.dsl.internal.WatchConnectionManager","thread":"OkHttp WebSocket https://10.253.0.1/...","level":"DEBUG"}
[...]

PR #570, opened 2020-06-23
Re-watch events when disconnected from k8s API

yuusaku-t:
If any errors occur and kubernetes client is disconnected from the API, `EventBasedConfigurationChangeDetector` should re-watch the events for external resources like ConfigMap. This patch implements the `onClose` methods of Watcher for the error handling:

```java
/**
 * Run when the watcher finally closes.
 *
 * @param cause What caused the watcher to be closed. Null means normal close.
 */
void onClose(KubernetesClientException cause);
```
ref: https://github.com/fabric8io/kubernetes-client/blob/v4.4.0/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/Watcher.java#L27

### An example of the errors

In my case, the following exception is passed to `onClose` method under the same situation as reported in https://github.com/spring-cloud/spring-cloud-kubernetes/issues/558:

```
io.fabric8.kubernetes.client.KubernetesClientException: too old resource version: 35168187 (41169815)
	at io.fabric8.kubernetes.client.dsl.internal.WatchConnectionManager$1.onMessage(WatchConnectionManager.java:259) ~[kubernetes-client-4.9.0.jar!/:na]
	at okhttp3.internal.ws.RealWebSocket.onReadMessage(RealWebSocket.java:323) ~[okhttp-3.12.0.jar!/:na]
	at okhttp3.internal.ws.WebSocketReader.readMessageFrame(WebSocketReader.java:219) ~[okhttp-3.12.0.jar!/:na]
	at okhttp3.internal.ws.WebSocketReader.processNextFrame(WebSocketReader.java:105) ~[okhttp-3.12.0.jar!/:na]
	at okhttp3.internal.ws.RealWebSocket.loopReader(RealWebSocket.java:274) ~[okhttp-3.12.0.jar!/:na]
	at okhttp3.internal.ws.RealWebSocket$2.onResponse(RealWebSocket.java:214) ~[okhttp-3.12.0.jar!/:na]
	at okhttp3.RealCall$AsyncCall.execute(RealCall.java:206) ~[okhttp-3.12.0.jar!/:na]
	at okhttp3.internal.NamedRunnable.run(NamedRunnable.java:32) ~[okhttp-3.12.0.jar!/:na]
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128) ~[na:na]
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628) ~[na:na]
	at java.base/java.lang.Thread.run(Thread.java:834) ~[na:n
```

The kubernetes client will not automatically reconnect when such an error occurs. The following code is causing the reconnection to be interrupted:

https://github.com/fabric8io/kubernetes-client/blob/v4.4.0/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/WatchConnectionManager.java#L285-L292

### Releated Issues

* https://github.com/spring-cloud/spring-cloud-kubernetes/issues/558
* https://github.com/spring-cloud/spring-cloud-kubernetes/issues/557

### DIFF
--- a/spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/reload/EventBasedConfigurationChangeDetector.java
+++ b/spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/reload/EventBasedConfigurationChangeDetector.java
@@ -16,8 +16,8 @@
 
 package org.springframework.cloud.kubernetes.config.reload;
 
-import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
 import javax.annotation.PostConstruct;
 import javax.annotation.PreDestroy;
@@ -59,7 +59,7 @@ public class EventBasedConfigurationChangeDetector extends ConfigurationChangeDe
 
 		this.configMapPropertySourceLocator = configMapPropertySourceLocator;
 		this.secretsPropertySourceLocator = secretsPropertySourceLocator;
-		this.watches = new HashMap<>();
+		this.watches = new ConcurrentHashMap<>();
 	}
 
 	@PostConstruct
@@ -70,18 +70,7 @@ public class EventBasedConfigurationChangeDetector extends ConfigurationChangeDe
 				&& this.configMapPropertySourceLocator != null) {
 			try {
 				String name = "config-maps-watch";
-				this.watches.put(name, this.kubernetesClient.configMaps()
-						.watch(new Watcher<ConfigMap>() {
-							@Override
-							public void eventReceived(Action action,
-									ConfigMap configMap) {
-								onEvent(configMap);
-							}
-
-							@Override
-							public void onClose(KubernetesClientException e) {
-							}
-						}));
+				watchConfigMap(name);
 				activated = true;
 				this.log.info("Added new Kubernetes watch: " + name);
 			}
@@ -97,17 +86,7 @@ public class EventBasedConfigurationChangeDetector extends ConfigurationChangeDe
 			try {
 				activated = false;
 				String name = "secrets-watch";
-				this.watches.put(name,
-						this.kubernetesClient.secrets().watch(new Watcher<Secret>() {
-							@Override
-							public void eventReceived(Action action, Secret secret) {
-								onEvent(secret);
-							}
-
-							@Override
-							public void onClose(KubernetesClientException e) {
-							}
-						}));
+				watchSecret(name);
 				activated = true;
 				this.log.info("Added new Kubernetes watch: " + name);
 			}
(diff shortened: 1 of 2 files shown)


spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/reload/EventBasedConfigurationChangeDetector.java
```diff
@@ -138,6 +117,46 @@ public void unwatch() {
...
+				onEvent(configMap);
+			}
+
+			@Override
+			public void onClose(KubernetesClientException error) {
+				if (error != null) {
+					if (log.isDebugEnabled()) {
+						log.warn("Try to re-watch: " + name, error);
+					}
+					watchConfigMap(name);
```
ryanjbaxter, line 133: Could we not get into some kind of infinite loop where we indefinitely retry here?  It would be nice if we could add some backoff functionality or even just try a set number of times.

yuusaku-t: Ok. I'll try to implement backoff because the errors reported in #558 occur periodically.

yuusaku-t: I added backoff functionality by ba47a20. Please check it.

yuusaku-t: @ryanjbaxter Should I make any additional changes?

brachipa:
@ryanjbaxter @yuusaku-t seems like I have same issue, watcher failed and doesn't reconnect again, what is the status of that PR? will it be merged soon?

yuusaku-t:
I'm waiting for a re-review.

ryanjbaxter (changes requested):
Sorry for the slow response!  Could you add some tests?

yuusaku-t:
@ryanjbaxter I added test cases by 613eb3d. Could you please check?

ryanjbaxter:
@yuusaku-t I was reviewing this with other team members and they were wondering if we could use Spring Retry?  This could simplify the implementation a bit.  Did you look into that at all?

yuusaku-t:
@ryanjbaxter I hadn't thought about using Spring Retry at all, because it's not included in the spring-cloud-kubernetes-config project's dependency libraries. Let me see if I can simplify the implementation using Spring Retry.

ryanjbaxter:
@yuusaku-t thanks, one thought might to make it optional and have retry implementation of spring retry is on the class path

ryanjbaxter:
Spring Retry is also being used in this PR https://github.com/spring-cloud/spring-cloud-kubernetes/pull/648

whgibbo:
Will this also address https://github.com/spring-cloud/spring-cloud-kubernetes/issues/730

yuusaku-t:
@ryanjbaxter Sorry for the delay in response. Could you please check 57bf977 ?
I didn't use Spring Retry, but I think it's a little bit cleaner. I've tried various approaches to see if I can implement this using Spring Retry, but it's difficult and more time consuming. I would like to remove it from the scope of this pull request.

PetricHwang:
I have also encountered this problem.

ryanjbaxter:
I actually think this was addressed in this PR https://github.com/spring-cloud/spring-cloud-kubernetes/pull/842

It should be in our 2020.0.4 release

PR #583, opened 2020-07-19
Allow activeProfile not to be appended

### LINKED ISSUE #508: Allow activeProfile not to be appended
When configuring the configMap, the activate Profile is appended to the name

```
spring:
  cloud:
    kubernetes:
      enabled: true
      client:
        namespace: default
      reload:
        enabled: true
        monitoring-config-maps: true
        monitoring-secrets: false
        mode: event
        strategy: refresh
      config:
        enable-api: true
        sources:
          - namespace: default
            name: feature-toggles
```

results in `https://172.20.0.1/api/v1/namespaces/default/configmaps/feature-toggles-dev`, which active profile `dev`.

Cause is the line 
`String mapNameWithProfile = name + "-" + activeProfile;` 

from
`m2/repository/org/springframework/cloud/spring-cloud-kubernetes-config/1.1.1.RELEASE/spring-cloud-kubernetes-config-1.1.1.RELEASE-sources.jar!/org/springframework/cloud/kubernetes/config/ConfigMapPropertySource.java:102`

Please make the addition of the activate profile to the configMap an configuration property ("append-profile: true/false,  default: true).

anshlykov:
Resolves #508

### DIFF
--- a/spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/ConfigMapConfigProperties.java
+++ b/spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/ConfigMapConfigProperties.java
@@ -28,6 +28,7 @@ import org.springframework.util.StringUtils;
  * Config map configuration properties.
  *
  * @author Ioannis Canellos
+ * @author Andrey Shlykov
  */
 @ConfigurationProperties("spring.cloud.kubernetes.config")
 public class ConfigMapConfigProperties extends AbstractConfigProperties {
@@ -106,6 +107,11 @@ public class ConfigMapConfigProperties extends AbstractConfigProperties {
 		 */
 		private String namespace;
 
+		/**
+		 * Enable or disable appending a profile name to the ConfigMap name.
+		 */
+		private boolean appendProfile = true;
+
 		public Source() {
 		}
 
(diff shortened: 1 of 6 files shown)


ryanjbaxter (changes requested):
Rather than disable this functionality I think it would be better to retrieve the config map without the profile appended if there isn't one with the profile appended

wind57:
I plan to take this one next, and I actually agree with the approach here. If we read the description of the initial issue, OP has complained that:

> When the implementation now tries to access `app-config-dev` a security violation is logged.

So we should not even try to read such a config map to begin with. Your proposal is (as far as I understand it), _try_ to read `app-config-dev` first, and if it's not present, read `app-config`, which I think is against what users would want. 

Please correct me if I am wrong @ryanjbaxter. If I am not, I will take this defect and fix it.

ryanjbaxter:
@wind57 if thats the case, if its the security violation warning that is the issue, then I agree.

wind57:
thank you for the feedback, I started work on this one, should have the PR ready by early next week.

ryanjbaxter:
Issue being fixed by #856

marcguilera:
Hello, I noticed this change after updating Spring Cloud. For my use-case it is not desirable to read every possible combination of active profile so I tried to disable it without success. My `bootstrap.yaml` looks like this:

```
---
spring:
  application:
    name: service
  cloud:
    kubernetes:
      config:
        enabled: false

kubernetes:
  enabled: true
  namespace: platform
  configmap: ${spring.application.name}-config

---
spring:
  config:
    activate:
      on-profile: kubernetes
  cloud:
    kubernetes:
      reload:
        enabled: ${kubernetes.enabled}
      config:
        enabled: ${kubernetes.enabled}
        namespace: ${kubernetes.namespace}
        useNameAsPrefix: false
        includeProfileSpecificSources: false
        name: ${kubernetes.configmap}
        # also tried with sources array
```

According to the documentation the `includeProfileSpecificSources` should disable the reading but it didn't as I see in the logs. Am I missing something or is this a problem? Moreover what was the rationale to update this behaviour?

cc: @wind57

wind57:
Hey, I can't really tell what is going on without a minimal example. Can you please share one? Like a github repo? When you provide such, I will for surr take a look. Thx for looking into this feature

Issue #585, opened 2020-07-22
Not using discovery dependency, disabled config, but still see attempted call to K8s API


stevengibbszions:
**Describe the bug**
We are using spring-cloud-kubernetes only for the purposes of reading secrets into the spring environment (and that is working well!).  However, the library is still making a call to the Kubernetes API, even after disabling the config option via bootstrap.yaml.

Dependency in `pom.xml`:

```xml
		<dependency>
			<groupId>org.springframework.cloud</groupId>
			<artifactId>spring-cloud-starter-kubernetes-config</artifactId>
			<version>1.1.3.RELEASE</version>
		</dependency>
```

`bootstrap.yaml`

```yaml
spring:
  cloud:
    kubernetes:
      config:
        enabled: false
        enable-api: false
      secrets:
        enable-api: false
        paths:
        - /a/secrets
        - /b/secrets
```
**PROBLEM**
A call is still being made to the k8s API.  We see this error on startup.

```log
Failed to get pod with name:[my-pod]. You should look into this if things aren't working as you expect. Are you missing serviceaccount permissions?
io.fabric8.kubernetes.client.KubernetesClientException: Failure executing: GET at: https://10.43.0.1/api/v1/namespaces/default/pods/my-pod. Message: Forbidden!Configured service account doesn't have access. Service account may have been revoked. pods "my-pod" is forbidden: User "system:serviceaccount:default:default" cannot get resource "pods" in API group "" in the namespace "default".
at io.fabric8.kubernetes.client.dsl.base.OperationSupport.requestFailure(OperationSupport.java:503) ~[kubernetes-client-4.4.1.jar:na]
at io.fabric8.kubernetes.client.dsl.base.OperationSupport.assertResponseCode(OperationSupport.java:440) ~[kubernetes-client-4.4.1.jar:na]
at io.fabric8.kubernetes.client.dsl.base.OperationSupport.handleResponse(OperationSupport.java:406) ~[kubernetes-client-4.4.1.jar:na]
at io.fabric8.kubernetes.client.dsl.base.OperationSupport.handleResponse(OperationSupport.java:365) ~[kubernetes-client-4.4.1.jar:na]
at io.fabric8.kubernetes.client.dsl.base.OperationSupport.handleGet(OperationSupport.java:330) ~[kubernetes-client-4.4.1.jar:na]
[...]

PR #599, opened 2020-07-28
toggling k8s configuration with a relevant property fixes gh-585

Haybu:
To enable or disable only the k8s configuration feature it's expected to use the relevant property `spring.cloud.kubernetes.config.enabled=true/false`, without a need to disable Kubernetes itself using property 'spring.cloud.kubernetes.enabled`. With this modification, it's possible to disable only the configuration feature using property `spring.cloud.kubernetes.config.enabled` or when choosing to completely turn off Kubernetes using `spring.cloud.kubernetes.enabled`.


spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/reload/ConfigReloadAutoConfiguration.java
```diff
@@ -48,7 +48,8 @@
  * @author Nicolla Ferraro
  */
 @Configuration(proxyBeanMethods = false)
-@ConditionalOnProperty(value = "spring.cloud.kubernetes.enabled", matchIfMissing = true)
+@ConditionalOnProperty(name = "spring.cloud.kubernetes.config.enabled",
```
spencergibb, line 51: This probably needs both right? If the whole thing is disabled or config is disabled.

spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/reload/ConfigReloadAutoConfiguration.java
```diff
@@ -48,7 +48,8 @@
...
+		havingValue = "true")
```
spencergibb, line 52: Why this change?

Haybu: change the property itself and explicitly specify it should be true in value for clarity. don't mind using the old pattern.

spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/reload/ConfigReloadAutoConfiguration.java
Haybu: Although a test triage didn't active config, but agree, it would be clear to have both.

ryanjbaxter (approved):
@Haybu could you resolve the merge conflicts?

ryanjbaxter:
@Haybu can you look at the failures?

ryanjbaxter (commented):
@Haybu looks good go ahead and merge it to the 1.1.x branch and master!

ryanjbaxter:
@Haybu merge it into master as well

Issue #513, opened 2020-01-09
start fail in spring cloud gateway with elasticsearch


hanlubiao:
**Describe the bug**
start fail in spring cloud gateway with elasticsearch

ExceptionStackTrace:
Caused by: java.lang.IllegalStateException: null
        at java.util.Optional.orElseThrow(Optional.java:290)
        at org.springframework.cloud.kubernetes.discovery.KubernetesDiscoveryClient.findEndpointPort(KubernetesDiscoveryClient.java:189)
        at org.springframework.cloud.kubernetes.discovery.KubernetesDiscoveryClient.getInstances(KubernetesDiscoveryClient.java:157)
        at org.springframework.cloud.client.discovery.composite.CompositeDiscoveryClient.getInstances(CompositeDiscoveryClient.java:53)
        at reactor.core.publisher.FluxMapFuseable$MapFuseableConditionalSubscriber.tryOnNext(FluxMapFuseable.java:301)
        at reactor.core.publisher.FluxIterable$IterableSubscriptionConditional.slowPath(FluxIterable.java:515)

Reason:
When I install elasticsearch by cloud-on-k8s(https://github.com/elastic/cloud-on-k8s), it auto create a service myes-es-default without Port, but,  must have Port in spring-cloud-kubernetes

spencergibb:
Things have changed since this was reported, can you try again with Hoxton.SR6?

joshrosso:
Hello! I ran into this as well and wanted to talk about my findings in case root cause has not been assessed!

Elasticsearch endpoint does *not* specify a port, which as far as k8s is concerned is OK.

```
logging                  elastic-webhook-server               100.99.232.133:9443            
logging                  elasticsearch-es-default              100.106.192.199              
logging                  elasticsearch-es-http                   100.106.192.199:9200                          
logging                  elasticsearch-es-transport           100.106.192.199:9300    
```

[...]

PR #607, opened 2020-08-24
Support endpoints without ports specified in discovery client.

joshrosso:
Support endpoints without ports specified in discovery client.

Kubernetes considers endpoints without ports to be valid.

resolves https://github.com/spring-cloud/spring-cloud-kubernetes/issues/513


ryanjbaxter:
@joshrosso can you pull in the latest changes from master, it should fix the failing test.  Also, I think this would be fine to put in the 1.1.x branch as well, if you agree can you please change the PR to target that branch?  We will merge it forward?

Issue #516, opened 2020-01-17
Add support for spring cloud loadbalancer


spencergibb:

spencergibb:
Needs to include docs referencing ribbon removed in #530

anjanitsip:
-Any time line for this issue Add support for spring cloud loadbalancer #516

OlgaMaciaszek:
Done for 2020.0.0.M4.

PR #610, opened 2020-08-27
LB integration hoxton

OlgaMaciaszek:


OlgaMaciaszek:
Fixes gh-516 in Hoxton.

OlgaMaciaszek:
Have verified that the SC LoadBalancer opt-in mechanism works OOTB here.

PR #617, opened 2020-09-06
refactored + tests : changed method that deals with MapPropertySource

wind57:


wind57:
I've separated this into 3 commits, to easier review; though the code is fairly trivial to understand.

PR #621, opened 2020-09-07
some improvements

wind57:


spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/ConfigUtils.java
```diff
@@ -39,38 +39,33 @@ private ConfigUtils() {
 		throw new IllegalStateException("Can't instantiate a utility class");
 	}
 
-	public static <C extends AbstractConfigProperties> String getApplicationName(
-			Environment env, String configName, String configurationTarget) {
-		String name = configName;
-		if (StringUtils.isEmpty(name)) {
+	public static String getApplicationName(Environment env, String configName,
```
wind57, line 42: - un-used type parameter
 - Strings are immutable, no need for the local copy
 - the check if `Log::debug` is enabled is already done _inside_ `Log::debug`; no need to do that yourself. The only time this does make sense, is when the message itself is expensive to compute, but that is by far not the case here

spring-cloud-kubernetes-config/src/test/java/org/springframework/cloud/kubernetes/config/MultipleConfigMapsTests.java
```diff
@@ -67,27 +67,23 @@ public static void setUpBeforeClass() {
...
-
-		createConfigmap(server, "defname", "s2", new HashMap<String, String>() {
-			{
-				put("bean.common-message", "c2");
-				put("bean.message2", "m2");
-			}
-		});
-
-		createConfigmap(server, "othername", "othernamespace",
-				new HashMap<String, String>() {
```
wind57, line 85: there is nothing wrong with this anti-pattern, especially in tests. The biggest problem is the number of users that look at Spring code and might think - "hey, Spring does it!". As such, let's get rid of this common anti-pattern.

wind57:
@ryanjbaxter imho: this should be a no-brainer too, as it just clears things a bit

PR #622, opened 2020-09-07
LockSupport instead of sleep

wind57:


spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/reload/ConfigReloadAutoConfiguration.java
```diff
@@ -140,11 +142,8 @@ public ConfigurationUpdateStrategy configurationUpdateStrategy(
 		private static void wait(ConfigReloadProperties properties) {
 			final long waitMillis = ThreadLocalRandom.current()
 					.nextLong(properties.getMaxWaitForRestart().toMillis());
-			try {
-				Thread.sleep(waitMillis);
-			}
-			catch (InterruptedException ignored) {
```
wind57, line 146: if you want to ignore `InterruptedException` (which is a terrible decision IMHO), then do not use `Thread:sleep`, at all; instead use a construct that does not throw a checked exception and is silent on problems, too.

ryanjbaxter:
Based on the thread on SO I think it might be better to just leave it as is
https://stackoverflow.com/questions/10397881/java-locksupport-parknanos-vs-thread-sleep

PR #623, opened 2020-09-07
minor clean-up

wind57:


spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/reload/ConfigurationUpdateStrategy.java
```diff
@@ -26,15 +26,13 @@
  */
 public class ConfigurationUpdateStrategy {
 
-	private String name;
+	private final String name;
```
wind57, line 29: `final` is not only a good practice, but also a safely published Object.

spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/reload/ConfigurationUpdateStrategy.java
```diff
@@ -26,15 +26,13 @@
...
 
-	private Runnable reloadProcedure;
+	private final Runnable reloadProcedure;
 
 	public ConfigurationUpdateStrategy(String name, Runnable reloadProcedure) {
-		Objects.requireNonNull(name, "name cannot be null");
-		Objects.requireNonNull(reloadProcedure, "reloadProcedure cannot be null");
-		this.name = name;
-		this.reloadProcedure = reloadProcedure;
+		this.name = Objects.requireNonNull(name, "name cannot be null");
```
wind57, line 34: inline

spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/reload/ConfigurationUpdateStrategy.java
```diff
@@ -47,10 +45,7 @@ public void reload() {
 
 	@Override
 	public String toString() {
-		final StringBuilder sb = new StringBuilder("ConfigurationUpdateStrategy{");
-		sb.append("name='").append(this.name).append('\'');
-		sb.append('}');
-		return sb.toString();
+		return "ConfigurationUpdateStrategy{name='" + this.name + "'}";
```
wind57, line 48: far simpler `toString` and `StringBuilder` is going to be generated by the compiler _anyway_; even in java-9 where `String::concat` compiles as `invokedynamic`, but will still _use_ `StringBuilder::append`, by default.

PR #624, opened 2020-09-07
changes

wind57:


spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/ConfigMapConfigProperties.java
```diff
@@ -36,9 +35,9 @@
 
 	private boolean enableApi = true;
 
-	private List<String> paths = new LinkedList<>();
+	private List<String> paths = Collections.emptyList();
```
wind57, line 38: it does not look like the code cares what type of `List` this holds, so why have it a `LinkedList` when it could be simply an empty one? I guess the point of initializing this with _something_ is so that the callers will never get a `null`, but an _empty_ List instead; in such a case let's make this explicit.

spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/ConfigMapConfigProperties.java
```diff
@@ -74,12 +73,9 @@ public void setSources(List<Source> sources) {
 	 */
 	public List<NormalizedSource> determineSources() {
 		if (this.sources.isEmpty()) {
-			return new ArrayList<NormalizedSource>() {
-				{
-					add(new NormalizedSource(ConfigMapConfigProperties.this.name,
```
wind57, line 79: this is a _very_ common anti-pattern.

PR #625, opened 2020-09-08
improvements

wind57:


wind57:
There are many things going on this PR. This method:

```
	private void readFile(Path path, CompositePropertySource composite) {
		try {
			result.put(path.getFileName().toString(),
					new String(Files.readAllBytes(path)).trim());
			if (!result.isEmpty()) {
				composite.addFirstPropertySource(new MapPropertySource(
						path.getFileName().toString().toLowerCase(), result));
			}
		}
		catch (IOException e) {
			LOG.warn("Error reading properties file", e);
		}
	}
```

because it wraps too much code inside a `catch (IOException e) {` block, does a weird check: `if (!result.isEmpty()) {` - which can only happen when there is an Exception thrown. Then, because it only creates a Map from a single entry, can leverage `Collections::singletonMap`. This entire method was moved to the custom `Collector` with this refactoring:

```
private MapPropertySource property(Path filePath) {

			String fileName = filePath.getFileName().toString();

			try {
				String content = new String(Files.readAllBytes(filePath)).trim();
				return new MapPropertySource(fileName.toLowerCase(),
						Collections.singletonMap(fileName, content));
			}
			catch (IOException e) {
				LOG.warn("Error reading properties file", e);
				return null;
			}
		}
```

The same method `readFile` does also call internally:  `composite.addFirstPropertySource(new MapPropertySource(...`. When you look at `addFirstPropertySource` - it surely looks like you care about order, but at the same time this comes from a `path` that we get from `Files::walk` - which _does not_ guarantee any order at all. So why invoke this method to begin with? Not only internally this method is not very cheap (`addFirstPropertySource`), but implying that the code cares about order when the underlying source is not ordered (`Files::walk` and `this.properties.getPaths().stream()`) makes very little sense.

There is a also a problem here:

```
	private void putPathConfig(CompositePropertySource composite) {
		this.properties.getPaths().stream().map(Paths::get).filter(Files::exists)
				.forEach(p -> putAll(p, composite));
	}
```

which relies on side-effects via `forEach` (stream documentation strongly discourages this) and in general is seen as incorrect (if you need an unexpected failure - I can provide one). The same problem exists in `putAll`, via the same `forEach`.

---

But if you look in perspective, this class tries to achieve a rather simple thing. Having a `SecretsConfigProperties` - we want to get all of its File contents from all the directories, from all the paths that it has; and store that in a `CompositePropertySource`.  This is why a custom Collector was added. It takes in as "input" a File Path and returns `List<MapPropertySource>`. As such, we can chain and create a single stream pipeline, via :

```
		private void putPathConfig(CompositePropertySource composite) {

		this.properties
			.getPaths()
			.stream()
			.map(Paths::get)
			.filter(Files::exists)
			.flatMap(x -> {
				try {
					return Files.walk(x);
				}
				catch (IOException e) {
					LOG.warn("Error walking properties files", e);
					return null;
				}
			})
			.filter(Objects::nonNull)
			.filter(Files::isRegularFile)
			.collect(new MapPropertySourceCollector())
			.forEach(composite::addPropertySource);
	}
```

spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/SecretsPropertySourceLocator.java
```diff
@@ -61,22 +70,23 @@ public SecretsPropertySourceLocator(KubernetesClient client,
...
-	public PropertySource locate(Environment environment) {
+	public PropertySource<?> locate(Environment environment) {
 		if (environment instanceof ConfigurableEnvironment) {
 			ConfigurableEnvironment env = (ConfigurableEnvironment) environment;
 
 			List<SecretsConfigProperties.NormalizedSource> sources = this.properties
 					.determineSources();
 			CompositePropertySource composite = new CompositePropertySource(
 					"composite-secrets");
-			if (this.properties.isEnableApi()) {
```
wind57, line 72: the previous code would make the return from `getKubernetesPropertySourceForSingleSecret` to be the _last_ one in `composite`; and that is the only guarantee that existed here. 

To maintain it, I `putPathConfig(composite);` before it; and also changed to `composite.addPropertySource` instead of `composite.addFirstPropertySource`

ryanjbaxter:
Please fix the merge conflicts

wind57:
@ryanjbaxter done.

PR #627, opened 2020-09-11
Add loadbalancer example

OlgaMaciaszek:


spring-cloud-kubernetes-examples/kubernetes-loadbalancer-example/greeting-service/.mvn/wrapper/maven-wrapper.properties
```diff
@@ -0,0 +1,2 @@
+distributionUrl=https://repo.maven.apache.org/maven2/org/apache/maven/apache-maven/3.6.1/apache-maven-3.6.1-bin.zip
```
spencergibb, line 1: should probably not include mvnw

PR #634, opened 2020-09-17
Fix port filtering logic

gmaiztegi:
The sensible logic would be "use all ports if there isn't a port name set in the configuration, and use only those that match if the por is set in the configuration". This fixes the logic.


ryanjbaxter (commented):
I am not sure I understand what the problem is that change is fixing, maybe add a unit test that demonstrates the issue?

ryanjbaxter:
Closing in favor of #663

PR #639, opened 2020-09-21
Issue #635 - Decouples configMap and secret property source locators by making the…

### LINKED ISSUE #635: spring-cloud-kubernetes-config options and autoconfigure not working with reload.
**Describe the bug**

Attempting to configure  spring-cloud-kubernetes-config to just add support for kubernetes configMaps. (and disable secrets and discovery client) However would like to support config reload with event. It appears that when bringing up spring boot with the configuration below along with AutoConfigure, we run into a  bean resolution issue as posted below.

**Sample Configuration**

```
spring:
  cloud:
    kubernetes:
      config:
        enabled: true
      secrets:
        enabled: false
      discovery:
        enabled: false
      reload:
        enabled: true
        monitoring-config-maps: true
```

Exception:

```
Caused by: org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'org.springframework.cloud.kubernetes.config.reload.ConfigReloadAutoConfiguration$ConfigReloadAutoConfigurationBeans': Unsatisfied dependency expressed through field 'secretsPropertySourceLocator'; nested exception is org.springframework.beans.factory.NoSuchBeanDefinitionException: No qualifying bean of type 'org.springframework.cloud.kubernetes.config.SecretsPropertySourceLocator' available: expected at least 1 bean which qualifies as autowire candidate. Dependency annotations: {@org.springframework.beans.factory.annotation.Autowired(required=true)}
```

If the above seems a valid set of configs, Ideally, i would expect the configuration flags that either define or not define a given bean to be honored for autoconfigs as well.

Tested with:

```
Spring Boot: 2.3.3.RELEASE
spring-cloud-kubernetes-config: 1.1.3.RELEASE
```

krisiye:
…m optional dependencies for reload auto config as they could be enabled/disabled through config independent of each other. Fixes #635

**Describe the bug**

Attempting to configure  spring-cloud-kubernetes-config to just add support for kubernetes configMaps. (and disable secrets and discovery client) However would like to support config reload with event. It appears that when bringing up spring boot with the configuration below along with AutoConfigure, we run into a  bean resolution issue as posted below.

**Sample Configuration**

```
spring:
  cloud:
    kubernetes:
      config:
        enabled: true
      secrets:
        enabled: false
      discovery:
        enabled: false
      reload:
        enabled: true
        monitoring-config-maps: true
```

Exception:

```
Caused by: org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'org.springframework.cloud.kubernetes.config.reload.ConfigReloadAutoConfiguration$ConfigReloadAutoConfigurationBeans': Unsatisfied dependency expressed through field 'secretsPropertySourceLocator'; nested exception is org.springframework.beans.factory.NoSuchBeanDefinitionException: No qualifying bean of type 'org.springframework.cloud.kubernetes.config.SecretsPropertySourceLocator' available: expected at least 1 bean which qualifies as autowire candidate. Dependency annotations: {@org.springframework.beans.factory.annotation.Autowired(required=true)}
```

If the above seems a valid set of configs, Ideally, i would expect the configuration flags that either define or not define a given bean to be honored for autoconfigs as well.

Tested with:

```
Spring Boot: 2.3.3.RELEASE
spring-cloud-kubernetes-config: 1.1.3.RELEASE
```

### DIFF
--- a/spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/reload/ConfigReloadAutoConfiguration.java
+++ b/spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/reload/ConfigReloadAutoConfiguration.java
@@ -25,6 +25,7 @@ import org.springframework.boot.actuate.autoconfigure.endpoint.EndpointAutoConfi
 import org.springframework.boot.actuate.autoconfigure.info.InfoEndpointAutoConfiguration;
 import org.springframework.boot.autoconfigure.AutoConfigureAfter;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnExpression;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
@@ -84,6 +85,7 @@ public class ConfigReloadAutoConfiguration {
 		 */
 		@Bean
 		@ConditionalOnMissingBean
+		@ConditionalOnExpression("${spring.cloud.kubernetes.config.enabled:true} or ${spring.cloud.kubernetes.secrets.enabled:true}")
 		public ConfigurationChangeDetector propertyChangeWatcher(
 				ConfigReloadProperties properties, ConfigurationUpdateStrategy strategy) {
 			switch (properties.getMode()) {
--- a/spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/reload/ConfigReloadAutoConfiguration.java
+++ b/spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/reload/ConfigReloadAutoConfiguration.java
@@ -71,10 +71,10 @@ public class ConfigReloadAutoConfiguration {
 		@Autowired
 		private KubernetesClient kubernetesClient;
 
-		@Autowired
+		@Autowired(required = false)
 		private ConfigMapPropertySourceLocator configMapPropertySourceLocator;
 
-		@Autowired
+		@Autowired(required = false)
 		private SecretsPropertySourceLocator secretsPropertySourceLocator;
 
 		/**
--- a/spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/reload/ConfigReloadAutoConfiguration.java
+++ b/spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/reload/ConfigReloadAutoConfiguration.java
@@ -24,8 +24,8 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.actuate.autoconfigure.endpoint.EndpointAutoConfiguration;
 import org.springframework.boot.actuate.autoconfigure.info.InfoEndpointAutoConfiguration;
 import org.springframework.boot.autoconfigure.AutoConfigureAfter;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
-import org.springframework.boot.autoconfigure.condition.ConditionalOnExpression;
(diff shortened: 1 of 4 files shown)


krisiye:
@ryanjbaxter  - For some reason, i could not get a new test to pass -  https://github.com/spring-cloud/spring-cloud-kubernetes/pull/639/files#diff-cc127b21a34a91ae31d8e30c83f3d673R89. However in local testing appear to work as expected. Suggestions are welcome.

spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/reload/ConfigReloadAutoConfiguration.java
```diff
@@ -71,32 +72,31 @@
...
 		 * @param strategy configuration update strategy
 		 * @return a bean that listen to configuration changes and fire a reload.
 		 */
 		@Bean
 		@ConditionalOnMissingBean
+		@ConditionalOnBean({ ConfigMapPropertySourceLocator.class,
+				SecretsPropertySourceLocator.class })
 		public ConfigurationChangeDetector propertyChangeWatcher(
-				ConfigReloadProperties properties, ConfigurationUpdateStrategy strategy) {
+				ConfigReloadProperties properties, ConfigurationUpdateStrategy strategy,
```
ryanjbaxter, line 85: No need for the `@Autowired`

spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/reload/ConfigReloadAutoConfiguration.java
```diff
@@ -71,32 +72,31 @@
...
+				@Autowired(
+						required = false) ConfigMapPropertySourceLocator configMapPropertySourceLocator,
+				@Autowired(
+						required = false) SecretsPropertySourceLocator secretsPropertySourceLocator) {
+			System.out.println("init propertyChangeWatcher");
```
ryanjbaxter, line 90: Remove this

spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/reload/EventBasedConfigurationChangeDetector.java
```diff
@@ -66,7 +66,8 @@ public EventBasedConfigurationChangeDetector(AbstractEnvironment environment,
 	public void watch() {
 		boolean activated = false;
 
-		if (this.properties.isMonitoringConfigMaps()) {
+		if (this.properties.isMonitoringConfigMaps()
+				&& this.configMapPropertySourceLocator != null) {
```
ryanjbaxter, line 70: `null` checks wont be necessary if we are only creating these if the bean is present

ryanjbaxter (changes requested):

spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/reload/EventBasedConfigurationChangeDetector.java
krisiye: I put them in just because it is an either or. So you could have one or the other, and both watchers are being registered under the same changeDetector.

spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/reload/ConfigReloadAutoConfiguration.java
krisiye: Without making that optional, it wont work for the configs i posted originally. Since you would potentially only have one or the other or both or none.

krisiye:
@ryanjbaxter  - was able to fix the test cases. Added a few more dependencies to the spring context used by the tests.

ryanjbaxter (approved):
Thanks for working on this.  I was missing the fact that you could create one PropertySourceLocator but not the other.  I would really like to refactor the way this works and have a `ConfigurationChangeDetector` for both ConfigMaps and Secrets (we would have 4 total I think because we polling and event based implementations).  That would make some of this bean creation logic simpler.  I don't want to do that in the 1.1.x branch but if you happen to want to take a stab at that in master please do!

PR #644, opened 2020-09-26
Refactor ConfigurationChangeDetector to simplify conditionals. Issue 643

### LINKED ISSUE #643: Refactor ConfigurationChangeDetector to simplify conditionals
As part of #635 we fixed ConfigurationChangeDetector to honor conditionals where either config or secret could be enabled/disabled or vice-versa. This adds a level of complexity and null checking that could be improved upon. Refactor so that we register a ConfigurationChangeDetector each for config/secret and per mode (Polling/Event). Simplify code and more readable.

krisiye:
…ge detectors each for reload detection mode event/polling. Fixes #643

1. Adds a changeDetector each for configMaps/Secrets and event/polling modes
2. Moves reload mode conditionals into a sub package, usable by both autoConfiguration and defaults
3. Additional tests
4. updates to logging to indicate event/polling watcher activation

### DIFF
--- a/spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/reload/ConfigReloadAutoConfiguration.java
+++ b/spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/reload/ConfigReloadAutoConfiguration.java
@@ -24,7 +24,6 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.actuate.autoconfigure.endpoint.EndpointAutoConfiguration;
 import org.springframework.boot.actuate.autoconfigure.info.InfoEndpointAutoConfiguration;
 import org.springframework.boot.autoconfigure.AutoConfigureAfter;
-import org.springframework.boot.autoconfigure.condition.AnyNestedCondition;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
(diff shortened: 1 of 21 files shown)


krisiye:
@ryanjbaxter  - This is an attempt to refactor the configuration watcher implementation. It also sounds like based on the docs, the watcher as a side-car is the recommended approach as opposed to the native reload mechanism that is deprecated (wonder why? would benefit from some more background on this.) . May be best to also update the docs to reflect this and make it more obvious in the context of all of the reload config options available.

ryanjbaxter:
@krisiye wow!  Thanks for this! Amazing work!

The reasons for the controller based approach is because using the Kubernetes API by applications running on Kubernetes is generally discouraged.  Most of what spring-cloud-kubernetes-config did was read ConfigMaps and Secrets from Kubernetes and creating PropertySources from them.  That piece of functionality can easily be replaced by mounting the ConfigMaps and Secrets as volumes.  The piece of functionality that is missing when mounting them as volumes is notifying the app that the configuration changed and should be refreshed.  This is where the new controller comes in.  It is the one watching the ConfigMaps and Secrets for changes and then uses Spring Cloud Bus to notify the app of the changes.  

Does that make sense?

krisiye:
Thanks @ryanjbaxter  That makes a lot of sense! Decoupling from app and running as a sidecar seems more scalable! For the bus implementation, It may also make sense to support other messaging implementations in the future. For example: kafka

ryanjbaxter:
@krisiye yes, plan is to support everything bus supports!  (Again, if you have any interest in helping out there we would appreciate it!). Could you fix the merge conflicts and update the PR?

krisiye:
@ryanjbaxter  - Fixed conflicts. Should be good to go. CI is failing for unrelated modules looking for dependencies. Possible to re-run? ii am unable to re-run on my end. missing permissions.

krisiye:
@ryanjbaxter  The watcher (controller) also needs some improved logging both info and debug. I'll work on those and log issues.

ryanjbaxter (approved):
This looks good to me.  I also did some testing on my own and everything seemed to work.

@spencergibb @Haybu if you have a second could you take a look as well?

Haybu (approved):

PR #446, opened 2019-08-08
Feature/retry policy configmap property source

anavidad3:
### What changes were proposed in this pull request?

About issue #441.  This pull request offers the possibility of defining a retry policy to read kubernetes configmaps to avoid some ephemeral network problems.

This closes #441

Issue #441, opened 2019-07-31
ConfigMapPropertySource can't read configMap. Retry Policy?


anavidad3:
Related with this Issue #400 

Could a retry policy be sense to read ConfigMap? 

We work with:
spring-cloud-starter-kubernetes-config 1.0.2.RELEASE version.
gke cluster 1.12.8-gke.10 version.

We have a microservice(MS) where we read a configMap. **Most of the time the configmap load is successful** but sometimes a warning message like this appears:

`2019-07-30 10:18:52,184 [main]WARN  org.springframework.cloud.kubernetes.config.ConfigMapPropertySource - Can't read configMap with name: [employee-ms] in namespace:[test]. Ignoring`

In our case, load of external configuration in our MS is mandatory. 

As we didn't know what the error might be, we made a few changes to the spring-cloud-kubernetes-config.

1) We add the exception to the warning trace.

We discover that the error was for a connection refuse

`2019-07-30 11:38:52,184 [main] WARN  org.springframework.cloud.kubernetes.config.ConfigMapPropertySource - Can't read configMap with name: [employee-ms] in namespace:[test]. Ignoring
io.fabric8.kubernetes.client.KubernetesClientException: Operation: [get]  for kind: [ConfigMap]  with name: [employee-ms]  in namespace: [test]  failed.
    at io.fabric8.kubernetes.client.KubernetesClientException.launderThrowable(KubernetesClientException.java:64)
    at io.fabric8.kubernetes.client.KubernetesClientException.launderThrowable(KubernetesClientException.java:72)
    at io.fabric8.kubernetes.client.dsl.base.BaseOperation.getMandatory(BaseOperation.java:221)
    at io.fabric8.kubernetes.client.dsl.base.BaseOperation.get(BaseOperation.java:177)
    at org.springframework.cloud.kubernetes.config.ConfigMapPropertySource.getData(ConfigMapPropertySource.java:101)
    at org.springframework.cloud.kubernetes.config.ConfigMapPropertySource.<init>(ConfigMapPropertySource.java:77)
    at org.springframework.cloud.kubernetes.config.ConfigMapPropertySourceLocator.getMapPropertySourceForSingleConfigMap(ConfigMapPropertySourceLocator.java:95)
    at org.springframework.cloud.kubernetes.config.ConfigMapPropertySourceLocator.lambda$locate$0(ConfigMapPropertySourceLocator.java:78)
[...]

PR #648, opened 2020-09-29
Allow loading of config maps to be retried on failure

sagacity:
I'm trying to solve https://github.com/spring-cloud/spring-cloud-kubernetes/issues/441. There was already a pull request (https://github.com/spring-cloud/spring-cloud-kubernetes/pull/446) for this, but it stalled. Instead of rebasing that PR I've tried to simply add a wrapper around spring-retry, since the discussion in that pull request seemed to indicate this is a valid approach.

Can you confirm this approach would be ok for you? If so, I will spend some more time on the PR adding appropriate tests. Otherwise I'm open to changing directions.

The reason I didn't try to abstract away `spring-retry` is that it's a fairly well known project in the Spring ecosystem and it felt unnecessary to hide that. The configuration is set up so that if the dependency is not available retry logic will simply be unavailable and the old behaviour will be used.

### DIFF
--- a/spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/BootstrapConfiguration.java
+++ b/spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/BootstrapConfiguration.java
@@ -16,6 +16,8 @@
 
 package org.springframework.cloud.kubernetes.config;
 
+import java.util.Optional;
+
 import io.fabric8.kubernetes.api.model.ConfigMap;
 import io.fabric8.kubernetes.api.model.Secret;
 import io.fabric8.kubernetes.client.KubernetesClient;
@@ -30,8 +32,6 @@ import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.Import;
 import org.springframework.retry.support.RetryTemplate;
-
-import java.util.Optional;
 
 /**
  * Auto configuration that reuses Kubernetes config maps as property sources.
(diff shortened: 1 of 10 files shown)


RoyJacobs:
A test is failing on CircleCI, but this seems unrelated to my change (a `KubernetesClientTimeoutException`)?

RoyJacobs:
Any updates on this?

spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/BootstrapConfiguration.java
```diff
@@ -49,8 +53,10 @@
 
 		@Bean
 		@ConditionalOnProperty(name = "spring.cloud.kubernetes.config.enabled", matchIfMissing = true)
-		public ConfigMapPropertySourceLocator configMapPropertySourceLocator(ConfigMapConfigProperties properties) {
-			return new ConfigMapPropertySourceLocator(this.client, properties);
+		public ConfigMapPropertySourceLocator configMapPropertySourceLocator(ConfigMapConfigProperties properties,
```
ryanjbaxter, line 56: Instead of the `Optional` here I think it might be cleaner if you used `@ConditionalOnMissingClass`

ryanjbaxter (commented):
Awesome, thanks for this PR!

I added support for reading ConfigMaps with the Kuberenetes Java client which unfortunately has caused a bunch of merge conflicts.  Could we resolve those and maybe take a look at adding the same type of retry logic there as well?

Also could the same functionality be added for secrets?

RoyJacobs:
@ryanjbaxter I am currently on paternity leave but hopefully I can take another look in the new year!

ryanjbaxter:
Great!

RoyJacobs:
@ryanjbaxter I'm back from leave but unfortunately not in the position to continue work. @vy @breun any chance of picking this up?

ryanjbaxter:
No worries, I can probably take a look in a few weeks.

wind57:
@ryanjbaxter it seems this can be closed also, as we already have retry thx to the excellent effort from @isikerhan

PR #649, opened 2020-09-30
Refactor spring-cloud-core To Allow For Alternative Implementation

ryanjbaxter:
This PR basically adds abstractions so we can being to provide alternative implementations to Fabric8io in other modules


pom.xml
```diff
@@ -87,6 +87,7 @@
 
 	<modules>
 		<module>spring-cloud-kubernetes-dependencies</module>
+		<module>spring-cloud-kubernetes-commons</module>
 		<module>spring-cloud-kubernetes-core</module>
```
spencergibb, line 91: I don't think we need to deal with it right this second, but commons and core seem similar. Is core now fabric8 specific?

Haybu:
LGTM. Well done. 

Subjective cosmetic thought:

- Wonder why the original `profile` sub-package in `core`  and now in `common` and not just its one file each to be outside along with the others.

- And what if instead of a separate common module, to have a `common` sub-package in the existing `core` module, with current fabric8 core classes to go into a new sub-package `fabrtic8 ` (or any other name that we later just delete), and the upcoming k8s java client impl to go into its own new sub-package. So one core sub-module.

ryanjbaxter:
@Haybu 
> Wonder why the original profile sub-package in core and now in common and not just its one file each to be outside along with the others.

Good question....I can move it now I suppose.

>And what if instead of a separate common module, to have a common sub-package in the existing core module, with current fabric8 core classes to go into a new sub-package fabrtic8 (or any other name that we later just delete), and the upcoming k8s java client impl to go into its own new sub-package. So one core sub-module.

I chose to structure it that way so it is clear which module is using which client rather than have a mixture if clients in one module.  Then we can decide which module is used via starters.  The question we need to answer is the naming conventions we want to use.  Right now I am thinking `spring-cloud-kubernetes-fabric8-core` and `spring-cloud-kubernetes-k8sclient-core` but I am open to suggestions as well.

Haybu:
> Then we can decide which module is used via starters

 is the idea to have a starter per client used?

Haybu (approved):

ryanjbaxter:
@Haybu thats my thought at the moment, yes

PR #663, opened 2020-10-21
fix a bug when there are more than one ports

decylus:
fix a bug when there are more than one ports in a k8s service.

for example, if there are 2 ports in my service, one for the tomcat named default, one for other purpose named other
The request can be only send to the default port.

so, I have a config 'ribbon.PortName=default' which will be set in the KubernetesServerList.portName.

In the origin code, 'Utils.isNotNullOrEmpty(this.getPortName())' means once I have this config, all ports will be add to the KubernetesServicesServerList.
so what I set in the ribbon.PortName has no effects.

Changing 'Utils.isNotNullOrEmpty(this.getPortName())'  to  'Utils.isNullOrEmpty(this.getPortName())' will fix that.
Only the portName match what I set in config will be added to the ServerList.

### DIFF
--- a/spring-cloud-kubernetes-ribbon/src/main/java/org/springframework/cloud/kubernetes/ribbon/KubernetesServicesServerList.java
+++ b/spring-cloud-kubernetes-ribbon/src/main/java/org/springframework/cloud/kubernetes/ribbon/KubernetesServicesServerList.java
@@ -76,7 +76,7 @@ public class KubernetesServicesServerList extends KubernetesServerList {
 			}
 			else {
 				for (ServicePort servicePort : service.getSpec().getPorts()) {
-					if (Utils.isNotNullOrEmpty(this.getPortName())
+					if (Utils.isNullOrEmpty(this.getPortName())
 							|| this.getPortName().endsWith(servicePort.getName())) {
 						result.add(new Server(concatServiceFQDN(service),
 								servicePort.getPort()));
--- a/spring-cloud-kubernetes-ribbon/src/test/java/org/springframework/cloud/kubernetes/ribbon/KubernetesServiceServerListTest.java
+++ b/spring-cloud-kubernetes-ribbon/src/test/java/org/springframework/cloud/kubernetes/ribbon/KubernetesServiceServerListTest.java
@@ -1,0 +1,115 @@
+package org.springframework.cloud.kubernetes.ribbon;
+
+import com.netflix.loadbalancer.Server;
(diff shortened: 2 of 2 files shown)


spring-cloud-kubernetes-ribbon/src/main/java/org/springframework/cloud/kubernetes/ribbon/KubernetesServicesServerList.java
```diff
@@ -76,7 +76,7 @@ private String concatServiceFQDN(Service service) {
 			}
 			else {
 				for (ServicePort servicePort : service.getSpec().getPorts()) {
-					if (Utils.isNotNullOrEmpty(this.getPortName())
```
ryanjbaxter, line 79: It seems like this should be `if (Utils.isNotNullOrEmpty(this.getPortName()) && this.getPortName().endsWith(servicePort.getName()))`

Maybe a unit test would help clarify things?

ryanjbaxter (changes requested):

rainwolf01:
bug still in 1.1.8.RELEASE 😭

ryanjbaxter:
@rainwolf01 can you provide a sample that reproduces the problem?

I am not sure if @decylus is still willing to work on this, but if I have a sample to reproduce the problem I can probably make the right fix.

spring-cloud-kubernetes-ribbon/src/main/java/org/springframework/cloud/kubernetes/ribbon/KubernetesServicesServerList.java
decylus: Hi, I have add a unit test to explain this.

decylus:
@ryanjbaxter I have add a unit test to explain this, please check it. thank you.

ryanjbaxter (changes requested):
@decylus Thanks this makes sense now.  Can you remove some of the comments from the unit test and fix the checkstyle errors and we can merge this fix?

decylus:
> @decylus Thanks this makes sense now. Can you remove some of the comments from the unit test and fix the checkstyle errors and we can merge this fix?

Of crouse,  done~

create1072627512:
ribbon.PortName=default,   Where to set it? I loaded the configuration file directly and it didn't work

ryanjbaxter:
If you are having an issue please open a separate issue

PR #664, opened 2020-10-21
Refactor Config Module And Add A New Config Module Using Kubernetes Client

ryanjbaxter:
* Extract common code to commons module
* Rename config module to be specific to fabric8
* Add new module that implements config functionality using the Kubernetes Java Client
* Adds integration test for new config module


spring-cloud-kubernetes-fabric8-config/src/main/java/org/springframework/cloud/kubernetes/fabric8/config/Fabric8ConfigMapPropertySource.java
```diff
@@ -0,0 +1,94 @@
...
+			Environment environment) {
+		super(getName(name, getNamespace(client, namespace)),
+				asObjectMap(getData(client, name, getNamespace(client, namespace), environment)));
+	}
+
+	private static String getNamespace(KubernetesClient client, String namespace) {
+		return StringUtils.isEmpty(namespace) ? client.getNamespace() : namespace;
+	}
+
+	private static Map<String, Object> getData(KubernetesClient client, String name, String namespace,
```
spencergibb, line 57: Maybe this should be an abstract method rather than a constructor requirement?

spencergibb (approved):

Haybu (approved):

spring-cloud-kubernetes-fabric8-config/src/main/java/org/springframework/cloud/kubernetes/fabric8/config/Fabric8ConfigMapPropertySource.java
ryanjbaxter: Not sure how we would do that, we need the data in order to call the super constructor

ryanjbaxter:
If anyone has any ideas about how to best reuse the integration test I wrote for both fabric 8 an the k8s client I would love to hear your thoughts.  the only difference would be the dependency on the classpath.

Issue #584, opened 2020-07-20
Optional property to get notReadyAddress instances.


claitonp:
Hi,

You could add an optional property to get the **notReadyAddress** instances in the classes:

1. KubernetesDiscoveryClient - method getNamespaceServiceInstances ();
2. KubernetesCatalogWatch - method catalogServicesWatch ()

Example:  spring.cloud.kubernetes.discovery.notReadyAddress.enabled,  default false.

```
private List<ServiceInstance> getNamespaceServiceInstances(EndpointSubsetNS es, String serviceId) {
	...
	List<EndpointAddress> addresses = s.getAddresses();
	for (EndpointAddress endpointAddress : addresses) {
		...
	}
	
       // HERE /////////////////////////////
	if (notReadyAddressEnabled) 
		List<EndpointAddress> notReadyAddresses = s.getNotReadyAddresses();
		for (EndpointAddress endpointAddress : notReadyAddresses) {  // the same logic addresses.
		   ...
		}
	}
        //////////////////////////

	return instances;  // return instances addresses e notReadyAddresses.
}
```

Thanks
Claiton

spencergibb:
Can you please describe why you would want this?

claitonp:
For better monitoring integration. For example with spring boot admin and spring cloud kubernentes, to get the bad instances.

spencergibb:
I'm afraid I don't see how that is helpful. Why would a user of spring cloud kubernetes want the bad instances?

claitonp:
In Spring Cloud Gateway I use Spring Cloud Kubernetes to get only good instances and two other applications.

I also use Spring Cloud Kubernetes in the SBA and in this one it would be interesting to get the good and bad instances, for a better visualization on the wallboard.

Bad instances using Spring Cloud Kubernetes are removed from the wallboard view.

[wallboard ](https://user-images.githubusercontent.com/8779225/37558711-84f93dc2-2a18-11e8-8e1b-10ad60f8e80a.png)

stevengibbszions:
@spencergibb  - I agree that this would be a useful feature.  When monitoring spring boot microservices with a tool like Spring Boot Admin, you could maintain visibility of a service endpoints (and the reason's why its failing probes, for example) if the service discovery mechanism returned the bad endpoints also.  That endpoint might be "live" but not "ready".  

In the current state, if the pod goes to "not ready" then you lose visibility.

I'd be interested in doing a PR if you were open to the idea.

[...]

PR #673, opened 2020-11-14
Allow opt-in for discovering "not-ready" service endpoints

srgibbs99:
Discovering "not-ready" service endpoints will help tools like Spring Boot Admin continue to monitor all endpoints, ready or not.

Should resolve https://github.com/spring-cloud/spring-cloud-kubernetes/issues/584

### DIFF
--- a/docs/src/main/asciidoc/discovery-client.adoc
+++ b/docs/src/main/asciidoc/discovery-client.adoc
@@ -59,7 +59,7 @@ To discover service endpoint addresses that are not marked as "ready" by the kub
 ----
 spring.cloud.kubernetes.discovery.include-not-ready-addresses=true
 ----
-NOTE: This might be useful when discovering services for monitoring purposes, using a tool like https://codecentric.github.io/spring-boot-admin/current/[spring-boot-admin].
+NOTE: This might be useful when discovering services for monitoring purposes, and would enable inspecting the `/health` endpoint of not-ready service instances.
 ====
 
 If, for any reason, you need to disable the `DiscoveryClient`, you can set the following property in `application.properties`:
--- a/docs/src/main/asciidoc/discovery-client.adoc
+++ b/docs/src/main/asciidoc/discovery-client.adoc
@@ -52,6 +52,16 @@ spring.cloud.kubernetes.discovery.all-namespaces=true
 ----
 ====
 
+To discover service endpoint addresses that are not marked as "ready" by the kubernetes api server, you can set the following property in `application.properties` (default: false):
+
+====
+[source]
+----
+spring.cloud.kubernetes.discovery.include-not-ready-addresses=true
+----
+NOTE: This might be useful when discovering services for monitoring purposes, using a tool like https://codecentric.github.io/spring-boot-admin/current/[spring-boot-admin].
+====
+
 If, for any reason, you need to disable the `DiscoveryClient`, you can set the following property in `application.properties`:
 
 ====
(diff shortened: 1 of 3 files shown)


spencergibb:
I'd argue against this change. DiscoverClient is not a monitoring tool.

srgibbs99:
> I'd argue against this change. DiscoveryClient is not a monitoring tool.

Agree with that, but the option to discover not ready clients will allow other monitoring tools (that rely on spring cloud kubernetes to discover endpoints) to have all information needed.

To me this is a **discovery** concern.  Let the library user opt-in if they need/want the not-ready addresses.

spencergibb:
We'll see what the team thinks. If we do move forward, I'd remove any reference to a third party project that we don't have specific integration with.

ryanjbaxter:
I guess the problem I see with this is that the information is pretty much stale as soon as you retrieve it.  It seems like Spring Cloud Admin would want to know when the Pod becomes ready.  Issuing something like a watch event on the pod might be more useful in that case.

srgibbs99:
> I guess the problem I see with this is that the information is pretty much stale as soon as you retrieve it. It seems like Spring Cloud Admin would want to know when the Pod becomes ready. Issuing something like a watch event on the pod might be more useful in that case.

I believe it does exactly this, at some configured frequency, via Spring Cloud Kubernetes discovery.  You certainly need to know information as up to date as possible, and that applies to both ready and not-ready service instances.

This really feels like a **discovery** concern to me, and that's a feature that Spring Cloud Kubernetes delivers.  This small change could make it that much more useful.

ryanjbaxter:
Can you describe how Spring Boot Admin would use this feature?

srgibbs99:
> Can you describe how Spring Boot Admin would use this feature?

Spring Boot Admin can already be configured to use spring cloud kubernetes for service discovery.  It uses the list of discovered endpoints for a particular service to probe the spring boot actuator endpoints of each service instance.  It can show the health status and lots of other information as exposed by actuator.  The problem is that it can only discover "healthy" endpoints, therefore defeating one of its purposes, to report in the health of services.  With this feature, it would be able to discover both ready and not-ready service instances, giving visibility into "live" but "not-ready" services.   

For example, If a certain service instance (pod) is only failing readiness check, but remains alive, spring boot admin could continue to display status and the underlying reason for failing the readiness probe.  Whereas now it simply is removed from the list services altogether and is assumed _offline_.

https://codecentric.github.io/spring-boot-admin/current/#discover-clients-via-spring-cloud-discovery

ryanjbaxter:
But what happens is the Pod then becomes ready?  How will you reflect that change in state?

srgibbs99:
> But what happens is the Pod then becomes ready? How will you reflect that change in state?

Each service instance exposes a `/health` REST endpoint that returns an status of `UP` or `DOWN` and SB Admin probes each service instance using that endpoint to determine status, and then reports that in the UI.  Its actually the same endpoint that we typically use for the "readiness" probe in k8s.  If the "not-ready" instances are never discovered then there's no way to report on them either way.

So if all my service instances are "not-ready" the list that currently gets discovered is empty, and therefore SB admin has to assume there are 0 service endpoints and removes it from monitoring altogether.  That's same behavior as if I purposefully scaled my deployment to 0 replicas.  Clearly those 2 scenarios are unique and should be treated differently by a tool like this.

I totally agree that including not-ready endpoints should not be the default behavior.  But I feel like allowing an opt-in to discover those instances serves a valid purpose for monitoring tools that want to use spring-cloud-kubernetes.  All the heavy lifting is already done with connecting to the k8s api, etc.  One could certainly write their own API client code to go get not-ready instances, but if I do that, why would I use spring-cloud-kubernetes for discovery at all?  The data is there in API, let the library user decide if its valid for them.

ryanjbaxter:
Ultimately this is not going to add a ton of overhead and is off by default, so I don't see much harm.  I am OK with the change.

srgibbs99:
@ryanjbaxter Any idea when the v2.0+ releases will make their way to Maven Central?

image

spencergibb:
They are already there since December. I think the artifact id changed

srgibbs99:
> They are already there since December. I think the artifact id changed

Oh.  Does this look like the new one?

https://mvnrepository.com/artifact/org.springframework.cloud/spring-cloud-starter-kubernetes-fabric8-all

ryanjbaxter:
Yes that is it

Issue #613, opened 2020-09-02
KubernetesServiceInstanceMapper may throw NPE when service.yml miss labels and annotations


EvanPro:
When I apply a service without labels and annotations properties like
`
apiVersion: v1
kind: Service
metadata:
  name: hostnames
spec:
  selector:
    app: hostnames
  ports:
  - name: default
    protocol: TCP
    port: 80
    targetPort: 9376
`

KubernetesServiceInstanceMapper will throw NullPointException.

If these properties are required to enabe load balancing based on Kubernetes Service name, then it should be specified in the documentation.

https://github.com/spring-cloud/spring-cloud-kubernetes/blob/5882245cb5f786feef18845c7fc469b7e9c822e2/spring-cloud-kubernetes-loadbalancer/src/main/java/org/springframework/cloud/kubernetes/loadbalancer/KubernetesServiceInstanceMapper.java#L108-L113

EvanPro:
> Can you provide a [complete, minimal, verifiable sample](https://stackoverflow.com/help/mcve) that reproduces the problem? It should be available as a GitHub (or similar) project or attached to this issue as a zip file.

I‘ve create a demo project for this issue：https://github.com/EvanPro/npedemo
Apply a service in my kubernetes cluster like below.
Call http://localhost:8088/hostnames, the code will throw npe

madaram:
I exactly have the same problem, see NPE stack trace below. 

Quick and dirty workaround is to add dummy annotations in Kubernetes services
```yml
metadata:
  name: uaa
  annotations:
    test: test
```
NPE location :
```
java.lang.NullPointerException: null
	at org.springframework.cloud.kubernetes.loadbalancer.KubernetesServiceInstanceMapper.isSecure(KubernetesServiceInstanceMapper.java:114)
	at org.springframework.cloud.kubernetes.loadbalancer.KubernetesServiceInstanceMapper.map(KubernetesServiceInstanceMapper.java:70)
...
```

PR #675, opened 2020-11-24
trivial NPE when no annotations or labels to fix gh-613

madaram:
Add not null check to avoid null pointer exception on kubernetes objects without labels or annotations in metadata. Fixes gh-613


spencergibb (changes requested):
Please add a test so we can catch regressions

ryanjbaxter:
I have a fix for this in a branch that I will commit this afternoon.

Issue #654, opened 2020-10-07
Support Kafka for Spring Cloud Kubernetes Configuration Watcher


krisiye:
**Is your feature request related to a problem? Please describe.**
Spring Cloud Kubernetes Configuration Watcher supports RabbitMQ only. Spring Cloud Bus however supports both rabbitMQ and kafka. Would be great to expand on the configuration watcher capability to support kafka.

**Describe the solution you'd like**
Support Kafka For the Bus integration under Spring Cloud Kubernetes Configuration Watcher

**Describe alternatives you've considered**
Alternatives are Either using HTTP watcher or RabbitMQ.

**Additional context**
Currently the watcher supports a single bus profile and defaults to AMQP/Rabbit. One way to potentially implement this would be to introduce a default binder associated to a bus profile each per spring cloud bus implementations (amqp/kafka).

PR #676, opened 2020-11-25
fix conflicts upon merge

krisiye:
Refactor config watcher to resolve conflicts upon merge to master.


krisiye:
@ryanjbaxter  - Looks like the #654  was out of sync and did not catch this as a conflict either. This should fix master though.

PR #682, opened 2020-12-09
Minor cleanup in pom files

wind57:


wind57:
`groupId` is inherited, a very minor clean-up.

PR #686, opened 2020-12-09
simplify some conditions and add tests

wind57:


wind57:
rather interesting that I am fighting `ConfigMapAndSecretIT` for the second day now

wind57:
I am going to open a different one, there are already way too many commits here

PR #692, opened 2020-12-13
minor simplification

wind57:


wind57:
trivial simplifications, imho.

PR #695, opened 2020-12-15
Shorten IT: wait conditions in parallel

yue9944882:


spring-cloud-kubernetes-test-support/src/main/java/org/springframework/cloud/kubernetes/integration/tests/commons/K8SUtils.java
```diff
@@ -172,11 +173,20 @@ public void waitForDeployment(String deploymentName, String namespace) {
...
 		await().timeout(
-				Duration.ofSeconds(90)).until(
-						() -> appsApi
-								.listNamespacedDeployment(namespace, null, null, null,
-										"metadata.name=" + deploymentName, null, null, null, null, null)
-								.getItems().isEmpty());
+			Duration.ofSeconds(90)).until(
+			() -> {
+				try {
+					appsApi
```
ryanjbaxter, line 179: What is the advantage of doing it this way?

yue9944882: the original setup flow is working like "deploy" -> "wait" -> "deploy" -> "wait" ->..., this pull makes "deploy" and "wait" work in parallel, wdyt?

ryanjbaxter: I mean the change from `listNamespaceDeployment` to `readNamespacedDeplayment`, what is the advantage?

yue9944882: kube-apiserver will be loading all deployments from etcd into memory and filtering w/ the given selector `metata.name` which has a time cost of `O(# of list items)` upon `listNamespaceDeployment`, while `readNamespacedDeplayment` will only load single object. the difference also applies to `deleteCollection` and single `delete`.

PR #696, opened 2020-12-15
some refactoring

wind57:


wind57:
This PR is broken into separate commits for easier review.

# First PR includes minor code changes:
 
- `final` to arguments set in the constructor
- method reference instead of lambda
- removed `synchronized` from `internalGetPod`, as `LazilyInstantiate::get` already takes care of that
- moved common branches return statement (`return null`) into a single place
- inverted `isHostNameEnvVarPresent` and `isServiceAccountFound` because [a JMH test](https://github.com/wind57/0x-github-jmh-runner/pull/4/checks?check_run_id=1554720292) shows that accessing the ENV si around 40 times faster then checking if the file exists, so if we want to fail early, it makes sense to invert those

PR #698, opened 2020-12-15
slight changes in health indicator

wind57:


wind57:
nothing big:

 - use `Collections.singletonMap`
 - use pre-sized `HashMap` according to : https://github.com/spring-projects/spring-framework/pull/25349
 - `final` added
 - removed `throws Exception` as it is not needed

PR #700, opened 2020-12-18
Loadbalancer Module Using Kubernetes Java Client

ryanjbaxter:


spring-cloud-kubernetes-client-discovery/src/main/java/org/springframework/cloud/kubernetes/client/discovery/KubernetesDiscoveryClientAutoConfiguration.java
```diff
@@ -64,10 +62,6 @@ public KubernetesInformerFactoryProcessor kubernetesInformerFactoryProcessor() {
 		@Bean
 		@ConditionalOnMissingBean
 		public CatalogSharedInformerFactory catalogSharedInformerFactory(ApiClient apiClient) {
-			apiClient.getJSON()
-					.setGson(apiClient.getJSON().getGson().newBuilder()
-							.addDeserializationExclusionStrategy(new ServiceTrimmingStrategy())
```
ryanjbaxter, line 69: @yue9944882 was this added as just an efficiency?  It is causing problems because I need the spec from services.

yue9944882: @ryanjbaxter this is for easing memory pressure, deserializing watch events will take the most space from heap

ryanjbaxter: @yue9944882 maybe we should have the InformerDiscoveryClient use its own API client then?  If you have the discovery client module on the classpath than there is there is going to be no way to get the spec object from the service if you need it.  That caused an issue for me in this PR.

PR #701, opened 2020-12-19
Reactive DiscoveryClient Implementation Using Kubernetes Java Client

ryanjbaxter:


wind57:
@ryanjbaxter one minor question, is the general approach to move to kubernetes java client instead of fabric8?

ryanjbaxter:
@wind57 Yes we would like to get there, unless the community tells us otherwise

wind57:
> @wind57 Yes we would like to get there, unless the community tells us otherwise

thank you.

This also means that `spring-cloud-kubernetes-fabric8-autoconfig` will be replaced with `spring-cloud-kubernetes-client-autoconfig` in `spring-cloud-starter-kubernetes`, eventually, right?

Thing is I did some PRs against `spring-cloud-kubernetes-fabric8-autoconfig` and it seems they can be closed since this code will not be maintained in the long run and if I would like to do something it should happen in `spring-cloud-kubernetes-client-autoconfig`. Did I understand correctly?

ryanjbaxter:
> Did I understand correctly?

@wind57 No that is not entirely correct.

I saw the PRs but have not had a chance to look at them yet, hopefully tomorrow.

Right now we are maintaining both, so if you are fixing issues in the Fabric8 modules that is more than welcome.  If you are adding functionality then we would want that functionality to be present in both.

There are separate starters for the kubernetes java client and fabric8 client modules now.

wind57:
I see, I was more interested into understanding how things are rather than triggering you into looking the PRs, they are pretty minor anyway. I am mainly trying to understand the overall code and that is my way of doing it. thank you again for the answers.

PR #704, opened 2020-12-21
some additions in KubernetesClientUtils

wind57:


wind57:
changes are minor, but should help anyone debugging the problem from logs:

- rename to `Factory`, because this is one, even if it has a single method.
- use `ClientBuilder.defaultClient()` instead of `ClientBuilder.standard().build()`. It is documented to do the same exact thing.
- slightly change the logs when `cluster` mode fails:

```
LOG.info("Could not create the Kubernetes ApiClient in a cluster environment, because : ", e);
LOG.info("Trying to use a \"standard\" configuration to create the Kubernetes ApiClient");
```

At least, I personally (which is biased), find this far easier to understand what is going on.

- add a different log when `standard` client can be created:

```
LOG.info("Created standard API client. Unless $KUBECONFIG or $HOME/.kube/config is defined, "
					+ "this client will try to connect to localhost:8080");
```

For example, if we do not do the above and `standard` is created, the very next thing present in logs will be:

```
java.net.ConnectException: Failed to connect to localhost/127.0.0.1:8080
```

from some watcher kubernetes java client internal package. Which confuses the "reader" of logs if they face this for the first time. If we add an explicit log stating that the client connects to `localhost:8080` and even provide possible remediations and directions (via `$KUBECONFIG` or `$HOME/.kube/config`), this should make it far more clear and helpful.

- the next log message is slightly enhanced with the same reasoning as the previous one:  

```
LOG.warn("Could not create a Kubernetes ApiClient from either a cluster or standard environment. "
					+ "Will return one that always connects to localhost:8080", e1);
```

wind57:
The concern that I have is that this is a breaking change - since I have changed the name to `Factory`. If that is not acceptable, I will revert, of course. Just let me know.

ryanjbaxter:
Yes changing the class name won't be possible now

ryanjbaxter (approved):

ryanjbaxter (approved):

PR #706, opened 2020-12-21
EnvReader to commons

wind57:


wind57:
I want to move it to commons and use it everywhere, since it will me much easier to write tests against it, being impossible to mock `System` (at least with `Mockito`)

ryanjbaxter (approved):

PR #710, opened 2020-12-23
removed un-used package

wind57:


ryanjbaxter:
We can't remove these classes at this point, we will have to wait for the next major release next year

PR #711, opened 2020-12-23
SecurePortResolver clean-up

wind57:


wind57:
- I've renamed the class from `DefaultIsServicePortSecureResolver` to `ServicePortSecureResolver`. It used to make sense when this class implemented `IsServicePortSecureResolver`, but not anymore
- `log` to `LOG`
- `TRUTHY_STRINGS` build without the double brace initialization anti-pattern
- added documentation to the method instead of the class (and also removed the comment)
- the documentation does not mention `spring.cloud.kubernetes.discovery.secured` anymore. the class was updated not to take it into account, but documentation was left

wind57:
@ryanjbaxter this build fails because it can't pull `spring-cloud-kubernetes-configuration-watcher` with version `2.0.1-SNAPSHOT` (I did look in docker hub and it's not there). I also looked at how to push it, but I doubt I have rights for that. Can you take a look? thank you.

spring-cloud-kubernetes-fabric8-discovery/src/main/java/org/springframework/cloud/kubernetes/fabric8/discovery/KubernetesDiscoveryClientAutoConfiguration.java
```diff
@@ -48,8 +48,8 @@
 
 	@Bean
 	@ConditionalOnMissingBean
-	public DefaultIsServicePortSecureResolver isServicePortSecureResolver(KubernetesDiscoveryProperties properties) {
-		return new DefaultIsServicePortSecureResolver(properties);
+	public ServicePortSecureResolver isServicePortSecureResolver(KubernetesDiscoveryProperties properties) {
```
ryanjbaxter, line 51: I dont see why we need this defined as a bean, as since the class is not public no one could provide their own implementation anyways

ryanjbaxter (changes requested):

ryanjbaxter:
Let me now when this is ready

spring-cloud-kubernetes-fabric8-discovery/src/main/java/org/springframework/cloud/kubernetes/fabric8/discovery/KubernetesDiscoveryClientAutoConfiguration.java
wind57: indeed. changed.

wind57:
@ryanjbaxter ready.

ryanjbaxter:
OK I am on vacation this week, I will merge it next week when I get back, thanks!

PR #726, opened 2021-02-04
fix NullPointerException when primaryPortName is null

zhaozhihom:
line 132 throw NullPointerException if primaryPortName is not specified.

### DIFF
--- a/spring-cloud-kubernetes-client-discovery/src/main/java/org/springframework/cloud/kubernetes/client/discovery/KubernetesInformerDiscoveryClient.java
+++ b/spring-cloud-kubernetes-client-discovery/src/main/java/org/springframework/cloud/kubernetes/client/discovery/KubernetesInformerDiscoveryClient.java
@@ -83,7 +83,7 @@ public class KubernetesInformerDiscoveryClient implements DiscoveryClient, Initi
 	public List<ServiceInstance> getInstances(String serviceId) {
 		Assert.notNull(serviceId, "[Assertion failed] - the object argument must not be null");
 
-		if (StringUtils.hasText(namespace) && !properties.isAllNamespaces()) {
+		if (!StringUtils.hasText(namespace) && !properties.isAllNamespaces()) {
 			log.warn("Namespace is null or empty, this may cause issues looking up services");
 		}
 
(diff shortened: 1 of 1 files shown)


spring-cloud-kubernetes-client-discovery/src/main/java/org/springframework/cloud/kubernetes/client/discovery/KubernetesInformerDiscoveryClient.java
```diff
@@ -128,7 +128,8 @@ public String description() {
 			if (this.properties.getMetadata() != null && this.properties.getMetadata().isAddPorts()) {
 				subset.getPorts().stream().forEach(p -> metadata.put(p.getName(), Integer.toString(p.getPort())));
 			}
-			V1EndpointPort port = subset.getPorts() != null && subset.getPorts().size() == 1 ? subset.getPorts().get(0)
+			V1EndpointPort port = subset.getPorts() != null &&
+				(subset.getPorts().size() == 1 || this.properties.getPrimaryPortName() == null) ? subset.getPorts().get(0)
```
ryanjbaxter, line 132: I dont think this logic makes sense.  If there are multiple ports they will need names, so I think it would be safe to change the logic slightly to avoid the potential NPE 

`p.getName().equasIgnoreCase(this.properties.getPrimaryPortName())`

ryanjbaxter (changes requested):

spring-cloud-kubernetes-client-discovery/src/main/java/org/springframework/cloud/kubernetes/client/discovery/KubernetesInformerDiscoveryClient.java
zhaozhihom: Okay, that would work.  
The other problem is the code get exceptions when there are multiple ports and none of them matches with primaryPortName. I spent half an hours to figure it out. It should be better to provide more information in the log.

zhaozhihom: > I dont think this logic makes sense. If there are multiple ports they will need names, so I think it would be safe to change the logic slightly to avoid the potential NPE
> 
> `p.getName().equasIgnoreCase(this.properties.getPrimaryPortName())`

Okay, that would work.
The other problem is the code get exceptions when there are multiple ports and none of them matches with primaryPortName. I spent half an hours to figure it out. It should be better to provide more information in the log.

ryanjbaxter: @zhaozhihom feel free to add some logging and/or documentation around that problem as well

ryanjbaxter:
@zhaozhihom is there a reason why you closed this PR?

ryanjbaxter:
Fixed in a4f8087

PR #747, opened 2021-02-25
Remove clusterrole requirement

### LINKED ISSUE #678: Informer Based Discovery Client Should Default To Current Namespace
Currently the informer based discovery client needs cluster wider permissions.  This is due to a limitation in the Kubernetes Java Client.

See https://github.com/kubernetes-client/java/pull/1409

We will need to update the Kubernetes Java Client version and revert these changes
https://github.com/spring-cloud/spring-cloud-kubernetes/commit/40cef547ced165f4ddc3adfd753b442f29b45b57#diff-4a16634dd21836d945914f0b1c0a5ce44df4246dd23394a19d1801e3c7350b53L12

ryanjbaxter:
Fixes #678

In order to remove the clusterrole requirement from the service account when using the informer based DiscoveryClient we needed to specify the namespace.

We were reading the namespace from the file in the container when running on k8s and adding a PropertySource with `spring.cloud.kubernetes.namespace` set to that value in `AbstractKubernetesProfileEnvironmentPostProcessor`.
https://github.com/spring-cloud/spring-cloud-kubernetes/blob/dc4def977f5ea46f2ce0734fb7977ba25fb686ff/spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/profile/AbstractKubernetesProfileEnvironmentPostProcessor.java#L97

We would then set that property in `KubernetesClientProperties`.

However if we try to use `KubernetesClientProperties` set the namespace in the informers that happens before `AbstractKubernetesProfileEnvironmentPostProcessor` runs.

To solve this I created `KubernetesNamespaceProvider` which has a static variable for the namespace.  I tried to use this as the source of truth for the namespace throughout the code.

As far as getting the informers to use the namespace I had to copy and paste much of the code that creates them from the kubernetes java client and insert a couple of lines that use `KubernetesNamespaceProvider`.


ryanjbaxter:
@yue9944882 would appreciate if you could take a look at this

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/KubernetesNamespaceProvider.java
```diff
@@ -0,0 +1,85 @@
...
+
+	public String getNamespace() {
+		return environment.getProperty(NAMESPACE_PROPERTY, getServiceAccountNamespace());
+	}
+
+	private String getServiceAccountNamespace() {
+		String serviceAccountNamespacePathString = environment.getProperty(NAMESPACE_PATH_PROPERTY,
+				SERVICE_ACCOUNT_NAMESPACE_PATH);
+		if (serviceAccountNamespacePathString == null) {
+			LOG.debug("Looking for service account namespace at: [" + serviceAccountNamespacePathString + "].");
```
spencergibb, line 61: LOG.isDebugEnabled()

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/KubernetesNamespaceProvider.java
```diff
@@ -0,0 +1,85 @@
...
+	/**
+	 * Property name for namespace.
+	 */
+	public static final String NAMESPACE_PROPERTY = "spring.cloud.kubernetes.client.namespace";
+
+	private static final String NAMESPACE_PATH_PROPERTY = "spring.cloud.kubernetes.client.serviceAccountNamespacePath";
+
+	private static final DeferredLog LOG = new DeferredLog();
+
+	private static String serviceAccountNamespace;
```
spencergibb, line 45: Can you explain why this has to be static when, AFAICT, there is no static access to it?

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/KubernetesNamespaceProvider.java
wind57: why though?  this is done internally in the logger?

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/KubernetesNamespaceProvider.java
ryanjbaxter: I want to make sure we are not constantly reading the file if multiple instances are created (even though that should not be the case since it is a managed spring bean)

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/KubernetesNamespaceProvider.java
ryanjbaxter: I guess the one advantage is that you wouldnt construct the logging message unnecessarily

PR #756, opened 2021-03-11
spring-cloud-kubernetes-test-support dependency moved to Testing Dependencies section and scope test added

leaqui:
org.springframework.cloud:spring-cloud-kubernetes-test-support dependency moved to Testing Dependencies section and scope test added

What about com.squareup.okhttp3 dependencies defined in Testing Dependencies section, should they have the test scope too?

### DIFF
--- a/spring-cloud-kubernetes-dependencies/pom.xml
+++ b/spring-cloud-kubernetes-dependencies/pom.xml
@@ -141,12 +141,6 @@
 				<version>${project.version}</version>
 			</dependency>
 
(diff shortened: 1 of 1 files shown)


leaqui:
What about com.squareup.okhttp3 dependencies defined in Testing Dependencies section, should they have test scope too?

ryanjbaxter:
Possibly, I would have to take a deeper look.  You can try and set the scope and if everything compiles it would appear that way.

PR #757, opened 2021-03-12
test scope added to com.squareup.okhttp3 dependencies

leaqui:
This dependencies are defined in Testing Dependencies section


leaqui:
spring-cloud-kubernetes-test-support needs okhttp dependency with compile scope.
So if we add test scope to these dependencies we must replicate the same dependency in other poms, replicating okhttp.version property.

What do you think?

leaqui:
this requires major dependencies refactor in several projects

PR #768, opened 2021-03-24
[WIP] Setting user-agent for ApiClient automatically

yue9944882:
/cc @ryanjbaxter

### DIFF
--- a/spring-cloud-kubernetes-client-autoconfig/src/main/java/org/springframework/cloud/kubernetes/client/KubernetesClientAutoConfiguration.java
+++ b/spring-cloud-kubernetes-client-autoconfig/src/main/java/org/springframework/cloud/kubernetes/client/KubernetesClientAutoConfiguration.java
@@ -46,9 +46,10 @@ public class KubernetesClientAutoConfiguration {
 
 	@Bean
 	@ConditionalOnMissingBean
-	public ApiClient apiClient() throws IOException {
+	public ApiClient apiClient(KubernetesClientProperties kubernetesClientProperties) {
 		ApiClient apiClient = kubernetesApiClient();
 		io.kubernetes.client.openapi.Configuration.setDefaultApiClient(apiClient);
+		apiClient.setUserAgent(kubernetesClientProperties.getUserAgent());
 		return apiClient;
 	}
 
--- a/spring-cloud-kubernetes-client-autoconfig/src/main/java/org/springframework/cloud/kubernetes/client/KubernetesClientInfoContributor.java
+++ b/spring-cloud-kubernetes-client-autoconfig/src/main/java/org/springframework/cloud/kubernetes/client/KubernetesClientInfoContributor.java
@@ -50,7 +50,7 @@ public class KubernetesClientInfoContributor extends AbstractKubernetesInfoContr
 			details.put(HOST_IP, current.getStatus().getHostIP());
 			return details;
 		}
-		return Collections.singletonMap(INSIDE, false);
+		return Collections.singletonMap(INSIDE, "IMPOSSIBLE"); // false -> IMPOSSIBLE
 	}
 
 }
--- a/spring-cloud-kubernetes-client-autoconfig/src/test/java/org/springframework/cloud/kubernetes/client/KubernetesClientInfoContributorTests.java
+++ b/spring-cloud-kubernetes-client-autoconfig/src/test/java/org/springframework/cloud/kubernetes/client/KubernetesClientInfoContributorTests.java
@@ -16,31 +16,8 @@
 
 package org.springframework.cloud.kubernetes.client;
 
-import java.util.Map;
-
-import io.kubernetes.client.openapi.models.V1Pod;
-import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
-import org.mockito.Mock;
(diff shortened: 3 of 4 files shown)


spring-cloud-kubernetes-client-autoconfig/src/main/java/org/springframework/cloud/kubernetes/client/KubernetesClientAutoConfiguration.java
```diff
@@ -46,9 +47,11 @@
 
 	@Bean
 	@ConditionalOnMissingBean
-	public ApiClient apiClient() throws IOException {
```
ryanjbaxter, line 49: I would rather get this value from `KubernetesClientProperties`.  The default should be `application`.  

Why is this user agent needed?

ryanjbaxter (changes requested):

spring-cloud-kubernetes-client-autoconfig/src/main/java/org/springframework/cloud/kubernetes/client/KubernetesClientAutoConfiguration.java
yue9944882: > I would rather get this value from KubernetesClientProperties. The default should be application.

updated

> Why is this user agent needed?

most kubernetes admins are tracking the resource requests by [auditing](https://kubernetes.io/docs/tasks/debug-application-cluster/audit/). and the user-agent header is one of the most useful things that helps us to understanding the source of the requests. otherwise the default user-agent will be [hard-coded](https://github.com/kubernetes-client/java/blob/master/kubernetes/src/main/java/io/kubernetes/client/openapi/ApiClient.java#L106)

ryanjbaxter (changes requested):
Can you please merge in the latest changes from master?  And take a look at the build failures?

spring-cloud-kubernetes-client-autoconfig/src/main/java/org/springframework/cloud/kubernetes/client/KubernetesClientAutoConfiguration.java
```diff
@@ -46,9 +47,11 @@
...
+	public ApiClient apiClient(@Value("${spring.application.name:application}") String applicationName) {
```
yue9944882, line 50: @ryanjbaxter when i moved `@Value` annotation injection to `KubernetesClientProperties` the integration test becomes constantly failing so i reverted to injection via parameters. i can't reproduce that locally b/c all the integration tests are passing perfectly. wdyt?

ryanjbaxter: We need to figure out why they are failing.  We should not need to do this via parameter injection.

yue9944882: is there any way to browse the log from the integration-testing container? by adding some extra printing lines, i found that in the failing tests, the `/actuator/info` endpoint is returning `{"inside":false}` which results the failure --- while actually it's running inside a cluster.

ryanjbaxter: Not 100% sure, maybe just reproduce outside of the integration test in your own sample?

yue9944882: oddly, everytime i ran the integration tests locally, they pass perfectly... let me figure out what's wrong the in the remote test..

ryanjbaxter:
@yue9944882 sorry for taking so long to get back to you on this.  I had to step away from this work to focus on something else.  I know what the problem is.

Basically by requiring the `KubernetesClientProperties` bean in the creation of the `ApiClient` bean it causes `KubernetesClientProperties` to be intialized before `KubernetesClientProfileEnvironmentPostProcessor` runs and therefore the `PropertySource` that adds `spring.cloud.kubernetes.client.namespace` is not there yet.  This results in the `namespace` in `KubernetesClientProperties` to be `null`.  Any API call that requires the namespace is going to fail, so when the `/actuator/health` or `/actuator/info` endpoint is hit in the test, the API call to get the current POD information fails.

The only way I can think of working around this is to actually get the property `spring.cloud.kubernetes.client.user-agent` from the `Environment` directly so we don't instantiate the `KuberentesClientProperties` bean too early.

spring-cloud-kubernetes-client-autoconfig/src/main/java/org/springframework/cloud/kubernetes/client/KubernetesClientAutoConfiguration.java
wind57: @yue9944882 if you do not mind, I will be taking this one where you left and bring it to a PR ready to review state.

wind57:
same PR, with a bit of clean-up and tests [here](https://github.com/spring-cloud/spring-cloud-kubernetes/pull/898)

PR #789, opened 2021-05-16
some polishing in KubernetesClientHealthIndicator

wind57:


wind57:
these same changes were done in Fabric8HealthContributor, this this simply brings things on par with that one

PR #799, opened 2021-05-20
remove deprecated KubernetesServer from tests

wind57:


wind57:
@ryanjbaxter do you think you might be able to take a look at this one too? as it is very close to the other PR that is now merged.

PR #801, opened 2021-05-21
Enable SecretsPropertySource to support properties embedded as key-va…

nightswimmings:
…lue in secrets

Enable parsing of Secrets that represent a property file

### DIFF
--- a/spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/SecretsPropertySource.java
+++ b/spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/SecretsPropertySource.java
@@ -17,6 +17,8 @@
 package org.springframework.cloud.kubernetes.commons.config;
 
 import java.util.Base64;
+import java.util.Collections;
+import java.util.stream.Collectors;
 import java.util.Map;
 
 import org.springframework.core.env.MapPropertySource;
@@ -42,9 +44,18 @@ public class SecretsPropertySource extends MapPropertySource {
 
 	protected static void putAll(Map<String, String> data, Map<String, Object> result) {
 		if (data != null) {
-			data.forEach((k, v) -> result.put(k, new String(Base64.getDecoder().decode(v)).trim()));
+			data.forEach((k, v) -> result.putAll(this.decodeProperty(k,v)));
 		}
 	}
+	
+	protected static Map<String,String> decodeProperty(String key, String value){
+		String decodedValue = new String(Base64.getDecoder().decode(value)).trim();
+		if (key.endsWith(".properties")){
+		   return decodedValue.lines().filter(line -> line.contains("=")).collect(Collectors.toMap(line -> line.split("=")[0], line -> line.split("=",2)[1]));
+		} else {
+		   return Collections.singletonMap(key,decodedValue);
+		}	
+	}	
 
 	@Override
 	public String toString() {


wind57:
why can't you do this in your business code, though? I doubt this will be approved to be honest.

As an addendum, I had to read helm chart secrets recently and I have used sppring-cloud-kubernetes also. Not only are those secrets Base64 encoded _twice_, they are also gzipped and have embedded properties. I did all the needed parsing in my code... I guess you can do the same exact thing.

spencergibb:
Let's see if there's more community interest.

nightswimmings:
@wind57 I can do it in my business code by interpreting the property, but I found it quite infrastructural to do it myself. Besides, through spring cloud, it fits nicely into properties configuration and sources resolution framework in spring boot.

ryanjbaxter (changes requested):
I think we would like to move forward with this PR.

@nightswimmings can you add some tests?

Zetten:
One aspect I think not covered here is when Secrets are mounted as volumes in the Pod.

It seems this side of things is handled directly in the [SecretsPropertySourceLocator](https://github.com/spring-cloud/spring-cloud-kubernetes/blob/main/spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/SecretsPropertySourceLocator.java#L91), which collects each file as key:value :: filename:contents (i.e. you would have a single property `application.properties` with value `<all the properties concatenated>`). 

The [equivalent for ConfigMaps](https://github.com/spring-cloud/spring-cloud-kubernetes/blob/main/spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/ConfigMapPropertySourceLocator.java#L90) does handle the decoding of single-key properties, so perhaps this logic could be extracted to PropertySourceUtils and shared.

DaviGia:
Hi, any updates on this? I encouter the same limitation. For context, I deploy my microservices through helm and I've even tried to specify the secrets keys using [the same format](https://github.com/spring-projects/spring-boot/wiki/Relaxed-Binding-2.0#lists-1) supported by the environment variables, but it's not working. Perhaps it may be included as alternative to single encoded file?

I currently have implemented some helper functions in helm to flatten a map into a list of properties (env vars format compatible) to simplify the chart configuration. This approach works on other frameworks like Steeltoe (the format is slightly different, but the solution is equivalent).

An example of secret definition would be the following:

``` yaml
# chart values
serviceConfig:
  spring:
    security:
      user:
        password: test
    redis:
      password: test
test:
  array:
     basic:
     - test
     - test
     complex:
     - key: test
       value: test

# rendered template
apiVersion: v1
kind: Secret
metadata:
  name: test-service
  namespace: chart-testing
  labels:
    app.kubernetes.io/instance: test
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: test-service
    app.kubernetes.io/version: 1.13.0
    helm.sh/chart: test-service-2.4.0
data:
  spring_security_user_password: dGVzdA==
  spring_redis_password: dGVzdA==
  test_array_basic_0: dGVzdA==
  test_array_basic_1: dGVzdA==
  test_array_complex_0_key: dGVzdA==
  test_array_complex_0_value: dGVzdA==
```

wind57:
@DaviGia this is definitely on my radar, but there are many other pending issues that must be taken care of.

wind57:
@ryanjbaxter this can be closed, as secrets now do support single yaml/properties files, just like config maps.

PR #803, opened 2021-05-24
fabric8 bootstrap configuration changes

wind57:


spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/ConditionalOnKubernetesConfigEnabled.java
```diff
@@ -0,0 +1,40 @@
...
+import java.lang.annotation.Target;
+
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+
+/**
+ * Provides a more succinct conditional <code>spring.cloud.kubernetes.config.enabled</code>.
+ *
+ * @author wind57
+ */
+
```
wind57, line 33: imho, this and the one above slightly improve reading the code. they could also be used in kubernetes-client

spring-cloud-kubernetes-fabric8-config/src/main/java/org/springframework/cloud/kubernetes/fabric8/config/Fabric8BootstrapConfiguration.java
```diff
@@ -39,31 +40,24 @@
...
-	@Configuration(proxyBeanMethods = false)
-	@Import({ KubernetesCommonsAutoConfiguration.class, Fabric8AutoConfiguration.class })
-	protected static class KubernetesPropertySourceConfiguration {
-
-		@Autowired
-		private KubernetesClient client;
-
-		@Bean
-		@ConditionalOnProperty(name = "spring.cloud.kubernetes.config.enabled", matchIfMissing = true)
-		public Fabric8ConfigMapPropertySourceLocator configMapPropertySourceLocator(
```
wind57, line 56: remove the inner class. initially I though that it is needed because of the `@Import({ KubernetesCommonsAutoConfiguration.class, Fabric8AutoConfiguration.class })`. So when `spring.cloud.kubernetes.enabled=false` these would not be imported but:

 - `KubernetesCommonsAutoConfiguration` and `Fabric8AutoConfiguration` are annotated with  `@ConditionalOnKubernetesEnabled`, so that is not a problem
 - even if I add a dummy config (without any annotations), import it, it will still not be picked up (the constructor inside it is not even called) when `spring.cloud.kubernetes.enabled=false`.

As such, this looks like a safe change to me. I hope it makes sense

spring-cloud-kubernetes-fabric8-config/src/test/java/org/springframework/cloud/kubernetes/fabric8/config/Fabric8BootstrapConfigurationTests.java
```diff
@@ -0,0 +1,128 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.context.ConfigurableApplicationContext;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * @author wind57
+ */
+public class Fabric8BootstrapConfigurationTests {
```
wind57, line 31: some tests that prove everything is still correct

spring-cloud-kubernetes-fabric8-config/src/test/java/org/springframework/cloud/kubernetes/fabric8/config/Fabric8ConfigUtilsTests.java
```diff
@@ -16,29 +16,29 @@
 
 package org.springframework.cloud.kubernetes.fabric8.config;
 
-import io.fabric8.kubernetes.client.server.mock.KubernetesServer;
-import org.junit.ClassRule;
-import org.junit.Test;
```
wind57, line 21: unfortunately, it is me to blame for this change. I missed it in one of the previous PRs.

Issue #804, opened 2021-05-27
Fails to change leadership when leader pod in configMap doesn't exist


jaredstehler:
In the case where the leadership configMap references a non-existent pod (perhaps in the case of a cluster rolling upgrade?), the leadership controller gets into a state where it cannot update leader to a valid active pod.

This is due to the underlying fabric8 library's `Readiness::isReady()` throwing IllegalArgumentException when the queried item is null / missing. See: https://github.com/fabric8io/kubernetes-client/issues/2672

Since this is resolved in fabric8 5.1+ and there seems to be a question of even continuing on with fabric8, it would seem that the simplest solution here would be to wrap `Fabric8LeadershipController::isPodReady()` with a try/catch and return false on IllegalArgument? 

https://github.com/spring-cloud/spring-cloud-kubernetes/blob/main/spring-cloud-kubernetes-fabric8-leader/src/main/java/org/springframework/cloud/kubernetes/fabric8/leader/Fabric8LeadershipController.java#L131

jaredstehler:
project has since moved fabric8 dep to 5.5.

PR #805, opened 2021-05-27
Handle case where given entity to check readiness for is null/missing

jaredstehler:
resolves https://github.com/spring-cloud/spring-cloud-kubernetes/issues/804


spencergibb:
can you add a test?

steven-sheehy:
@jaredstehler Any update? Do you need any help on this? We're impacted by this as well so I can take this over if you don't have bandwidth to add those tests.

jaredstehler:
let me take a crack at it this evening; i'll punt it over to you if I can't get something working.

jaredstehler:
added a test

jaredstehler:
@spencergibb this should be good to go now?

steven-sheehy:
Is this still needed since fabric8 kubernetes-client 5.5.0 is now integrated with the fix to not throw IllegalArgumentException?

PR #815, opened 2021-06-16
refactor PropertySourceUtils

wind57:


spring-cloud-kubernetes-commons/src/test/java/org/springframework/cloud/kubernetes/commons/config/PropertySourceUtilsTest.java
```diff
@@ -90,4 +91,41 @@ void yamlParserGenerator_springConfigActivateOnProfile_mismatchProfile() {
 		assertThat(properties.getProperty("spring.config.activate.on-profile")).isNull();
 	}
```
wind57, line 93: added some tests

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/PropertySourceUtils.java
```diff
@@ -98,13 +102,14 @@ else if (properties.containsKey(SPRING_PROFILES)) {
 	}
 
 	/**
-	 * Throws IllegalStateException.
```
wind57, line 101: - slightly change the documentation of the method
 - rename `u, v` to `left, right`
 - drop `String.format`, use `+` instead. for a single concat `String::format` is an overkill (plus it is a lot more expensive)

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/PropertySourceUtils.java
```diff
@@ -80,12 +80,16 @@ private PropertySourceUtils() {
 			yamlFactory.setDocumentMatchers(properties -> {
 				if (environment != null) {
 					String profiles = null;
-					if (properties.containsKey(SPRING_CONFIG_ACTIVATE_ON_PROFILE)) {
```
wind57, line 83: this is a common anti-pattern: `contains`, followed by `get`; which performs _two_ hash look-ups. Instead, issue a `get` followed by a check against `null`.

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/PropertySourceUtils.java
```diff
@@ -54,20 +59,15 @@
...
 	};
 
 	/**
 	 * Function to convert Properties to a Map.
 	 */
 	public static final Function<Properties, Map<String, Object>> PROPERTIES_TO_MAP = p -> p.entrySet().stream()
-			.collect(Collectors.toMap(e -> String.valueOf(e.getKey()), Map.Entry::getValue, throwingMerger(),
-					java.util.LinkedHashMap::new));
-
-	private PropertySourceUtils() {
```
wind57, line 68: not deleted, but moved at the top. much easier to read the code, especially for a utility class. at least, imo.

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/PropertySourceUtils.java
```diff
@@ -54,20 +59,15 @@
 			return properties;
 		}
 		catch (IOException e) {
-			throw new IllegalArgumentException();
```
wind57, line 57: should be obvious that a `UncheckedIOException(e)` will provide a far better insight of what happened vs a `IllegalArgumentException()`.

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/PropertySourceUtils.java
```diff
@@ -54,20 +59,15 @@
...
+			throw new UncheckedIOException(e);
 		}
 	};
 
 	/**
 	 * Function to convert Properties to a Map.
 	 */
 	public static final Function<Properties, Map<String, Object>> PROPERTIES_TO_MAP = p -> p.entrySet().stream()
-			.collect(Collectors.toMap(e -> String.valueOf(e.getKey()), Map.Entry::getValue, throwingMerger(),
```
wind57, line 65: - I really dislike usages of `String::valueOf` for these cases, for the simple fact that internally (it is even documented) it will do a mapping of a `null` to a `"null"`. I would much rather prefer the code to fail at this point with a NPE, instead of this transformation. Instead I propose to use `e.getKey().toString()`
- then providing the `throwingMerger()` is not really needed, but, without such a merger, you can not specify a custom `Supplier` like `LinkedHashMap::new`. I perfectly understand that. If you take a step back, and think that the source of the stream is a `Properties` (a `Map`), that has _undefined order_, collecting this stream to a `LinkedHashMap` (that defines an order), makes no sense. As such, remove both the merger and the Supplier.

ryanjbaxter (approved):

PR #821, opened 2021-06-29
Remove unnecessary API groups in sample Role

yaraskm:
- The resource requirements listed by component are all members of the
  core API group. There is no need to include the `extensions` and
  `apps` groups. This Role will currently try to grant permissions on
  types that do not exist, such a `pod.extensions` and `pod.apps`.

- If you are a member of a multitenant cluster, there's a good chance
  that you are running with either the `admin` or `edit` built-in
  Clusterroles scoped down to your namespace(s). Since certain types
  such as `pod.extensions` do not exist, your binding to a built-in Role
  will not contain permissions on these types. As such, you will get a
  permission error when trying to create the Role (trying to grant
  permissions you do not hold). The Role as written could only currently
  be created in a cluster by someone who has a higher permission level,
  such as an admin.

### DIFF
--- a/docs/src/main/asciidoc/security-service-accounts.adoc
+++ b/docs/src/main/asciidoc/security-service-accounts.adoc
@@ -55,7 +55,7 @@ metadata:
   namespace: YOUR-NAME-SPACE
   name: namespace-reader
 rules:
-  - apiGroups: ["", "extensions", "apps"]
+  - apiGroups: [""]
     resources: ["configmaps", "pods", "services", "endpoints", "secrets"]
     verbs: ["get", "list", "watch"]
 


ryanjbaxter:
So to be clear...the problem will only happen in a multi-cluster environment when the admin role is scoped to namespaces?

yaraskm:
It's more that this will occur whenever a user is trying to create this role when they themselves aren't an admin on the cluster scope, which is usually the case when they're running in a multi-tenant setup. Kubernetes RBAC enforces that in order to grant permissions, you must already be permitted to use those same permissions. To get around this, the `cluster-admin` role uses wildcards like the following:

```bash
$ kubectl get clusterrole cluster-admin -o yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: cluster-admin
  ...
rules:
- apiGroups:
  - '*'
  resources:
  - '*'
  verbs:
  - '*'
- nonResourceURLs:
  - '*'
  verbs:
  - '*'
```

This allows anyone running with the `cluster-admin` role to create the sample `namespace-reader` as it was in the docs. For pretty much every other role though, the allowed `apiGroups`, `resources`, and `verbs` are specified explicitly (no wildcards). For any non-`cluster-admin`, in order to create the `namespace-reader` role as written they would need to already have explicit permissions on the non-existent `pod.extensions` and `pod.apps` types, which is highly unlikely. 

In short, only Cluster Admins can currently create the role as written in the docs. Removing the non-existent types would allow anyone with a reasonable set of permissions on their namespace(s) to create the role.

ryanjbaxter:
That makes sense, thanks!

PR #834, opened 2021-07-15
some refactoring in config maps functionality 

wind57:


spring-cloud-kubernetes-fabric8-config/src/main/java/org/springframework/cloud/kubernetes/fabric8/config/Fabric8ConfigUtils.java
```diff
@@ -45,4 +45,8 @@ public static String getApplicationNamespace(KubernetesClient client, String con
 		return configNamespace;
 	}
 
+	static String getNamespace(KubernetesClient client, String namespace) {
```
wind57, line 48: - has moved from `Fabric8ConfigMapPropertySource` in here. Seems to me a more appropriate place to be. Both classes are in the same package, and to preserve the `private` modifier that was present before, we keep it here without any modifiers at all. So this is still "hidden" from the external world, just that now it is scoped at package level. 
- I have also removed the deprecated `StringUtils::isEmpty` with the inverted `StringUtils::hasLength`

spring-cloud-kubernetes-fabric8-config/src/main/java/org/springframework/cloud/kubernetes/fabric8/config/Fabric8ConfigMapPropertySource.java
```diff
@@ -41,25 +44,21 @@
...
 	}
 
-	public Fabric8ConfigMapPropertySource(KubernetesClient client, String name, String namespace,
+	public Fabric8ConfigMapPropertySource(KubernetesClient client, String applicationName, String namespace,
 			Environment environment) {
-		super(getName(name, getNamespace(client, namespace)),
-				asObjectMap(getData(client, name, getNamespace(client, namespace), environment)));
-	}
-
-	private static String getNamespace(KubernetesClient client, String namespace) {
```
wind57, line 53: was not really removed, but moved to `Fabric8ConfigUtils`. Might come handy in future other places too; besides this surely looks like a "utility" method, so why not place it in the proper place?

spring-cloud-kubernetes-fabric8-config/src/main/java/org/springframework/cloud/kubernetes/fabric8/config/Fabric8ConfigMapPropertySourceLocator.java
```diff
@@ -45,11 +45,11 @@ public Fabric8ConfigMapPropertySourceLocator(KubernetesClient client, ConfigMapC
 	}
 
 	@Override
```
wind57, line 47: - instead of `String name` use `String applicationName`; it's a minor change, but reading the code becomes a lot more easy (from personal experience at least)
- I've extracted this:

```
String namespaceName = getApplicationNamespace(this.client, normalizedSource.getNamespace(),
				configurationTarget);
```

to a separate call. Again, for the reading purposes only.

spring-cloud-kubernetes-fabric8-config/src/main/java/org/springframework/cloud/kubernetes/fabric8/config/Fabric8ConfigMapPropertySource.java
```diff
@@ -85,10 +84,11 @@ private static String getNamespace(KubernetesClient client, String namespace) {
 
 		}
 		catch (Exception e) {
-			LOG.warn("Can't read configMap with name: [" + name + "] in namespace:[" + namespace + "]. Ignoring.", e);
+			LOG.warn("Can't read configMap with name: [" + applicationName + "] in namespace:[" + namespace
+					+ "]. Ignoring.", e);
 		}
 
-		return new LinkedHashMap<>();
+		return Collections.emptyMap();
```
wind57, line 91: I guess an obvious change.

spring-cloud-kubernetes-fabric8-config/src/main/java/org/springframework/cloud/kubernetes/fabric8/config/Fabric8ConfigMapPropertySource.java
```diff
@@ -85,10 +84,11 @@ private static String getNamespace(KubernetesClient client, String namespace) {
 
 		}
 		catch (Exception e) {
-			LOG.warn("Can't read configMap with name: [" + name + "] in namespace:[" + namespace + "]. Ignoring.", e);
```
wind57, line 88: `applicationName` instead of `name`

spring-cloud-kubernetes-fabric8-config/src/main/java/org/springframework/cloud/kubernetes/fabric8/config/Fabric8ConfigMapPropertySource.java
```diff
@@ -68,9 +67,9 @@ private static String getNamespace(KubernetesClient client, String namespace) {
 			if (environment != null) {
 				for (String activeProfile : environment.getActiveProfiles()) {
 
-					String mapNameWithProfile = name + "-" + activeProfile;
+					String mapNameWithProfile = applicationName + "-" + activeProfile;
 
-					ConfigMap mapWithProfile = StringUtils.isEmpty(namespace)
```
wind57, line 73: replace `String::isEmpty` with an inverted `String::hasLength`

spring-cloud-kubernetes-fabric8-config/src/main/java/org/springframework/cloud/kubernetes/fabric8/config/Fabric8ConfigMapPropertySource.java
```diff
@@ -41,25 +44,21 @@
...
 	}
 
-	private static Map<String, Object> getData(KubernetesClient client, String name, String namespace,
+	private static Map<String, Object> getData(KubernetesClient client, String applicationName, String namespace,
 			Environment environment) {
 		try {
-			Map<String, Object> result = new LinkedHashMap<>();
-			ConfigMap map = StringUtils.isEmpty(namespace) ? client.configMaps().withName(name).get()
-					: client.configMaps().inNamespace(namespace).withName(name).get();
+			Map<String, Object> result = new HashMap<>();
```
wind57, line 59: I see no reason to have an order for the resulting `Map`, so replace it with `HashMap`. There is no significant change on this, but it creates a mental pressure when reading the code : "if it is a `LinkedHashMap`, means we want to preserve the order", so you are kind of stuck understanding why that order needs to be preserved; when the reality is simply not so.

spring-cloud-kubernetes-fabric8-config/src/main/java/org/springframework/cloud/kubernetes/fabric8/config/Fabric8ConfigMapPropertySource.java
```diff
@@ -41,25 +44,21 @@
...
+			ConfigMap map = !StringUtils.hasLength(namespace) ? client.configMaps().withName(applicationName).get()
```
wind57, line 60: same `StringUtils::hasLength` change

spring-cloud-kubernetes-fabric8-config/src/main/java/org/springframework/cloud/kubernetes/fabric8/config/Fabric8ConfigMapPropertySource.java
```diff
@@ -41,25 +44,21 @@
...
+	public Fabric8ConfigMapPropertySource(KubernetesClient client, String applicationName, String namespace,
 			Environment environment) {
-		super(getName(name, getNamespace(client, namespace)),
-				asObjectMap(getData(client, name, getNamespace(client, namespace), environment)));
-	}
-
-	private static String getNamespace(KubernetesClient client, String namespace) {
-		return StringUtils.isEmpty(namespace) ? client.getNamespace() : namespace;
+		super(getName(applicationName, getNamespace(client, namespace)),
+				getData(client, applicationName, getNamespace(client, namespace), environment));
```
wind57, line 53: careful here, as this removes one call : `asObjectMap`. If you look what `asObjectMap` is trying to achieve, it is simply converting a `Map` to a `LinkedHashMap`. Since such a change is not needed (surely looks like a initial artifact of an implementation), this call can be elided away.

spring-cloud-kubernetes-client-config/src/main/java/org/springframework/cloud/kubernetes/client/config/KubernetesClientConfigMapPropertySource.java
```diff
@@ -62,7 +63,7 @@ public KubernetesClientConfigMapPropertySource(CoreV1Api coreV1Api, String name,
 		catch (ApiException e) {
 			LOG.warn("Unable to get ConfigMap " + name + " in namespace " + namespace, e);
 		}
-		return new LinkedHashMap<>();
+		return Collections.emptyMap();
```
wind57, line 66: obvious change, imho

spring-cloud-kubernetes-client-config/src/main/java/org/springframework/cloud/kubernetes/client/config/KubernetesClientConfigMapPropertySource.java
```diff
@@ -38,7 +39,7 @@
 
 	public KubernetesClientConfigMapPropertySource(CoreV1Api coreV1Api, String name, String namespace,
 			Environment environment) {
-		super(getName(name, namespace), asObjectMap(getData(coreV1Api, name, namespace, environment)));
```
wind57, line 41: I've removed the `asObjectMap` call. It's implementation looked like this:

```
protected static Map<String, Object> asObjectMap(Map<String, Object> source) {
		return source.entrySet().stream()
				.collect(Collectors.toMap(Entry::getKey, Entry::getValue, throwingMerger(), LinkedHashMap::new));
}
```

It took a `Map` and created a `LinkedHashMap` out of it. So you take a `Map` interface, without any defined order and transform it to a `LinkedHashMap`? This is most probably an artifact of some previous implementation, in the current context - it would not make sense. 

The fact that it is `protected` also means - we can drop it; unless you declare the same package where it was declared initially, which no one should have done that anyway. imho, removing it - is safe.

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/ConfigMapPropertySource.java
```diff
@@ -45,14 +50,6 @@
 
 	private static final Log LOG = LogFactory.getLog(ConfigMapPropertySource.class);
 
-	protected static final String APPLICATION_YML = "application.yml";
```
wind57, line 48: moved these properties to `Constants`.

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/ConfigMapPropertySource.java
```diff
@@ -64,37 +61,26 @@ protected static Environment createEnvironmentWithActiveProfiles(String[] active
 	}
 
 	protected static String getName(String name, String namespace) {
-		return new StringBuilder().append(PREFIX).append(Constants.PROPERTY_SOURCE_NAME_SEPARATOR).append(name)
-				.append(Constants.PROPERTY_SOURCE_NAME_SEPARATOR).append(namespace).toString();
+		return PREFIX + PROPERTY_SOURCE_NAME_SEPARATOR + name + PROPERTY_SOURCE_NAME_SEPARATOR + namespace;
```
wind57, line 64: a slight change to drop `StringBuilder`, the compiled code would look exactly the same anyway.

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/ConfigMapPropertySource.java
```diff
@@ -64,37 +61,26 @@ protected static Environment createEnvironmentWithActiveProfiles(String[] active
...
 	}
 
 	protected static Map<String, Object> processAllEntries(Map<String, String> input, Environment environment) {
 
-		Set<Entry<String, String>> entrySet = input.entrySet();
+		Set<Map.Entry<String, String>> entrySet = input.entrySet();
 		if (entrySet.size() == 1) {
 			// we handle the case where the configmap contains a single "file"
-			// in this case we don't care what the name of t he file is
```
wind57, line 76: nitpick `the`, as a single word.

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/ConfigMapPropertySource.java
```diff
@@ -64,37 +61,26 @@ protected static Environment createEnvironmentWithActiveProfiles(String[] active
...
-			Entry<String, String> singleEntry = entrySet.iterator().next();
+			// in this case we don't care what the name of the file is
+			Map.Entry<String, String> singleEntry = entrySet.iterator().next();
 			String propertyName = singleEntry.getKey();
 			String propertyValue = singleEntry.getValue();
 			if (propertyName.endsWith(".yml") || propertyName.endsWith(".yaml")) {
-				if (LOG.isDebugEnabled()) {
```
wind57, line 81: remove `isDebugEnabled`. I understand why it was done here - so that the message that is supposed to do in debug mode, is not computed unless debug is on. But the price to pay for a branch is usually much higher than an allocation. 

Overall, I blame the library itself, for not providing a `Supplier<String>` as an argument to `LOG::debug`...

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/ConfigMapPropertySource.java
```diff
@@ -64,37 +61,26 @@ protected static Environment createEnvironmentWithActiveProfiles(String[] active
...
 			else if (propertyName.endsWith(".properties")) {
-				if (LOG.isDebugEnabled()) {
-					LOG.debug("The single property with name: [" + propertyName
-							+ "] will be treated as a properties file");
-				}
-
+				LOG.debug("The single property with name: [" + propertyName + "] will be treated as a properties file");
 				return KEY_VALUE_TO_PROPERTIES.andThen(PROPERTIES_TO_MAP).apply(propertyValue);
 			}
-			else {
```
wind57, line 95: this `else` branch is duplicated by the final `return` statement - it can be removed.

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/ConfigMapPropertySource.java
```diff
@@ -104,7 +90,7 @@ else if (propertyName.endsWith(".properties")) {
 
 		return input.entrySet().stream().map(e -> extractProperties(e.getKey(), e.getValue(), environment))
 				.filter(m -> !m.isEmpty()).flatMap(m -> m.entrySet().stream())
-				.collect(Collectors.toMap(Entry::getKey, Entry::getValue, throwingMerger(), LinkedHashMap::new));
+				.collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue, throwingMerger(), HashMap::new));
```
wind57, line 93: although, I did some changes in this PR and some others where I removed `throwingMerger(), LinkedHashMap::new`, I think that these are justified here. I prefer a failure, rather then an override, from diff properties of `ConfigMap`.

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/ConfigMapPropertySource.java
```diff
@@ -117,16 +103,7 @@ else if (resourceName.equals(APPLICATION_PROPERTIES)) {
...
-			{
-				put(resourceName, content);
-			}
-		};
-	}
-
-	protected static Map<String, Object> asObjectMap(Map<String, Object> source) {
-		return source.entrySet().stream()
-				.collect(Collectors.toMap(Entry::getKey, Entry::getValue, throwingMerger(), LinkedHashMap::new));
+		return Collections.singletonMap(resourceName, content);
```
wind57, line 106: replace double brace initialization with `Collections::singletonMap`

spring-cloud-kubernetes-fabric8-config/src/main/java/org/springframework/cloud/kubernetes/fabric8/config/Fabric8ConfigMapPropertySource.java
```diff
@@ -41,25 +44,21 @@
 	private static final Log LOG = LogFactory.getLog(Fabric8ConfigMapPropertySource.class);
 
 	public Fabric8ConfigMapPropertySource(KubernetesClient client, String name) {
-		this(client, name, null, (Environment) null);
```
wind57, line 44: cast not needed

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/ConfigMapPropertySourceLocator.java
```diff
@@ -57,17 +57,17 @@ public ConfigMapPropertySourceLocator(ConfigMapConfigProperties properties) {
...
 			String configurationTarget, ConfigurableEnvironment environment);
 
 	@Override
 	public PropertySource<?> locate(Environment environment) {
 		if (environment instanceof ConfigurableEnvironment) {
 			ConfigurableEnvironment env = (ConfigurableEnvironment) environment;
 
-			List<NormalizedSource> sources = this.properties.determineSources();
 			CompositePropertySource composite = new CompositePropertySource("composite-configmap");
 			if (this.properties.isEnableApi()) {
```
wind57, line 69: call `determineSources` only when this is needed

ryanjbaxter (approved):

wind57:
thank you for re-running this!

PR #844, opened 2021-08-04
Fix duplicate config maps

wind57:


spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/ConfigMapPropertySourceLocator.java
```diff
@@ -67,7 +68,7 @@ protected abstract MapPropertySource getMapPropertySource(String applicationName
 
 			CompositePropertySource composite = new CompositePropertySource("composite-configmap");
 			if (this.properties.isEnableApi()) {
-				List<NormalizedSource> sources = this.properties.determineSources();
```
wind57, line 70: I am proposing a minor change here. If users have configured (even by accident), a duplicate config map, we should not be reading it twice and create two PropertySources out of that. Ideally, we should be doing that in `ConfigMapConfigProperties`, where we could change `paths` and `sources` to be `Set`s, not `List`s. Of course, that is a breaking change, so the next best thing is to get a `Set` here and thus unique sources.

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/ConfigMapPropertySourceLocator.java
```diff
@@ -88,7 +89,8 @@ private MapPropertySource getMapPropertySourceForSingleConfigMap(ConfigurableEnv
 	}
 
 	private void addPropertySourcesFromPaths(Environment environment, CompositePropertySource composite) {
-		properties.getPaths().stream().map(Paths::get).filter(p -> {
```
wind57, line 91: and unique here. 

addendum: later in the code, we do `collect(Collectors.toList())`, where the result is a `List<Path>`. imo, it's not safe to do `Collectors.toSet()` there, because `map(Paths::get)`, specifically `Paths::get` does not guarantee any specific implementation of `Path`, which might not override hashCode/equals. In practice, it will return a `UnixPath` that _does_ override those methods, but I would stay on the safer side here.

spring-cloud-kubernetes-fabric8-config/src/test/java/org/springframework/cloud/kubernetes/fabric8/config/MultipleConfigMapsTests.java
```diff
@@ -85,11 +86,10 @@ private static void createConfigmap(KubernetesClient client, String configMapNam
 				.addToData(data).done();
 	}
 
-	// the last confimap defined in 'multiplecms.yml' has the highest priority, so
```
wind57, line 88: I do not agree with this comment, but most probably at time of the writing it made sense. Overall, I do not think that the order in which you define config maps in bootstrap files should matter. Especially, enforcing this in a unit tests is somehow weird.

wind57:
@ryanjbaxter ready to review. thank you.

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/ConfigMapPropertySourceLocator.java
ryanjbaxter: Could this not end up messing with the order of the property sources?  That is important

spring-cloud-kubernetes-fabric8-config/src/test/java/org/springframework/cloud/kubernetes/fabric8/config/MultipleConfigMapsTests.java
ryanjbaxter: Not sure I agree, order matters a lot when it comes to property sources, can you explain why you dont think it should matter?

ryanjbaxter (changes requested):

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/ConfigMapPropertySourceLocator.java
wind57: hmm, then I am a bit confused, may be I miss something. Let's suppose this is my config :

```
spring:
  cloud:
    kubernetes:
      config:
        namespace: spring-k8s
        sources:
          - name: config-map-c
          - name: config-map-a
          - name: config-map-a
```

This is the course of actions that we will do:

```
CompositePropertySource composite = new CompositePropertySource("composite-configmap");

 - MapPropertySource source = read "config-map-c"; composite.addFirstPropertySource(source);
 - MapPropertySource source = read "config-map-a"; composite.addFirstPropertySource(source);
 - MapPropertySource source = read "config-map-a"; composite.addFirstPropertySource(source);
```

With the change that I am proposing, it will be : 


```
 - MapPropertySource source = read "config-map-a"; composite.addFirstPropertySource(source);
 - MapPropertySource source = read "config-map-c"; composite.addFirstPropertySource(source);
```

Notice that reading happens only for two of them and they are read in reverse (I did this on purpose here to emphasize my point ).

---

If I get your point correctly, you are saying that if users specify : 

```
- name: config-map-c
- name: config-map-a
```

vs 

```
- name: config-map-a
- name: config-map-c
```

this somehow matters?

ryanjbaxter: Yes is does, because the order of the property sources determines which one takes priority when there are conflicting properties.  So if config map a and c have the same property defined if config map c comes last it will override the property from config map a.

wind57: this is _exactly_ what I thought. You make sense now, thank you.

Issue #846, opened 2021-08-06
Upgrade to okhttp 5.x


ovolynets:
**Is your feature request related to a problem? Please describe.**
The currently used version of [okhttp](https://github.com/square/okhttp) (version `3.14`) has been discovered [CVE-2021-0341](https://nvd.nist.gov/vuln/detail/CVE-2021-0341), here it is used:

https://github.com/spring-cloud/spring-cloud-kubernetes/blob/1f17ff5b505c293dcf6611214779f1e5315752b6/spring-cloud-kubernetes-dependencies/pom.xml#L42

Also Sonatype IQ Server reports this transient dependency while scanning the project for CVEs.

**Describe the solution you'd like**
I suggest to consider an upgrade to a more recent release of okhttp, latest being `5.0.0-alpha.2`. Earlier releases such as 4.9.1 currently do not have the fix included yet, unfortunately.
I am not sure without looking at the code that uses okhttp whether this vulnerability can be exploited at all here - I'd ask the develop team to check for this. Thank you.

**Additional context**
[Related issue](https://github.com/square/okhttp/issues/6724) reported in okhttp

wind57:
Thank you for reporting this. I have a few bullet points here:

- using `alpha` does not sound like a very safe thing to do, imo. 

- we only use that in `spring-cloud-kubernetes-integration-tests` and `spring-cloud-kubernetes-examples` directly; so from this point of view _only_, this could be a safe-ish change.

- the problem is that this is used internally in the `kubernetes-client`, which means we will override the version that they use. Currently we are on `11.0.2` version of `kubernetes-java-client`, but even their latest version [is still not on 5.0.x](https://github.com/kubernetes-client/java/blob/master/pom.xml#L54). This point, invalidates the previous one; making this a non-safe upgrade. 

To me, the proper solution is to open this same defect in [kubernetes-client](https://github.com/kubernetes-client/java) and see what the leads think about it over there, because from our side directly, it looks like a fairly safe change (since the direct dependency is part of a test project and example, which are not used in production). 

Just my 0.02$

ovolynets:
Dear erabii,

Thanks for your 0.02$ - gives a bit more context. My comments inlined:

using alpha does not sound like a very safe thing to do, imo.
>

I agree, but right now it's the only published version of okhttp.

we only use that in spring-cloud-kubernetes-integration-tests and
> spring-cloud-kubernetes-examples directly; so from this point of view
[...]

PR #849, opened 2021-08-09
fix ok http client explicit dependency

wind57:


wind57:
The changes are small, but it will take a significant amount of words explaining it, so please bear with me.

We need to first read carefully [this part of maven documentation](https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#importing-dependencies) (the second example with Project `X`, `Y` and `Z` and what `a` dependency ends up being resolved). After we transpose that example to our code base (removed all irrelevant parts), we have this in `spring-cloud-kubernetes-dependencies`:


```
....
<dependencyManagement>
    <dependencies>
        ....
        <dependency>
		<artifactId>client-java</artifactId>
		<version>${kubernetes-java-client.version}</version>
	</dependency>

         .....

        <dependency>
		<groupId>com.squareup.okhttp3</groupId>
		<artifactId>okhttp</artifactId>
		<version>${okhttp.version}</version>
	</dependency>
	<dependency>
		<groupId>com.squareup.okhttp3</groupId>
		<artifactId>logging-interceptor</artifactId>
		<version>${okhttp.version}</version>
	</dependency>

         .....

```

So `io.kubernetes:client-java:11.0.2` will have a transitive dependency : `com.squareup.okhttp3:okhttp:3.14.9` (https://repo1.maven.org/maven2/io/kubernetes/client-java-parent/11.0.2/client-java-parent-11.0.2.pom); but we try to bring an explicit one with version: `3.14.4`. (I am showing against `2.0.x` branch btw, but things have the same relevance in `main` too). 

Because the order in which these dependencies are written matters (that maven documentation might not be super explicit about it), it means that currently the explicit version that we try to use, is simply ignored. This means we can drop it. 

The downside is that if k8s-client ever drops okhttp client, some of our tests will fail, because the only place where we use okhttp is in tests or examples:

```
cd spring-cloud-kubernetes
 grep -r "com.squareup.okhttp3" --exclude .flattened-pom.xml . | sort --unique  
./spring-cloud-kubernetes-examples/kubernetes-loadbalancer-example/greeting-service/pom.xml:                    <groupId>com.squareup.okhttp3</groupId>
./spring-cloud-kubernetes-integration-tests/discovery/tests/pom.xml:                    <groupId>com.squareup.okhttp3</groupId>
./spring-cloud-kubernetes-integration-tests/pom.xml:                            <groupId>com.squareup.okhttp3</groupId>
./spring-cloud-kubernetes-integration-tests/simple-configmap/pom.xml:                   <groupId>com.squareup.okhttp3</groupId>
./spring-cloud-kubernetes-integration-tests/simple-core/pom.xml:                        <groupId>com.squareup.okhttp3</groupId>
```

wind57:
@ryanjbaxter leaving a note here also - as this is ready for review too. It should address [this](https://github.com/spring-cloud/spring-cloud-kubernetes/issues/846)

ryanjbaxter:
Thanks for the detailed explaination!

So the reason why everything (the tests and examples) still compiles is because of the transitive dependency from `kubernetes-client`?

wind57:
sort of. currently, even if `kubernetes-client` still drops ok http, our code will still compile and work, but only because we have that dependency in `dependencyManagement` and require it explicitly in those projects.

At the same time, if we drop it, it will _still_ work (as this PR proves), _because_ of the transitive dependency. The problem is that we override that transitive dependency version. To try and simplify this : `dependencyManagement` will tell what version of transitive dependency to use. 

So, if we drop it like in this PR : we rely on the transitive one.
At the same time, if we do not want to rely on that transitive one, we should not manage ok-http via bom; because in that case it simply overrides the dependency version that kubernetes client really requires. It's like a vicious circle.

ryanjbaxter:
Right I agree.  I think its fine to drop the explicit dependency in the BOM and rely on `kubernetes-client`.  If they do drop it we can address what approach to take if/when that happens.

wind57:
in that case, this is ready to be merged and the defect that this triggered, also. thank you or looking into this.

PR #852, opened 2021-08-12
Minor nitpicks k8s client config maps functionality

wind57:


spring-cloud-kubernetes-client-config/src/main/java/org/springframework/cloud/kubernetes/client/config/KubernetesClientConfigMapPropertySource.java
```diff
@@ -46,17 +47,18 @@ public KubernetesClientConfigMapPropertySource(CoreV1Api coreV1Api, String name,
 			Environment environment) {
 
 		try {
-			List<String> names = new ArrayList<>();
+			Set<String> names = new HashSet<>();
```
wind57, line 50: unique config map names, so use `HashSet` instead of a `ArrayList`

spring-cloud-kubernetes-client-config/src/main/java/org/springframework/cloud/kubernetes/client/config/KubernetesClientConfigMapPropertySource.java
```diff
@@ -46,17 +47,18 @@ public KubernetesClientConfigMapPropertySource(CoreV1Api coreV1Api, String name,
...
 			names.add(name);
 			if (environment != null) {
 				for (String activeProfile : environment.getActiveProfiles()) {
 					names.add(name + "-" + activeProfile);
 				}
 			}
-			Map<String, Object> result = new LinkedHashMap<>();
+			Map<String, Object> result = new HashMap<>();
```
wind57, line 57: the order of properties that are stored inside a config map should not matter (we did the same change in fabric8 btw), so replace with a `HashMap`

spring-cloud-kubernetes-client-config/src/main/java/org/springframework/cloud/kubernetes/client/config/KubernetesClientConfigMapPropertySource.java
```diff
@@ -46,17 +47,18 @@ public KubernetesClientConfigMapPropertySource(CoreV1Api coreV1Api, String name,
...
 			coreV1Api.listNamespacedConfigMap(namespace, null, null, null, null, null, null, null, null, null, null)
 					.getItems().stream().filter(cm -> names.contains(cm.getMetadata().getName()))
-					.forEach(map -> result.putAll(processAllEntries(map.getData(), environment)));
```
wind57, line 59: `AnyCollection::forEach` is different then `Stream::forEach`; so first collect to a `List`, then to a `forEach`

spring-cloud-kubernetes-client-config/src/main/java/org/springframework/cloud/kubernetes/client/config/KubernetesClientConfigUtils.java
```diff
@@ -53,22 +53,14 @@ public static String getNamespace(SecretsConfigProperties.NormalizedSource norma
 
 	public static String getNamespace(ConfigMapConfigProperties.NormalizedSource normalizedSource,
 			String fallbackNamespace) {
-		if (!StringUtils.hasText(normalizedSource.getNamespace())) {
-			return fallbackNamespace;
-		}
-		else {
-			return normalizedSource.getNamespace();
-		}
```
wind57, line 61: minor nitpick

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/KubernetesNamespaceProvider.java
```diff
@@ -22,16 +22,16 @@
...
-import org.springframework.context.ApplicationEvent;
-import org.springframework.context.ApplicationListener;
 import org.springframework.core.env.Environment;
 
 import static org.springframework.cloud.kubernetes.commons.KubernetesClientProperties.SERVICE_ACCOUNT_NAMESPACE_PATH;
 
 /**
  * @author Ryan Baxter
  */
-public class KubernetesNamespaceProvider implements ApplicationListener<ApplicationEvent> {
```
wind57, line 34: I removed `implements ApplicationListener<ApplicationEvent>` and moved the `LOG.replayTo(KubernetesNamespaceProvider.class);` into the constructor. I am actually not sure why that `implements` exists, is it for the case when reload happens?

wind57:
@ryanjbaxter ready for review, with a pending question in the PR though.

spring-cloud-kubernetes-client-config/src/main/java/org/springframework/cloud/kubernetes/client/config/KubernetesClientConfigMapPropertySource.java
ryanjbaxter: Hmmm it could matter though.

If (for some strange reason) you have the same property defined twice in a config map I would expect the last one to take precedence.

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/KubernetesNamespaceProvider.java
ryanjbaxter: I thought it was a timing thing...I thought the `DeferredLog` needed to be initialized when the application started, if not its fine to remove it

wind57: yeah, I've tested this locally and the logs are present if it is initialized in the constructor.

spring-cloud-kubernetes-client-config/src/main/java/org/springframework/cloud/kubernetes/client/config/KubernetesClientConfigMapPropertySource.java
wind57: I am not very sure I understand your point here; this code is not about same property from same config map. In contrast, it is about properties from different config maps.

So you process `config-map-A`, get all of its properties (the order in which you get them is undefined for both clients), put these into `result`. Process `config-map-B`, put all its properties into same `result`. As such, "last property" wins, is still preserved. If I understood your point correctly.

ryanjbaxter: Makes sense

ryanjbaxter (approved):

PR #855, opened 2021-08-20
moved code + tests added

wind57:


spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/ConfigMapConfigProperties.java
```diff
@@ -176,28 +176,8 @@ public NormalizedSource normalize(String defaultName, String defaultNamespace) {
 		public NormalizedSource normalize(String defaultName, String defaultNamespace, boolean defaultUseNameAsPrefix) {
 			String normalizedName = StringUtils.hasLength(this.name) ? this.name : defaultName;
 			String normalizedNamespace = StringUtils.hasLength(this.namespace) ? this.namespace : defaultNamespace;
-
```
wind57, line 179: came out as I have realized that more code will be needed here (related to profile name suffix), so this will become very messy. AS such move it to the already existing Util class and while here, let's add tests to it. Though, proper tests are already enforced elsewhere; but to be as safe as possible...

wind57:
@ryanjbaxter one more, please.

ryanjbaxter (approved):

PR #866, opened 2021-09-11
clean-up @Deprecated things

wind57:


wind57:
I'll keep this here as a WIP thing. At some point in time, when we move to a higher version, `@Deprecated` things need to be removed. They make the code clutter, hard to maintain and hard to support. 

Once in a while I will come here and slowly clean-up the code via new commits.

Issue #769, opened 2021-03-25
Support for fail policy on the event that ConfigMap can’t be read


gabortega:
Hello,

Due to business requirements, we require the use of a Kubernetes client for an app running on Spring Boot 2.4. Whilst testing the ConfigMap loader (specified via a bootstrap file), we noticed that, in the event that the ConfigMap cannot be loaded, the app simply ignores this issue and continues the start-up process. For our requirements, our app must at least fail to start-up if it cannot load the ConfigMap.

At first glance, we didn’t see anything in the documentation suggesting a possible workaround to this… Then, going through older issues, we see that something similar to this has already been asked for (i.e. #441 ) but in the context of a retry policy. 

We would like to know if there is, by any chance, a fail policy feature already being worked on?

If not, is there an already existing workaround for this? 

Thank you for your attention on this matter.

As a side note, for the long term, we would also be interested in the retry policy.

frosiere:
Hello,

Any feedback would be more than welcome...

François

frosiere:
Is there someone maintaining this project and able to simply respond to the question or gives something to progress with the issue?! Thanks.

PR #873, opened 2021-09-28
Kubernetes Config Fail-fast and Retry Support

isikerhan:
Hello,

This PR adds support for fail fast and retry loading Kubernetes ConfigMaps and Secrets to resolve issues #769 and #441. There are some PRs around about these, but they seem to be stalled.

Thank you.

### DIFF
--- a/docs/src/main/asciidoc/property-source-config.adoc
+++ b/docs/src/main/asciidoc/property-source-config.adoc
@@ -415,6 +415,11 @@ NOTE:  If you use `spring.cloud.kubernetes.config.paths` or `spring.cloud.kubern
 functionality will not work.  You will need to make a `POST` request to the `/actuator/refresh` endpoint or
 restart/redeploy the application.
 
+[#config-map-fail-fast]
+In some cases, your application may be unable to load some of your `ConfigMaps` using the Kubernetes API.
+If you want your application to fail the start-up process in such cases, you can set the
+`spring.cloud.kubernetes.config.fail-fast=true` to make the application start-up fail with an Exception.
+
 .Properties:
 [options="header,footer"]
 |===
@@ -424,6 +429,7 @@ restart/redeploy the application.
 | `spring.cloud.kubernetes.config.namespace` | `String`  | Client namespace             | Sets the Kubernetes namespace where to lookup
 | `spring.cloud.kubernetes.config.paths`     | `List`    | `null`                       | Sets the paths where `ConfigMap` instances are mounted
 | `spring.cloud.kubernetes.config.enableApi` | `Boolean` | `true`                       | Enable or disable consuming `ConfigMap` instances through APIs
+| `spring.cloud.kubernetes.config.fail-fast` | `Boolean` | `false`                      | Enable or disable failing the application start-up when an error occurred while loading a `ConfigMap`
 |===
 
 === Secrets PropertySource
@@ -578,6 +584,8 @@ the `Secret` named `s1` would be looked up in the namespace that the application
 See <<namespace-resolution,namespace-resolution>> to get a better understanding of how the namespace
 of the application is resolved.
 
+<<config-map-fail-fast,Similar to the `ConfigMaps`>>; if you want your application to fail to start 
+when it is unable to load `Secrets` property sources, you can set `spring.cloud.kubernetes.secrets.fail-fast=true`.
 
 .Properties:
 [options="header,footer"]
@@ -589,6 +597,7 @@ of the application is resolved.
 | `spring.cloud.kubernetes.secrets.labels`    | `Map`     | `null`                       | Sets the labels used to lookup secrets
 | `spring.cloud.kubernetes.secrets.paths`     | `List`    | `null`                       | Sets the paths where secrets are mounted (example 1)
 | `spring.cloud.kubernetes.secrets.enableApi` | `Boolean` | `false`                      | Enables or disables consuming secrets through APIs (examples 2 and 3)
+| `spring.cloud.kubernetes.secrets.fail-fast` | `Boolean` | `false`                      | Enable or disable failing the application start-up when an error occurred while loading a `Secret`
 |===
 
 Notes:
--- a/docs/src/main/asciidoc/property-source-config.adoc
+++ b/docs/src/main/asciidoc/property-source-config.adoc
@@ -420,16 +420,27 @@ In some cases, your application may be unable to load some of your `ConfigMaps` 
 If you want your application to fail the start-up process in such cases, you can set the
 `spring.cloud.kubernetes.config.fail-fast=true` to make the application start-up fail with an Exception.
 
+[#config-map-retry]
+You can also make your application retry loading `ConfigMap` property sources on a failure. First, you need to
+set `spring.cloud.kubernetes.config.fail-fast=true`. Then you need to add `spring-retry` 
+and `spring-boot-starter-aop` to your classpath. You can configure retry properties such as
+the maximum number of attempts, backoff options like initial interval, multiplier, max interval by setting the
+`spring.cloud.kubernetes.config.retry.*` properties.
+
 .Properties:
 [options="header,footer"]
 |===
-| Name                                       | Type      | Default                      | Description
-| `spring.cloud.kubernetes.config.enabled`   | `Boolean` | `true`                       | Enable ConfigMaps `PropertySource`
-| `spring.cloud.kubernetes.config.name`      | `String`  | `${spring.application.name}` | Sets the name of `ConfigMap` to look up
-| `spring.cloud.kubernetes.config.namespace` | `String`  | Client namespace             | Sets the Kubernetes namespace where to lookup
-| `spring.cloud.kubernetes.config.paths`     | `List`    | `null`                       | Sets the paths where `ConfigMap` instances are mounted
-| `spring.cloud.kubernetes.config.enableApi` | `Boolean` | `true`                       | Enable or disable consuming `ConfigMap` instances through APIs
-| `spring.cloud.kubernetes.config.fail-fast` | `Boolean` | `false`                      | Enable or disable failing the application start-up when an error occurred while loading a `ConfigMap`
+| Name                                                    | Type      | Default                      | Description
+| `spring.cloud.kubernetes.config.enabled`                | `Boolean` | `true`                       | Enable ConfigMaps `PropertySource`
+| `spring.cloud.kubernetes.config.name`                   | `String`  | `${spring.application.name}` | Sets the name of `ConfigMap` to look up
+| `spring.cloud.kubernetes.config.namespace`              | `String`  | Client namespace             | Sets the Kubernetes namespace where to lookup
+| `spring.cloud.kubernetes.config.paths`                  | `List`    | `null`                       | Sets the paths where `ConfigMap` instances are mounted
+| `spring.cloud.kubernetes.config.enableApi`              | `Boolean` | `true`                       | Enable or disable consuming `ConfigMap` instances through APIs
+| `spring.cloud.kubernetes.config.fail-fast`              | `Boolean` | `false`                      | Enable or disable failing the application start-up when an error occurred while loading a `ConfigMap`
+| `spring.cloud.kubernetes.config.retry.initial-interval` | `Long`    | `1000`                       | Initial retry interval in milliseconds.
+| `spring.cloud.kubernetes.config.retry.max-attempts`     | `Integer` | `6`                          | Maximum number of attempts.
+| `spring.cloud.kubernetes.config.retry.max-interval`     | `Long`    | `2000`                       | Maximum interval for backoff.
+| `spring.cloud.kubernetes.config.retry.multiplier`       | `Double`  | `1.1`                        | Multiplier for next interval.
 |===
 
 === Secrets PropertySource
@@ -587,17 +598,27 @@ of the application is resolved.
 <<config-map-fail-fast,Similar to the `ConfigMaps`>>; if you want your application to fail to start 
 when it is unable to load `Secrets` property sources, you can set `spring.cloud.kubernetes.secrets.fail-fast=true`.
 
+It is also possible to enable retry for `Secret` property sources <<config-map-retry,like the `ConfigMaps`>>.
+As with the `ConfigMap` property sources, first you need to set `spring.cloud.kubernetes.secrets.fail-fast=true`. 
+Then you need to add `spring-retry` and `spring-boot-starter-aop` to your classpath. 
+Retry behavior of the `Secret` property sources can be configured by setting `spring.cloud.kubernetes.secrets.retry.*`
+configuration properties.
+
 .Properties:
 [options="header,footer"]
 |===
-| Name                                        | Type      | Default                      | Description
-| `spring.cloud.kubernetes.secrets.enabled`   | `Boolean` | `true`                       | Enable Secrets `PropertySource`
-| `spring.cloud.kubernetes.secrets.name`      | `String`  | `${spring.application.name}` | Sets the name of the secret to look up
-| `spring.cloud.kubernetes.secrets.namespace` | `String`  | Client namespace             | Sets the Kubernetes namespace where to look up
-| `spring.cloud.kubernetes.secrets.labels`    | `Map`     | `null`                       | Sets the labels used to lookup secrets
-| `spring.cloud.kubernetes.secrets.paths`     | `List`    | `null`                       | Sets the paths where secrets are mounted (example 1)
-| `spring.cloud.kubernetes.secrets.enableApi` | `Boolean` | `false`                      | Enables or disables consuming secrets through APIs (examples 2 and 3)
-| `spring.cloud.kubernetes.secrets.fail-fast` | `Boolean` | `false`                      | Enable or disable failing the application start-up when an error occurred while loading a `Secret`
+| Name                                                     | Type      | Default                      | Description
+| `spring.cloud.kubernetes.secrets.enabled`                | `Boolean` | `true`                       | Enable Secrets `PropertySource`
+| `spring.cloud.kubernetes.secrets.name`                   | `String`  | `${spring.application.name}` | Sets the name of the secret to look up
+| `spring.cloud.kubernetes.secrets.namespace`              | `String`  | Client namespace             | Sets the Kubernetes namespace where to look up
+| `spring.cloud.kubernetes.secrets.labels`                 | `Map`     | `null`                       | Sets the labels used to lookup secrets
+| `spring.cloud.kubernetes.secrets.paths`                  | `List`    | `null`                       | Sets the paths where secrets are mounted (example 1)
+| `spring.cloud.kubernetes.secrets.enableApi`              | `Boolean` | `false`                      | Enables or disables consuming secrets through APIs (examples 2 and 3)
+| `spring.cloud.kubernetes.secrets.fail-fast`              | `Boolean` | `false`                      | Enable or disable failing the application start-up when an error occurred while loading a `Secret`
+| `spring.cloud.kubernetes.secrets.retry.initial-interval` | `Long`    | `1000`                       | Initial retry interval in milliseconds.
+| `spring.cloud.kubernetes.secrets.retry.max-attempts`     | `Integer` | `6`                          | Maximum number of attempts.
+| `spring.cloud.kubernetes.secrets.retry.max-interval`     | `Long`    | `2000`                       | Maximum interval for backoff.
+| `spring.cloud.kubernetes.secrets.retry.multiplier`       | `Double`  | `1.1`                        | Multiplier for next interval.
 |===
 
 Notes:
--- a/spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/ConditionalOnConfigMapPropertiesRetryDisabled.java
+++ b/spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/ConditionalOnConfigMapPropertiesRetryDisabled.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright 2013-2021 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.kubernetes.commons.config;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Inherited;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import org.springframework.boot.autoconfigure.condition.NoneNestedConditions;
+import org.springframework.context.annotation.Conditional;
+
+/**
+ * {@link Conditional @Conditional} that matches when at least one of Spring Cloud
+ * Kubernetes, Kubernetes ConfigMap property sources or Kubernetes ConfigMap property
+ * sources fail fast (thus retry) is disabled.
+ *
+ * @author Isik Erhan
+ */
+@Target({ ElementType.TYPE, ElementType.METHOD })
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+@Inherited
+@Conditional(ConditionalOnConfigMapPropertiesRetryDisabled.OnConfigMapPropertiesRetryDisabled.class)
+public @interface ConditionalOnConfigMapPropertiesRetryDisabled {
+
+	class OnConfigMapPropertiesRetryDisabled extends NoneNestedConditions {
+
+		OnConfigMapPropertiesRetryDisabled() {
+			super(ConfigurationPhase.REGISTER_BEAN);
+		}
+
+		@ConditionalOnConfigMapPropertiesRetryEnabled
+		static class OnConfigMapPropertiesRetryEnabled {
+
+		}
+
+	}
+
+}
--- a/spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/ConditionalOnConfigMapPropertiesRetryEnabled.java
+++ b/spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/ConditionalOnConfigMapPropertiesRetryEnabled.java
@@ -1,0 +1,46 @@
+/*
+ * Copyright 2013-2021 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.kubernetes.commons.config;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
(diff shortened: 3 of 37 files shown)


wind57:
this is very nice! I would very much like to review once I have a bit more time.

ryanjbaxter:
This looks really good to me.  Would you mind pulling in the latest changes to main so we can try and run the ci build again?

docs/src/main/asciidoc/property-source-config.adoc
```diff
@@ -415,15 +415,32 @@ NOTE:  If you use `spring.cloud.kubernetes.config.paths` or `spring.cloud.kubern
...
 restart/redeploy the application.
 
+[#config-map-fail-fast]
+In some cases, your application may be unable to load some of your `ConfigMaps` using the Kubernetes API.
+If you want your application to fail the start-up process in such cases, you can set the
+`spring.cloud.kubernetes.config.fail-fast=true` to make the application start-up fail with an Exception.
+
+[#config-map-retry]
+You can also make your application retry loading `ConfigMap` property sources on a failure. First, you need to
+set `spring.cloud.kubernetes.config.fail-fast=true`. Then you need to add `spring-retry`
```
wind57, line 425: it's probably obvious to you, but can I ask why `spring-retry` and `spring-boot-starter-aop` are not on the classpath and need to be added?

ryanjbaxter: Its a pattern we have used in the past with Spring Retry, ultimately we only enable retry functionality if the user specifically wants it, so we require them to include Spring Retry and aop on their app's classpath.

isikerhan:
> This looks really good to me. Would you mind pulling in the latest changes to main so we can try and run the ci build again?

Pulled in the latest changes, and the build succeeded. @ryanjbaxter

docs/src/main/asciidoc/property-source-config.adoc
wind57: so, declare them `optional` and then use `@ConditionalOnClass({ Retryable.class, Aspect.class, AopAutoConfiguration.class })`, right?

isikerhan: @wind57 yes exactly. Also the Spring Cloud Config Client works in the same way :)

ryanjbaxter (approved):

docs/src/main/asciidoc/property-source-config.adoc
wind57: thank you both, an interesting pattern.

wind57: Ryan can I please ask for this not to be merged for a few days? I think I see a breaking change that I would like to have a look a closer look at. please.

ryanjbaxter: Sure no rush, curious what the change that you think breaks something?

docs/src/main/asciidoc/property-source-config.adoc
```diff
@@ -578,17 +595,30 @@ the `Secret` named `s1` would be looked up in the namespace that the application
 See <<namespace-resolution,namespace-resolution>> to get a better understanding of how the namespace
 of the application is resolved.
 
+<<config-map-fail-fast,Similar to the `ConfigMaps`>>; if you want your application to fail to start 
+when it is unable to load `Secrets` property sources, you can set `spring.cloud.kubernetes.secrets.fail-fast=true`.
```
wind57, line 599: I am sorry, but this looks very confusing to me. We say: "if I can't read a Secret and `spring.cloud.kubernetes.secrets.fail-fast=true`, we will fail without retrying". So let's take a step back and see what we do now, without this PR in place: 

- if we can't read a Secret, we will ignore this Exception and continue to start the application. 

This is what would happen if  `spring.cloud.kubernetes.secrets.fail-fast=false` (which is the default), which is good because it means we preserve backwards compatibility in user expectations. 

Now if we enable `spring.cloud.kubernetes.secrets.fail-fast=true` and we can't read a property, we fail - which is again very nice and without surprises.

Now, we want to enable retry and for that we say:

- `spring.cloud.kubernetes.secrets.fail-fast=true`
- put retry and aop on classpath. 

Isn't this confusing to you? The different between retry or not is something on the classpath. _Both_ cases mandate the presence of `spring.cloud.kubernetes.secrets.fail-fast=true`, let me emphasize the `fail-fast` in the naming, which, depending on what you have on the classpath is _not_ a "fail fast"?

As a user of this I would expect a `fail-fast` with `true/false` and a `retry` with `true/false`. 

I am missing something?  

---

This btw has some more implications that I have not yet reviewed entirely. For example the: `ConditionalOnConfigMapPropertiesRetryEnabled`. Look at the name `...RetryEnabled` with a `name = "fail-fast", havingValue = "true"`.  As an exercise (to see if I am not the real problem here), I explained this to 2 of my colleagues at work and they had the same confusion. 

I bet there is something I miss here, so please advise.

wind57: I have been thinking about this and as a matter of fact found a real case in our repository.

We have `spring-cloud-kubernetes` and `spring-retry` in the same project. If we merge this the way it is and for whatever reason I would like to say "I'd like to fail if I can't read a ConfigMap without ever retrying", I will be stuck as I can't do that. Essentially:

- I already have `spring-retry` and `aop` on the classpath because of business implications.
- I would specify  `spring.cloud.kubernetes.secrets.fail-fast=true`

This case will indeed be documented, no question about it. At the same time, I can't opt out of it. 

Unless there is an explicit flag to control both the fail and retries, someone might not like this.

docs/src/main/asciidoc/property-source-config.adoc
wind57: I will add my concern as soon as I reach that point in my review Ryan. It's a fabulous PR, so I will take a bit of time on it.

docs/src/main/asciidoc/property-source-config.adoc
isikerhan: Adding a separate flag for retry was something I've considered during the development. Spring Cloud Config Client and Spring Cloud Kubernetes Config do pretty much the same thing. One loads config from a config server, the other loads them from the Kubernetes objects. And I would expect them to bahave as similar as possible as a user. So I ended up sticking with a very similar design to the config client to keep the ecosystem consistent. The name "fail-fast" also comes from there :) I don't know, I may be wrong with this approach. Btw I don't think that adding one more flag is such a big deal and will create a huge inconsistency. IMHO, we can still change the design unless we get far away from the way other modules does the similar things. What do you think @ryanjbaxter ?

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/ConditionalOnConfigMapPropertiesRetryEnabled.java
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * @author Isik Erhan
+ */
+@Target({ ElementType.TYPE, ElementType.METHOD })
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+@Inherited
+@ConditionalOnKubernetesEnabled
+@ConditionalOnKubernetesConfigEnabled
+@ConditionalOnProperty(prefix = ConfigMapConfigProperties.PREFIX, name = "fail-fast", havingValue = "true")
```
wind57, line 43: what if we extract `@ConditionalOnProperty(prefix = ConfigMapConfigProperties.PREFIX, name = "fail-fast", havingValue = "true")` to its own annotation called `ConditionalOnKubernetesConfigRetryEnabled` and use that here: 

```
@ConditionalOnKubernetesEnabled
@ConditionalOnKubernetesConfigEnabled
@ConditionalOnKubernetesConfigRetryEnabled
```

reads pretty awesome, to me. what do you think?

isikerhan: Actually, the names `@ConditionalOnConfigMapPropertiesRetryEnabled` and `@ConditionalOnKubernetesConfigRetryEnabled` both refer to the same logical condition. However, I like the name you suggested; it's clearer and a bit shorter. Maybe we can rename `@ConditionalOnConfigMapPropertiesRetryEnabled` itself to `@ConditionalOnKubernetesConfigRetryEnabled` and extract `@ConditionalOnProperty(prefix = ConfigMapConfigProperties.PREFIX, name = "fail-fast", havingValue = "true")` to something like `@ConditionalOnKubernetesConfigFailFastEnabled`. It would look like the following:

ConditionalOnKubernetesConfigRetryEnabled .java:
```java
@ConditionalOnKubernetesEnabled
@ConditionalOnKubernetesConfigEnabled
@ConditionalOnKubernetesConfigFailFastEnabled
public @interface ConditionalOnKubernetesConfigRetryEnabled {
...
```

ConditionalOnKubernetesConfigFailFastEnabled.java
```java
@ConditionalOnProperty(prefix = ConfigMapConfigProperties.PREFIX, name = "fail-fast", havingValue = "true")
public @interface ConditionalOnKubernetesConfigFailFastEnabled {
...
```

The only thing that I don't like with this approach is the extracted condition `@ConditionalOnKubernetesConfigFailFastEnabled` will not check if Kubernetes itself or Kubernetes config is enabled or not. This could make it a bit useless when someone wants it to use in another place since it does not actually check if "Kubernetes config fail-fast" feature is enabled, it only checks if the property `*.fail-fast` is equal to `true`.

wind57:
```
@ConditionalOnKubernetesEnabled
@ConditionalOnKubernetesConfigEnabled
@ConditionalOnKubernetesConfigFailFastEnabled
public @interface ConditionalOnKubernetesConfigRetryEnabled {
```

this! very much this; good name, I like it more than what I proposed initially. 

As to you concern, imo : if we document the fact that this does not check neither for kubernetes nor config map - it's _documented_ now, so no problem? :)

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/KubernetesBootstrapConfiguration.java
```diff
@@ -16,16 +16,69 @@
...
+		}
+
+		@Bean
+		@ConditionalOnConfigMapPropertiesRetryEnabled
+		public RetryOperationsInterceptor configMapPropertiesRetryInterceptor(
+				ConfigMapConfigProperties configProperties) {
+			return retryOperationsInterceptor(configProperties.getRetry());
+		}
+
+		@Bean("configMapPropertiesRetryInterceptor")
```
wind57, line 64: can I ask why you name these with the same name, though one does not retry and the other does? Or this is again for consistency across projects?

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/KubernetesBootstrapConfiguration.java
```diff
@@ -16,16 +16,69 @@
...
+			return RetryInterceptorBuilder.stateless().retryPolicy(new NeverRetryPolicy()).build();
+		}
+
+		@Bean
+		@ConditionalOnSecretsPropertiesRetryEnabled
+		public RetryOperationsInterceptor secretsPropertiesRetryInterceptor(SecretsConfigProperties configProperties) {
+			return retryOperationsInterceptor(configProperties.getRetry());
+		}
+
+		@Bean("secretsPropertiesRetryInterceptor")
```
wind57, line 76: I did not understand initially why you have `@Bean("secretsPropertiesRetryInterceptor")`, but darn this is interesting. I don't know why but I would have expected for `@Retryable` to take an array of `interceptors`? And take the first one present in context? This way, you could have had two clean beans here:

```

		@Bean
		@ConditionalOnSecretsPropertiesRetryEnabled
		public RetryOperationsInterceptor secretsPropertiesRetryInterceptor(SecretsConfigProperties configProperties) {
			return retryOperationsInterceptor(configProperties.getRetry());
		}

		@Bean
		@ConditionalOnSecretsPropertiesRetryDisabled
		public RetryOperationsInterceptor secretsPropertiesNoRetryInterceptor() {
			return RetryInterceptorBuilder.stateless().retryPolicy(new NeverRetryPolicy()).build();
		}
```

and `@Retryable(interceptors = { "secretsPropertiesRetryInterceptor", "secretsPropertiesNoRetryInterceptor" })`. 

This is pure theoretical and I do not know what I am missing, but looking at this hack here, I would have wanted that to be a reality.

isikerhan: Unfortunately, `@Retryable` doesn't work like that. It takes a single `interceptor` bean name and tries to get the bean with that name from the bean factory if the bean name is not empty. And if the bean itself is not present in the bean factory, then it fails with `NoSuchBeanDefinitionException`. (see [AnnotationAwareRetryOperationsInterceptor.java](https://github.com/spring-projects/spring-retry/blob/main/src/main/java/org/springframework/retry/annotation/AnnotationAwareRetryOperationsInterceptor.java#L185-L188)) I'd expect it to work normally without retrying in this case but apparently, it doesn't work like that.

wind57: or like that, yes. Either way, this hack you made here is nice, thank you for the follow-up here.

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/KubernetesBootstrapConfiguration.java
```diff
@@ -16,16 +16,69 @@
...
  */
 @Configuration(proxyBeanMethods = false)
 @ConditionalOnKubernetesEnabled
 @EnableConfigurationProperties({ ConfigMapConfigProperties.class, SecretsConfigProperties.class })
 public class KubernetesBootstrapConfiguration {
 
+	@ConditionalOnKubernetesConfigPropertiesRetryEnabled
+	@ConditionalOnClass({ Retryable.class, Aspect.class, AopAutoConfiguration.class })
+	@Configuration(proxyBeanMethods = false)
+	@EnableRetry(proxyTargetClass = true)
```
wind57, line 46: btw, forgot to ask here, why `proxyTargetClass = true`? why cglib?

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/ConditionalOnKubernetesConfigPropertiesRetryEnabled.java
```diff
@@ -0,0 +1,62 @@
...
+ * @author Isik Erhan
+ */
+@Target({ ElementType.TYPE, ElementType.METHOD })
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+@Inherited
+@Conditional(ConditionalOnKubernetesConfigPropertiesRetryEnabled.OnKubernetesConfigPropertiesRetryEnabled.class)
+public @interface ConditionalOnKubernetesConfigPropertiesRetryEnabled {
+
+	class OnKubernetesConfigPropertiesRetryEnabled extends AnyNestedCondition {
```
wind57, line 44: I feel like I am starting to sound like a broken record here :| what do you think renaming this to : `ConditionalOnKubernetesConfigOrSecretsRetryEnabled`?

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/ConditionalOnConfigMapPropertiesRetryDisabled.java
```diff
@@ -0,0 +1,56 @@
...
+ * @author Isik Erhan
+ */
+@Target({ ElementType.TYPE, ElementType.METHOD })
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+@Inherited
+@Conditional(ConditionalOnConfigMapPropertiesRetryDisabled.OnConfigMapPropertiesRetryDisabled.class)
+public @interface ConditionalOnConfigMapPropertiesRetryDisabled {
+
+	class OnConfigMapPropertiesRetryDisabled extends NoneNestedConditions {
```
wind57, line 43: This is the first time I have used or reviewed some code with `NoneNestedConditions`, this in turn sparked my interest. I removed this part of the code below (after cloning your repo locally) : 

```
@ConditionalOnConfigMapPropertiesRetryEnabled
static class OnConfigMapPropertiesRetryEnabled {

}
```

did a `mvn clean package` on the whole project and to my surprise no test failed. :) Interesting. Took a while, but I found no way to make such a test. Ideally, such a test should be present in `KubernetesBootstrapConfigurationTests::SecretsFailFastEnabled::shouldDefineRequiredBeans`. The problem goes back to the one that we already talked before, that `@Retryable` and the need for the same interceptor name for the case when we want and not, to retry. 

On the other hand, this exercise made me go through the code and tests is much more detail and appreciate the effort you have put into this.

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/ConditionalOnSecretsPropertiesRetryEnabled.java
```diff
@@ -0,0 +1,46 @@
...
+ * @author Isik Erhan
+ */
+@Target({ ElementType.TYPE, ElementType.METHOD })
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+@Inherited
+@ConditionalOnKubernetesEnabled
+@ConditionalOnKubernetesSecretsEnabled
+@ConditionalOnProperty(prefix = SecretsConfigProperties.PREFIX, name = "fail-fast", havingValue = "true")
+public @interface ConditionalOnSecretsPropertiesRetryEnabled {
```
wind57, line 44: its probably obvious, but the same renamings we talked about for ConfigMap are going to be applied here too. Just making sure we do not miss it :)

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/ConfigMapPropertySourceLocator.java
```diff
@@ -80,6 +84,12 @@ protected abstract MapPropertySource getMapPropertySource(String applicationName
 		return null;
 	}
 
+	@Override
+	@Retryable(interceptor = "configMapPropertiesRetryInterceptor")
```
wind57, line 88: you are overriding this only for the sake of `@Retryable`. You are doing this as a safety net, correct? So that just in case we ever use `locateCollection`, it will have that `@Retryable`? If so, this is a great idea (I would have missed it if it was me doing this PR!).

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/KubernetesBootstrapConfiguration.java
```diff
@@ -16,16 +16,69 @@
...
 /**
  * @author Ryan Baxter
+ * @author Isik Erhan
  */
 @Configuration(proxyBeanMethods = false)
 @ConditionalOnKubernetesEnabled
 @EnableConfigurationProperties({ ConfigMapConfigProperties.class, SecretsConfigProperties.class })
 public class KubernetesBootstrapConfiguration {
 
+	@ConditionalOnKubernetesConfigPropertiesRetryEnabled
```
wind57, line 43: just a word of appreciation that I liked this pattern: activate the Config for either of the conditionals, but allow to disable them individually. Nice.

spring-cloud-kubernetes-fabric8-config/src/main/java/org/springframework/cloud/kubernetes/fabric8/config/Fabric8SecretsPropertySource.java
```diff
@@ -64,6 +67,11 @@ public Fabric8SecretsPropertySource(KubernetesClient client, String name, String
 
 		}
 		catch (Exception e) {
+			if (failFast) {
```
wind57, line 70: here is the concern I have and would like to hear your thoughts. Suppose we are reading `ConfigMapA`, at the same time we have enabled `fail-fast`, be that with or without retry, it is irrelevant to the point I am trying to make.

Now let's suppose that this `ConfigMapA` is read perfectly fine on the first attempt, no question here. At the same time, because of our flow, we will be trying to read `ConfigMapA-kubernetes` map. This happens because we add the `kubernetes` profile in the `AbstractKubernetesProfileEnvironmentPostProcessor::postProcessEnvironment` and in `KubernetesClientConfigMapPropertySource::getData` we generate and try to read from such a map (same idea applies for `fabric8`). 

Obviously such a ConfigMap is not present, in the vast majority of cases (if not all). We, on the other hand, will fail the application start-up, because of `fail-fast`. Users are not going to be happy and I guess even surprised.

There are two points here:

- there is a PR that I introduced that allows you to opt out of such `ConfigMapName-<ActiveProfiles>` readings, but it is not yet in `main` nor it will entirely fix this issue here. It allows to opt out of _all_ ConfigMap readings. So if you have a `ConfigMapB`, you can't say : "opt-out of `ConfigMapA-kubernetes`, but allow `ConfigMapB-kubernetes`"; neither do I think we want such granularity in that PR. 

- Initially, I thought that may be we should just ignore every failures related to `XXX-<ActiveProfile>`, but what if someone already relies on that? And this touches my first point above, of course. What if there are users that have `ConfigMapMyApp` and `ConfigMapMyApp-PROD`. When they were deploying to non PROD environments, the application would have started just fine (it only logged an Exception). If we merge this PR, and they enable retry (which is exactly what this PR wants them to do), it will now fail, because `ConfigMapMyApp-PROD` does not exist in non-prod environments. 

I will spent some more time trying to figure out an idea on how to approach this correctly (imho), but of course, your thoughts are absolutely a must too.

isikerhan: Indeed, we don't fail when a ConfigMap or a Secret is not found. (Since Kubernetes Java clients do not throw exception on 404) Instead, we create an empty `PropertySource `as far as I see. So, our application still won't fail just because a ConfigMap or a Secret is missing even we enable `fail-fast`. Btw, it's obvious that retrying to load unexistent ConfigMap/Secret is pointless even we have retry enabled.

The question here is: "Should we fail on not found?" IMHO this question belongs to another topic. Actually, I considered adding a failure policy per config/secrets source like below:

```yaml
spring:
  cloud:
    kubernetes:
      config:
        fail-fast: true # when this is set to false, all failure-policies will be normalized to "never"
        sources:
         - name: some-config-map
           failure-policy: fail-on-not-found # fails when the config map is not present or any error occurs while loading it
         - name: another-config-map
           failure-policy: fail-on-error # (default) fail only when an error occurs while loading the config map. ignore if it's not found
         - name: optional-config-map
           failure-policy: never # ignore not found or any errors
```

But the same problem you mentioned above emerges with that design: When do we consider a config source as "not found"? Should we consider a config source as not found when any of the ConfigMaps with names `MyConfigMap`, `MyConfigMap-<profile>`, is not found? Or is it enough to say a config source "is found" if we are able to load any of these? I ended up leaving this behavior as-is.

IMHO we can leave resolving these questions to another feature since we don't actually break anything with this PR (if I'm not mistaken). What do you think?

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/ConditionalOnConfigMapPropertiesRetryDisabled.java
isikerhan: This actually is about the order of the bean definitions in `KubernetesBootstrapConfiguration.RetryConfiguration`. It'd work even without annotating `configMapPropertiesRetryInterceptorNoRetry` method with `@ConditionalOnConfigMapPropertiesRetryDisabled` just beacuse of the order they're defined. If you change the bean definition order like below, some tests must start failing as you expected:

```java
@Bean("configMapPropertiesRetryInterceptor")
@ConditionalOnConfigMapPropertiesRetryDisabled
public RetryOperationsInterceptor configMapPropertiesRetryInterceptorNoRetry() {
	return RetryInterceptorBuilder.stateless().retryPolicy(new NeverRetryPolicy()).build();
}

@Bean
@ConditionalOnConfigMapPropertiesRetryEnabled
public RetryOperationsInterceptor configMapPropertiesRetryInterceptor(
		ConfigMapConfigProperties configProperties) {
	return retryOperationsInterceptor(configProperties.getRetry());
}
```

At this point, If you put the code block back that you removed, the tests must start running successfully again. With a correct ordering, `@ConditionalOnConfigMapPropertiesRetryDisabled` may seem a little bit useless, but I don't really like to rely on the order of something.

isikerhan: This actually is about the order of the bean definitions in `KubernetesBootstrapConfiguration.RetryConfiguration`. It'd work even without annotating `configMapPropertiesRetryInterceptorNoRetry` method with `@ConditionalOnConfigMapPropertiesRetryDisabled` just beacuse of the order they're defined. If you change the bean definition order like below, some tests must start failing as you expected:

```java
@Bean("configMapPropertiesRetryInterceptor")
@ConditionalOnConfigMapPropertiesRetryDisabled
public RetryOperationsInterceptor configMapPropertiesRetryInterceptorNoRetry() {
	return RetryInterceptorBuilder.stateless().retryPolicy(new NeverRetryPolicy()).build();
}

@Bean
@ConditionalOnConfigMapPropertiesRetryEnabled
public RetryOperationsInterceptor configMapPropertiesRetryInterceptor(
		ConfigMapConfigProperties configProperties) {
	return retryOperationsInterceptor(configProperties.getRetry());
}
```

At this point, if you put the code block that you removed back, the tests must start running properly again. With a correct ordering, `@ConditionalOnConfigMapPropertiesRetryDisabled` may seem a little bit useless, but I don't really like to rely on the order of something. It just provides extra safety here.

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/ConfigMapPropertySourceLocator.java
isikerhan: Thank you and yes. I'm overriding `locateCollection` to mark it with `@Retryable`.

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/ConditionalOnConfigMapPropertiesRetryDisabled.java
wind57: yeah, I had to take a closer look today and this is the same conclusion I came to. thank you.

spring-cloud-kubernetes-fabric8-config/src/main/java/org/springframework/cloud/kubernetes/fabric8/config/Fabric8SecretsPropertySource.java
wind57: this is what I get when I comment without trying and solely based on my memory. I definitely used to see these errors on start-up, but I do not remember the context anymore :( I am really sorry for this comment without actually trying it first - I did not, for both clients and we are good indeed.  

Teşekkürler!

I am more than OK with this PR.

docs/src/main/asciidoc/property-source-config.adoc
wind57: @ryanjbaxter no concerns and I am good to merge, once some minor things change.

spring-cloud-kubernetes-fabric8-config/src/main/java/org/springframework/cloud/kubernetes/fabric8/config/Fabric8SecretsPropertySource.java
isikerhan: No worries. And thank you very much for the review!

isikerhan:
@ryanjbaxter I guess we're good to go. Can you take a look and merge this if everything is OK?

PR #875, opened 2021-09-30
Fix 874

wind57:


wind57:
ready to be merged, @ryanjbaxter

ryanjbaxter:
I actually prefer the approach of specifying the builder, it just requires changes in 2 places to make it work.  I would rather Spring Boot use a specific version of the builder rather than the `base` tag so we avoid this problem in the future, but it doesn't look like they are going to change that.  Anyways I made the changes necessary and builds are working again.
https://github.com/spring-cloud/spring-cloud-kubernetes/blob/2.0.x/spring-cloud-kubernetes-controllers/spring-cloud-kubernetes-configuration-watcher/pom.xml#L85
https://github.com/spring-cloud/spring-cloud-kubernetes/blob/2.0.x/spring-cloud-kubernetes-integration-tests/run.sh#L123

wind57:
I understand your view here, thank you. This builder image obviously changes in time, this way we will have to upgrade it once in a while? I mean, the support for `NMT` was added, for example and many more might follow in the future. This way we tight ourselves to some particular version. But then again, we do the same with `kind`. 

I guess your reasoning derive more from the fact that this is for UT only, right?

ryanjbaxter:
What does UT mean?

I don't want to specify the version long term I want to revert my changes once the builder is fixed.

wind57:
my bad, `UT` == `unit tests`. 

> I don't want to specify the version long term I want to revert my changes once the builder is fixed.

Oh! nvm, this makes sense now. thank you

PR #897, opened 2021-10-30
split core test

wind57:


spring-cloud-kubernetes-fabric8-config/src/test/java/org/springframework/cloud/kubernetes/fabric8/config/CoreTest.java
```diff
@@ -50,13 +49,6 @@
 	@Autowired
 	private Config config;
 
-	// not a fan of changing the type from KubernetesClient, but because of:
```
wind57, line 53: the talks that I had on that defect made me realize that we indeed must split this test into two. If you follow the comments [under the actual issue](https://github.com/fabric8io/kubernetes-client/issues/3145), it all makes sense.

The change is trivial : split this test into two, and use `KubernetesClient` (instead of `DefaultKubernetesClient`) everywhere.

wind57:
@ryanjbaxter one more ready to review please

PR #898, opened 2021-10-31
User Agent Functionality

wind57:


spring-cloud-kubernetes-client-autoconfig/src/main/java/org/springframework/cloud/kubernetes/client/KubernetesClientAutoConfiguration.java
```diff
@@ -40,8 +42,11 @@
 
 	@Bean
 	@ConditionalOnMissingBean
-	public ApiClient apiClient() {
+	public ApiClient apiClient(KubernetesClientProperties properties) {
 		ApiClient apiClient = kubernetesApiClient();
```
wind57, line 46: user the default user agent, unless users have provided one

spring-cloud-kubernetes-client-autoconfig/src/test/java/org/springframework/cloud/kubernetes/client/KubernetesClientDefaultApiClientTests.java
```diff
@@ -1,61 +0,0 @@
...
-// io.kubernetes...KubernetesInformerAutoConfiguration
-// creates a defaultApiClient that will be autowired instead of the ApiClient
-// created in KubernetesClientAutoConfiguration
-public class KubernetesClientDefaultApiClientTests {
-
-	@Autowired
-	private ApiClient apiClient;
-
-	@Test
-	public void testCreatedApiClientIsNotDefault() {
```
wind57, line 47: this test was wrong, that is why I needed to move this class to a different package. The `apiClient` that was injected here actually came from `org.springframework.cloud.kubernetes.client.example.App` instead of the `KubernetesClientAutoConfiguration` like it was wanted. As such, it actually did not test what we hoped for. 

This was revealed when I started to add some more methods in here with regards to "User-Agent" functionality.

spring-cloud-kubernetes-client-autoconfig/src/test/java/org/springframework/cloud/kubernetes/client/default_api/KubernetesClientDefaultApiClientTests.java
```diff
@@ -0,0 +1,123 @@
...
+
+		@Autowired
+		private ApiClient apiClient;
+
+		@Autowired
+		ConfigurableApplicationContext context;
+
+		@Test
+		void testCreatedApiClientIsNotDefault() {
+			assertThat(apiClient).isNotNull();
```
wind57, line 65: because this is part of a separate package, it makes sure that `apiClient` comes from `KubernetesClientAutoConfiguration`

spring-cloud-kubernetes-fabric8-autoconfig/src/main/java/org/springframework/cloud/kubernetes/fabric8/Fabric8AutoConfiguration.java
```diff
@@ -96,7 +96,12 @@ public Config kubernetesClientConfig(KubernetesClientProperties kubernetesClient
 				.withHttpsProxy(or(kubernetesClientProperties.getHttpsProxy(), base.getHttpsProxy()))
 				.withProxyUsername(or(kubernetesClientProperties.getProxyUsername(), base.getProxyUsername()))
 				.withProxyPassword(or(kubernetesClientProperties.getProxyPassword(), base.getProxyPassword()))
-				.withNoProxy(or(kubernetesClientProperties.getNoProxy(), base.getNoProxy())).build();
+				.withNoProxy(or(kubernetesClientProperties.getNoProxy(), base.getNoProxy()))
+				.withUserAgent(
```
wind57, line 100: fabric8 already has a property for user-agent header, as such it is slightly different then native client. 

In case of the native client, it's easy:

- either `spring.cloud.kubernetes.client.userAgent` or 
- `Spring-Cloud-Kubernetes-Application`, if the above is missing

In case of `fabric8`:

- either `spring.cloud.kubernetes.client.userAgent`
- or `Config.KUBERNETES_USER_AGENT`
- or `Spring-Cloud-Kubernetes-Application`

My next commit will also document this properly.

spring-cloud-kubernetes-fabric8-autoconfig/src/test/java/org/springframework/cloud/kubernetes/fabric8/Fabric8ClientUserAgentEnvPropertyTests.java
```diff
@@ -0,0 +1,57 @@
...
+ * test "User-Agent" functionality via system properties
+ */
+@SpringBootTest(classes = App.class)
+class Fabric8ClientUserAgentEnvPropertyTests {
+
+	@Autowired
+	private KubernetesClient client;
+
+	@BeforeAll
+	static void beforeAll() {
```
wind57, line 42: this test is separate because of `@BeforeAll` that has to be static, and `@Nested` can not be static. From jdk-16 and up, `JLS` allows for `static` inside non-static, so this could change in the future.

wind57:
@ryanjbaxter ready for review

spring-cloud-kubernetes-client-autoconfig/src/main/java/org/springframework/cloud/kubernetes/client/KubernetesClientAutoConfiguration.java
```diff
@@ -40,8 +42,11 @@
...
+		String userAgent = StringUtils.hasText(properties.getUserAgent()) ? properties.getUserAgent()
```
ryanjbaxter, line 47: Why not just have the default value be set in `KubernetesClientProperties`?

ryanjbaxter (changes requested):
LGTM just one small nit

spring-cloud-kubernetes-client-autoconfig/src/main/java/org/springframework/cloud/kubernetes/client/KubernetesClientAutoConfiguration.java
wind57: I can't, it will mess with `fabric8` order in such a case : 

```
.withUserAgent(or(kubernetesClientProperties.getUserAgent(),
						or(base.getUserAgent(), KubernetesClientProperties.DEFAULT_USER_AGENT)))
				.build();
```

ryanjbaxter: What do you mean mess with this order?

wind57: let's see, may be I miss-understood your comment. I understood this:

> Why not just have the default value be set in KubernetesClientProperties?

as : 

```
	public String getUserAgent() {
		return userAgent;
	}
```

woud become:

```
	public String getUserAgent() {
		return StringUtils.hasLength(userAgent) ? userAgent : "Spring-Cloud-Kubernetes-Application";
	}
```

If this is the case, this would work just fine for `k8s-native` client, but not for `fabric8`, because for `fabric8` it would always ignore `base.getUserAgent()` in : 

```
.withUserAgent(
    or(kubernetesClientProperties.getUserAgent(),
   or(base.getUserAgent(), KubernetesClientProperties.DEFAULT_USER_AGENT)))
.build();
```

because `kubernetesClientProperties.getUserAgent()` would always provide a value...

---

Now that I think about it, it really seems you mean a rather different thing. If so, sorry for not getting

ryanjbaxter: Actually I was thinking this in `KubernetesClientProperties`
`private String userAgent = DEFAULT_USER_AGENT`

This would still result in the same problem though.  But I think we should check if `getUserAgent` returns the default value (meaning it was not set explicitly) then check `base.getUserAgent()`

wind57: I honestly can connect the dots in my head, can you please post some code here? This is the order I wanted to do:

 - `spring.cloud.kubernetes.client.userAgent`
 - base.getUserAgent()
 - `"Spring-Cloud-Kubernetes-Application"`

what is your suggestion here in order to keep the order the same?

ryanjbaxter: I think something like this would work.
in `KubernetesClientProperties`
```
private String userAgent = DEFAULT_USER_AGENT
```
Then in `Fabric8AutoConfiguration`
```
String userAgent = or(base.getUserAgent(), KubernetesClientProperties.DEFAULT_USER_AGENT);
if (!KubernetesClientProperties.DEFAULT_USER_AGENT.equals(kubernetesClientProperties.getUserAgent())) {
			userAgent = kubernetesClientProperties.getUserAgent();
}
....
.withUserAgent(userAgent);
```

wind57: oh wow! :) I would have never even thought about it like that. For me, this:

```
.withUserAgent(or(
    kubernetesClientProperties.getUserAgent(),
    or(base.getUserAgent(), KubernetesClientProperties.DEFAULT_USER_AGENT))
)
.build();
```

is more readable, but I have no strong opinion here.

wind57: that would not work, needs slight adjustments: 

```
String userAgent = or(base.getUserAgent(), KubernetesClientProperties.DEFAULT_USER_AGENT);
if (StringUtils.hasText(kubernetesClientProperties.getUserAgent())) {
	userAgent = kubernetesClientProperties.getUserAgent();
}
```

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/KubernetesClientProperties.java
```diff
@@ -87,6 +92,8 @@
 
 	private String serviceAccountNamespacePath = SERVICE_ACCOUNT_NAMESPACE_PATH;
 
+	private String userAgent;
```
ryanjbaxter, line 95: We are still not on the same page ;)

I think this should be set to `DEFAULT_USER_AGENT` which then changes the logic above.

ryanjbaxter (approved):

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/KubernetesClientProperties.java
wind57: sorry :( a bit late, but I got your point. thank you for your patience

ryanjbaxter: No worries at all.  I am going to hold off on merging it now because we are cutting a release tomorrow and since this is not critical I don't want to introduce anything else at the moment.

Issue #900, opened 2021-11-03
Mockserver dependency must be upgraded


pandrez:
As the fabric8 kubernetes client version is now `5.5.0`, the `io.fabric8:mockwebserver` dependency must also be upgraded from `0.1.2`to `0.1.8` [link to the maven repository](https://mvnrepository.com/artifact/io.fabric8/kubernetes-server-mock/5.5.0)

Without this upgrade, there are some classes missing (for example, `io.fabric8.mockwebserver.crud.AttributeType`) which makes some requests  to the mockserver fail with a `NoClassDefFoundError`

wind57:
@ryanjbaxter  I was looking at his one today a bit and before I even do anything here, I would to ask a small question. 

We had [this PR, that upgraded to 5.5.0](https://github.com/spring-cloud/spring-cloud-kubernetes/pull/836/files). It is present in `main` branch, but not in `2.0.x`.

I do not really have a good understanding how releases are made and of what branches... Can you please comment here? 

Moreover, there are many things in `2.0.x`, but not in `main` and also in reverse. Thank you.

ryanjbaxter:
The 2.0.x branch is for Spring Cloud 2020.0.x release train
The main branch is for Spring Cloud 2021.0.x release train

We try to merge what we can into the 2.0.x branch and then merge it forward.  So everything in 2.0.x should also be in main.

There may be things in main that are not in 2.0.x because there are new features we are part of the 2021.0.x release.

wind57:
thank you @ryanjbaxter, that makes sense.

wind57:
@pandrez while we fixed this one, can you explain against what release where you testing and what where your steps? This is only possible to do from our `main` branch, so exactly what release where you using then? thank you

pandrez:
@wind57 I was testing with Spring Cloud 2021.0.x release train, which will be the first one to use fabric8 5.x version!
Like you said, this is only needed on the main branch.

Thank you for your support!

wind57:
@pandrez thank you for the feedback. To be fair, I was more interested in this part of my questions: "what where your steps?". I am still wondering what where you doing, initially to hit this issue. Where you writing your own tests that relied on transitive test dependencies? If so, I am still wondering how they ended up on your classpath...

What I really hoped for is a very simple and minimal example.

PR #904, opened 2021-11-05
Fix issue 900

wind57:


spring-cloud-kubernetes-dependencies/pom.xml
```diff
@@ -35,10 +35,9 @@
 		<arquillian.version>1.6.0.Final</arquillian.version>
 		<arquillian-cube.version>1.15.2</arquillian-cube.version>
 		<hoverfly.version>0.13.0</hoverfly.version>
-		<kubernetes-client.version>5.5.0</kubernetes-client.version>
-		<kubernetes-java-client.version>13.0.0</kubernetes-java-client.version>
```
wind57, line 39: I have done two things here:

- give somehow more descriptive names for the properties: `kubernetes-fabric8-client.version` and `kubernetes-native-client.version`. 
- upgrade fabric8 to `5.9.0` from `5.5.0`

spring-cloud-kubernetes-dependencies/pom.xml
```diff
@@ -35,10 +35,9 @@
...
+		<kubernetes-fabric8-client.version>5.9.0</kubernetes-fabric8-client.version>
+		<kubernetes-native-client.version>13.0.0</kubernetes-native-client.version>
 		<istio-client.version>1.7.7.1</istio-client.version>
-		<mockwebserver.version>0.1.2</mockwebserver.version>
```
wind57, line 41: this dependency was never supposed to be here and I will explain why below.

The very first dependency we have in `spring-cloud-kubernetes-dependencies` is: 

```
<dependencyManagement>
		<dependencies>
			<dependency>
				<groupId>io.fabric8</groupId>
				<artifactId>kubernetes-client-bom</artifactId>
				<version>${kubernetes-fabric8-client.version}</version>
				<type>pom</type>
				<scope>import</scope>
			</dependency>
```

which is great, it means we rely on the on the bom that fabric8 provides. 

This bom has `io:fabric8:kubernetes-server-mock`, which internally provides: `io.fabric8:mockwebserver.version`, which means we should have never track this individually. I do not know the reasons it was added, but it surely does not matter at this point in time. We should not track that.
I also did a : 

```
bash-3.2$ read  -d ' ' x < <(grep -r "<artifactId>mockwebserver</artifactId>" .);echo "$x"
./spring-cloud-kubernetes-dependencies/pom.xml:                         <artifactId>mockwebserver</artifactId>
./spring-cloud-kubernetes-fabric8-config/pom.xml:                       <artifactId>mockwebserver</artifactId> 

``` 

To see where we have it as a dependency also. Since it was present in `spring-cloud-kubernetes-fabric8-config` (but not needed), I removed it from there also.

spring-cloud-kubernetes-fabric8-discovery/src/test/java/org/springframework/cloud/kubernetes/fabric8/discovery/KubernetesDiscoveryClientTest.java
```diff
@@ -16,43 +16,39 @@
```
wind57, line 16: This test was failing after I did the above change. The most sane thing to do, imho, is to fix it with the way we have already refactored tests in configmap/secrets implementation. In the long run (once I get here), this entire package should be refactored to junit-5. But in order to fix this current issue, this is OK.

wind57: It would take a while, but if you look closely, I have not changed the asserts, at all, to preserve the tests logic.

spring-cloud-kubernetes-integration-tests/load-balancer/src/test/java/org/springframework/cloud/kubernetes/fabric8/loadbalancer/LoadBalancerAllNamespacesTests.java
```diff
@@ -16,7 +16,9 @@
 
 package org.springframework.cloud.kubernetes.fabric8.loadbalancer;
```
wind57, line 18: this test was failing too, I had to sightly fix it.

wind57:
PR ready to review, will close [this defect](https://github.com/spring-cloud/spring-cloud-kubernetes/issues/900)

spring-cloud-kubernetes-dependencies/pom.xml
```diff
@@ -246,31 +245,10 @@
 				<scope>test</scope>
 			</dependency>
 
-			<dependency>
-				<groupId>io.fabric8</groupId>
-				<artifactId>mockwebserver</artifactId>
-				<version>${mockwebserver.version}</version>
-				<scope>test</scope>
-				<exclusions>
-					<exclusion>
```
ryanjbaxter, line 255: Could this have been included due to the exclusions?

wind57: may at the time it was messing some tests, but not currently. I would have given much more thought time to this (and probably even try to figure out transitive dependencies everywhere), but this has `<scope>test</scope>`, which imho, simplifies the explanation here greatly. wdyt?

ryanjbaxter: fair enough, I will merge this one, and the other ones on Monday

ryanjbaxter (approved):

spring-cloud-kubernetes-dependencies/pom.xml
wind57: ok, thank you.

PR #905, opened 2021-11-05
Fix fabric8 client dependency (related to issue 900)

wind57:


wind57:
This PR should go after [this one](https://github.com/spring-cloud/spring-cloud-kubernetes/pull/904) and then changes will make a lot more sense. I did not want to muddy the waters in that PR, that fixes an actual problem... When that one is merged, I'll comment on this one more if needed.

But basically the idea is the same as there, we track `kubernetes-client` dependency in our `spring-cloud-kubernetes-dependencies`, while it is provided by bom. And we do that for `<scope>test</scope>` and `<type>test-jar</type>`, which makes it even more compelling to remove it from there.

spring-cloud-kubernetes-dependencies/pom.xml
```diff
@@ -246,45 +245,6 @@
...
-						<artifactId>jackson-core</artifactId>
-					</exclusion>
-					<exclusion>
-						<artifactId>tools</artifactId>
-						<groupId>com.sun</groupId>
-					</exclusion>
-				</exclusions>
-			</dependency>
-
-			<dependency>
```
wind57, line 270: this is the actual change in this PR, we should drop `kubernetes-client` explicit dependency, it is managed by bom anyway

ryanjbaxter:
Seems like some of these changes duplicate whats in https://github.com/spring-cloud/spring-cloud-kubernetes/pull/904/

wind57:
right, I branched from that one - because otherwise the build would have failed. That is why I said that if we agree on that one and merge it, and _then_ look at this one, it will make more sense (the overlapping changes would not be present)

spring-cloud-kubernetes-dependencies/pom.xml
```diff
@@ -245,24 +245,6 @@
 				<scope>test</scope>
 			</dependency>
 
-			<dependency>
```
wind57, line 248: I think that now it makes a lot more sense - just like the mock server removal, we remove client also. It has the same `<scope>test</scope>`

spring-cloud-kubernetes-fabric8-autoconfig/pom.xml
```diff
@@ -84,6 +84,13 @@
 		<dependency>
 			<groupId>org.springframework.boot</groupId>
 			<artifactId>spring-boot-starter-test</artifactId>
+			<!-- in favor of mockito-inline -->
```
wind57, line 87: some tests were failing, so need mockito-inline.

wind57:
@ryanjbaxter it should make a lot more sense now

Issue #861, opened 2021-08-29
k8s-client might read an incorrect secret from a different namespace 


wind57:
This is more of a question, then a bug; but it's a bit of both...

I am looking at [this line](https://github.com/spring-cloud/spring-cloud-kubernetes/blob/main/spring-cloud-kubernetes-client-config/src/main/java/org/springframework/cloud/kubernetes/client/config/KubernetesClientSecretsPropertySource.java#L56). Notice the logic here: 

- if there is a `namespace` defined, use it via `listNamespacedSecret`. 

- if there is not, read _all_ namespaces via `listSecretForAllNamespaces`.

This is an interesting choice. We can obviously read the wrong thing easily and expose data from a secret that users never intended for (this is also easily reproducible).

On the other hand, if we take a step back, we only call `listSecretForAllNamespaces` because we think the `namespace` might be `null`. The future of this bug and discussion should really happen _after_ [this one](https://github.com/spring-cloud/spring-cloud-kubernetes/issues/859).

For the time being, I am leaving it here as a place-holder and will get back to it.

ryanjbaxter:
I think the reason why I wrote it that way was because if we didn't know the namespace we were out of luck.  I suppose we could throw and error at that point.  You are right it could mean we expose something we shouldn't be.

wind57:
You're right and I totally understand that, but it's tricky. Let's take each at a time. 

- For the `fabric8` client, we think we have a namespace, always, because we default as a last resort to `KubernetesClient::getNamespace` and hey, the client is really suppose to know where it runs, right? Well, [not really](https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/Config.java#L742). 

Basically `fabric8` says : I am going to try a read a few properties and env variables, if not set, I will try to read `/var/run/secrets/kubernetes.io/serviceaccount/namespace`. Not present? Oh well... 

- k8s-client does not have such a method to find out the namespace from the current client. I [have asked about this](https://github.com/kubernetes-client/java/issues/1857), no response yet. Judging by the python client [issue](https://github.com/kubernetes-client/python/issues/363), I am 99% of the answer I am going to get.

[...]

PR #906, opened 2021-11-08
namespace resolution must work the same across callers

wind57:


spring-cloud-kubernetes-fabric8-config/src/main/java/org/springframework/cloud/kubernetes/fabric8/config/Fabric8ConfigUtils.java
```diff
@@ -102,10 +110,6 @@ static String getApplicationNamespace(KubernetesClient client, String namespace,
 
 	}
 
-	public static String getApplicationNamespace(KubernetesClient client, String namespace) {
```
wind57, line 105: move this method a bit up, so that the two deprecated methods stay close to each other.

spring-cloud-kubernetes-fabric8-config/src/main/java/org/springframework/cloud/kubernetes/fabric8/config/Fabric8ConfigMapPropertySource.java
```diff
@@ -54,14 +54,15 @@ public Fabric8ConfigMapPropertySource(KubernetesClient client, String name) {
 	@Deprecated
 	public Fabric8ConfigMapPropertySource(KubernetesClient client, String name, String namespace,
 			Environment environment) {
-		super(getName(name, getApplicationNamespace(client, namespace)),
-				getData(client, name, getApplicationNamespace(client, namespace), environment, "", true));
+		super(getName(name, getApplicationNamespace(client, namespace, "Config Map", null)), getData(client, name,
+				getApplicationNamespace(client, namespace, "Config Map", null), environment, "", true));
```
wind57, line 58: This is a little bit complicated to follow, but I will try to explain. 

 - We have a `public` constructor for `Fabric8ConfigMapPropertySource`, that takes a `namespace` as parameter, among other properties, which I do not care about here.
 - This constructor is called in `Fabric8ConfigMapPropertySourceLocator::getMapPropertySource`. This method internally calls `Fabric8ConfigUtils::getApplicationNamespace`; a method that will throw a `NamespaceResolutionFailedException` in case a namespace can not be found. 
 - So far nothing new.

Now, because `Fabric8ConfigMapPropertySource` is a `public` class that has a `public` constructor, someone can still call this one directly, of course. Callers, as such, can pass the namespace directly. What we used to do is call : 

```
	public static String getApplicationNamespace(KubernetesClient client, String namespace) {
		return !StringUtils.hasLength(namespace) ? client.getNamespace() : namespace;
	}
```

Get the namespace provided or from the client. So in one case we used to throw a `NamespaceResolutionFailedException`, but in the other one, we would end up querying all namespaces (see `Fabric8ConfigUtils::getConfigMapData`), in case the namespace was not present. 

---

Now we will call : `getApplicationNamespace(client, namespace, "Config Map", null)`, effectively doing the same logical thing:

- if `namespace` was provided, use that
- if it was not provided get it from the client
- notice that instead of `KubernetesNamespaceProvider`, we pass a `null`, in order to preserve the previous logic. 

The trick here  is that we would ned up throwing a `NamespaceResolutionFailedException` if a namespace is not present, instead of querying all namespaces (like we used to do). This, imho, also closes a bug, since we should not really query all namespaces (we have the same open defect for Secrets, which I will handle too).

spring-cloud-kubernetes-client-config/src/main/java/org/springframework/cloud/kubernetes/client/config/KubernetesClientConfigMapPropertySource.java
```diff
@@ -42,13 +44,15 @@
 	@Deprecated
 	public KubernetesClientConfigMapPropertySource(CoreV1Api coreV1Api, String name, String namespace,
 			Environment environment) {
-		super(getName(name, namespace), getData(coreV1Api, name, namespace, environment, "", true));
+		super(getName(name, getApplicationNamespace(namespace, "Config Map", null)), getData(coreV1Api,
```
wind57, line 47: same as for fabric8 - if someone uses this directly, they must provide the `namespace`, otherwise this will fail

spring-cloud-kubernetes-fabric8-config/src/main/java/org/springframework/cloud/kubernetes/fabric8/config/Fabric8ConfigMapPropertySource.java
ryanjbaxter: Can you point to the issue?

wind57: there is no issue per-se, and can only happen if someone extends `Fabric8ConfigMapPropertySource` or the kubernetes-native property source, and do not provide a namespace. 

This PR makes such clients (as above) fail with the same `NamespaceResolutionFailedException` as we do in the usual flow. So : 

- plain users of our library + we can't resolve a namespace =>  `NamespaceResolutionFailedException`
- some clients that extends `Fabric8ConfigMapPropertySource/KubernetesClientConfigMapPropertySource` (for whatever reason) + we can't resolve a namespace => 

     a) in fabric8 we query all namespace to find such a configmap (not a great idea)
     b) in k8s-native, we would fail with a `ApiException`

This PR makes everything fail with the same `NamespaceResolutionFailedException`.

ryanjbaxter: OK this sentence made me think there was an issue open for this
>This, imho, also closes a bug, since we should not really query all namespaces (we have the same open defect for Secrets, which I will handle too).

To be honest I think that this change in behavior should only be made in main.

wind57: snap! I thought I made it against `main`, as I agree with you. I will point it to `main`.

wind57: now it's against `main`

ryanjbaxter (approved):

ryanjbaxter:
I think it would be worth including the secret changes in the same PR.

wind57:
I keep fighting with myself over at what point I should introduce the same changes for config maps (like the one in this PR) and for secrets too. 

The reason I am saying this is that I wanted to go with the route of : "close all pending defects and enchantments in config maps" and then move to secrets. 

I absolutely want to back-port this change to secrets also, but there are some other things I would like to fix also, before this one. If you are OK, I'd like to keep it the way it is and I promise that I will do the same change for secrets, when I am ready to do it.

ryanjbaxter:
My concern is that we have out GA release on Nov 30th so I would want the behavior between both to be the same before we do the release.

spring-cloud-kubernetes-fabric8-config/src/test/java/org/springframework/cloud/kubernetes/fabric8/config/Fabric8SecretsPropertySourceTest.java
```diff
@@ -41,23 +37,18 @@
...
 @SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT, classes = App.class,
 		properties = "spring.main.cloud-platform=KUBERNETES")
 @TestPropertySource("classpath:/application-secrets.properties")
 @EnableKubernetesMockClient(crud = true, https = false)
-public class Fabric8SecretsPropertySourceTest {
+class Fabric8SecretsPropertySourceTest {
 
 	private static final String NAMESPACE = "test";
 
 	private static KubernetesClient mockClient;
```
wind57, line 51: I want to remove `@ClassRule` and because of that this test needs slight changes, because of the way `@EnableKubernetesMockClient` works. This does not change the logic of the test at all. It does two things:

- removes `@ClassRule` and as such slightly changes
- moves two tests into `Fabric8SecretsPropertySourceMockTests` (where I will add more tests soon)

spring-cloud-kubernetes-fabric8-config/src/main/java/org/springframework/cloud/kubernetes/fabric8/config/Fabric8SecretsPropertySource.java
```diff
@@ -40,40 +42,40 @@
...
-		super(getSourceName(name, namespace), getSourceData(client, name, namespace, labels, failFast));
+		super(getSourceName(name, getApplicationNamespace(client, namespace, "Secret", null)),
+			getSourceData(client, name, getApplicationNamespace(client, namespace, "Secret", null), labels, failFast));
 	}
 
 	private static Map<String, Object> getSourceData(KubernetesClient client, String name, String namespace,
 			Map<String, String> labels, boolean failFast) {
 		Map<String, Object> result = new HashMap<>();
 
-		String namespaceToUse = StringUtils.hasLength(namespace) ? namespace : client.getNamespace();
```
wind57, line 50: we do not need this line anymore, as it already has happened earlier. If you have reached this point a namespace is already known

spring-cloud-kubernetes-client-config/src/main/java/org/springframework/cloud/kubernetes/client/config/KubernetesClientSecretsPropertySourceLocator.java
```diff
@@ -90,8 +90,8 @@ else if (kubernetesClientProperties != null) {
 					kubernetesNamespaceProvider);
 		}
 
-		return new KubernetesClientSecretsPropertySource(coreV1Api, secretName, namespace, environment,
-				normalizedSource.getLabels(), this.properties.isFailFast());
```
wind57, line 94: environment is not needed - so I removed it, this in turn triggered some changes in some other places; mainly in test classes

spring-cloud-kubernetes-client-config/src/main/java/org/springframework/cloud/kubernetes/client/config/KubernetesClientSecretsPropertySource.java
```diff
@@ -97,20 +82,13 @@ public KubernetesClientSecretsPropertySource(CoreV1Api coreV1Api, String name, S
 	}
 
 	private static String createLabelsSelector(Map<String, String> labels) {
-		StringBuilder selectorString = new StringBuilder();
```
wind57, line 100: simplify expression

spring-cloud-kubernetes-client-config/src/main/java/org/springframework/cloud/kubernetes/client/config/KubernetesClientSecretsPropertySource.java
```diff
@@ -97,20 +82,13 @@ public KubernetesClientSecretsPropertySource(CoreV1Api coreV1Api, String name, S
...
-			selectorString.append(key + "=" + labels.get(key));
-		}
-		return selectorString.toString();
+		return labels.entrySet().stream().map(e -> e.getKey() + "=" + e.getValue()).collect(Collectors.joining(","));
 	}
 
 	private static void putAll(V1Secret secret, Map<String, Object> result) {
 		Map<String, String> secretData = new HashMap<>();
-		secret.getData().forEach((key, value) -> secretData.put(key, Base64.getEncoder().encodeToString(value)));
-		if (secret != null) {
```
wind57, line 113: simplify expression here also

spring-cloud-kubernetes-client-config/src/main/java/org/springframework/cloud/kubernetes/client/config/KubernetesClientSecretsPropertySource.java
```diff
@@ -38,49 +40,32 @@
...
 			Map<String, String> labels, boolean failFast) {
 		Map<String, Object> result = new HashMap<>();
 
 		LOG.info("Loading Secret with name '" + name + "' or with labels [" + labels + "] in namespace '" + namespace
 				+ "'");
 		try {
-			// Read for secrets api (named)
 			if (StringUtils.hasText(name)) {
 				Optional<V1Secret> secret;
-				if (!StringUtils.hasText(namespace)) {
```
wind57, line 59: since we know the namespace for sure here, we can get rid of calls `api.listSecretForAllNamespaces...` This will also close [this defect](https://github.com/spring-cloud/spring-cloud-kubernetes/issues/861)

wind57:
I agree with your argument and added secrets also.

PR #907, opened 2021-11-10
Change the log level to prevent false-positives

osvso:
The `WARN` log level is being abused here which can flood the production logs (usually you are not interested in any lower log levels there). It is totally fine to not define a separate configMap for each profile that is active in the application therefore it shouldn't be a warning. The same applies to any other form of externalized configuration.

### DIFF
--- a/spring-cloud-kubernetes-fabric8-config/src/main/java/org/springframework/cloud/kubernetes/fabric8/config/Fabric8ConfigUtils.java
+++ b/spring-cloud-kubernetes-fabric8-config/src/main/java/org/springframework/cloud/kubernetes/fabric8/config/Fabric8ConfigUtils.java
@@ -111,7 +111,7 @@ public final class Fabric8ConfigUtils {
 				: client.configMaps().inNamespace(namespace).withName(name).get();
 
 		if (configMap == null) {
-			LOG.warn("config-map with name : '" + name + "' not present in namespace : '" + namespace + "'");
+			LOG.info("config-map with name : '" + name + "' not present in namespace : '" + namespace + "'");
 			return Collections.emptyMap();
 		}
 


spring-cloud-kubernetes-fabric8-config/src/main/java/org/springframework/cloud/kubernetes/fabric8/config/Fabric8ConfigUtils.java
```diff
@@ -111,7 +111,7 @@ public static String getApplicationNamespace(KubernetesClient client, String nam
 				: client.configMaps().inNamespace(namespace).withName(name).get();
 
 		if (configMap == null) {
-			LOG.warn("config-map with name : '" + name + "' not present in namespace : '" + namespace + "'");
+			LOG.info("config-map with name : '" + name + "' not present in namespace : '" + namespace + "'");
```
wind57, line 114: So your production logs are on `WARN`? 

To be honest, I don't think we are "abusing" the log level here, but since I wrote this, I might be biased.

As a side question why don't you lower the level of just this package if you do not want them? Am I missing something?

osvso: @wind57 Thanks for your fast reply!
Yes, that's the default behaviour. Why would like to spam your logs with information that is not really useful? If there are any problems observed of course it is fine to change the log levels to get more data. In other cases, it seems to be just a waste of resources (not only making the app itself slower but, also hammering your logging/observability infrastructure).
Another point to make here is the fact that this particular scenario is totally expected at correctly configured deployments therefore it should not be signalled as a warning. That's why I've said this implementation is abusing the discussed log level.

Lastly, I don't think that silencing logs for a class playing an important role is a way forward. I wouldn't propose this change otherwise 😉

wind57: > Why would like to spam your logs with information that is not really useful?

We are looking at this from different angles. You are looking at your app only, we are looking at a lot more. Profiles are baked into spring, so we expect to be used heavily, in general. 

> If there are any problems observed of course it is fine to change the log levels to get more data

We seem to agree here, btw and that was my initial point. Why don't you change the log level to `ERROR` then here? No log statements - you are good. 

> Another point to make here is the fact that this particular scenario is totally expected at correctly configured deployments therefore it should not be signalled as a warning. That's why I've said this implementation is abusing the discussed log level

Right, this is where I agree with you, somehow. For some people, a warn here, makes total sense, for some (like your example), it does not. There is no way to please both sides. Your point is understood, but someone will come tomorrow and report that their logs are lost - where they want them... It's a thin line.

> not only making the app itself slower but, also hammering your logging/observability infrastructure

These log statements are only present at start-up, and the impact is in the order of `ms`. I doubt this will ever have any measurable impact, compared to the heavy loading we are doing at start-up. Also, if you use any tracing apps, like elk, disabling not to see such log statements is totally an option.

> Lastly, I don't think that silencing logs for a class playing an important role is a way forward

I did not propose to silence them, but to reduce to `ERROR`.

----

Anyway, we are discussing a very minor aspect here and I do not disagree with you, completely, I was just providing my reasonings too. I will let @ryanjbaxter provide his input, but if you really care about this log statements, you need to provide a proper unit test here. Take a look into `org.springframework.cloud.kubernetes.fabric8.config.MissingActuatorTest` that uses `OutputCaptureExtension`. I was envisioning something close to that test.

osvso: Thanks, yes, I believe we are looking at this particular implementation detail from different perspectives. My main argument here is that warnings should be used in situations where there is indeed something wrong or not really expected happening in the system, like for example: 
- use of unsafe/marked to be removed method
- detected misconfiguration that doesn't fail the application
- unhappy but, still expected flow was executed
- etc.

I don't think it is particularly related to the services I work on but, overall good practice. In this case, we are using `WARN` to indicate a situation that occurs in most of the deployments by default in a correctly configured system. Going further, there arent' any guidelines or good practices that would outline the need of defining a configMap for every active profile.

ryanjbaxter: I can see an argument for both sides, info or warn.  If the log statement was being executed repeatedly I could see reason to maybe move it to info, but since its only happening once (in most cases) I am not sure I see it clogging up logs or having a huge impact on performance.  My bias is to leave it as is for now.

osvso: @ryanjbaxter What we observe is that the logs are literally polluted with this warning. The configMaps can change during the application lifecycle without any redeployments taking place. Also, Spring has a built-in mechanism to refresh the property sources and rebuild the affected dependency tree graph. That's the main functionality with triggers this warning to be repeatedly logged in a single application lifecycle. This is a real problem.

ryanjbaxter: How frequently is the refresh occurring?

osvso: I see it being logged 20times every 30 minutes.

osvso: The app is using two configMaps and one active profile, which means it is printed redundantly 10 times for each configMap.

osvso: Of course the more configMaps and acrtive profiles are being used by the service the more flooded the logs become.

ryanjbaxter: OK but I still am not sure why the refresh endpoint is being called every 30 minutes.

osvso: I would need to dive a bit deeper but, it seems there are two main implementations that are responsible for triggering this flow:
1. EventBased[ConfigMap/Secrets]ChangeDetector
2. Pollgin[ConfigMap/Secret]ChangeDetector
both located in the `reload` package of the spring.cloud.kubernetes.config. The polling uses a fixed interval while the event-based watches the deployment state using the Kubernetes client. 
Still, I believe it is expected to be called multiple times, regardless of the threshold and what we discuss is just an outcome or side-effect of the way it is implemented right now. If, as an app developer, we are interested to get to know that there was no configMap defined for the active profile I think it is fine to keep this log. At the same time, it shouldn't be using the `warn` level as it is simply invalid and unexpected.

osvso: If we accept the way it works it will mean that it is fine and expected to see warnings from every property source as soon as it detects that there is no profile-specific configuration provided.

osvso: I also believe that it is required to look at the [`Fabric8ConfigMapPropertySource`](https://github.com/spring-cloud/spring-cloud-kubernetes/blob/main/spring-cloud-kubernetes-fabric8-config/src/main/java/org/springframework/cloud/kubernetes/fabric8/config/Fabric8ConfigMapPropertySource.java#L74) which is the only caller of that particular method.

ryanjbaxter: But the event and polling refresh mechanism should only be creating a new property source if there was actually a change, so I find it odd that unless the config map is changing frequently that it would be called so frequently.

Regardless though, I am find with the info level, at the end of the day having the information present is what matters.  Unless @wind57 has a strong feeling that it shouldn't be changed its ok with me.

wind57: nope, I do not at all. I totally agree that as long as the info is there, I am good! It's not even mandatory to have a unit test. thank you @osvso for taking the time to respond and provide your feedback here.

wind57 (approved):

spring-cloud-kubernetes-fabric8-config/src/main/java/org/springframework/cloud/kubernetes/fabric8/config/Fabric8ConfigUtils.java
osvso: Thanks guys! Appreciate the discussion we had 👍 I will look whether I can get some tests in tomorrow and will get back to you.

ryanjbaxter (changes requested):
Would actually ie this to be submitted agains the 2.0.x branch

osvso:
> Would actually ie this to be submitted agains the 2.0.x branch

@ryanjbaxter Ok, I will open another PR as I've based this one on the `main` branch. 
https://github.com/spring-cloud/spring-cloud-kubernetes/pull/911

ryanjbaxter:
OK I merged #911 and merged it into main

PR #908, opened 2021-11-11
Fix 893

wind57:


spring-cloud-kubernetes-client-config/src/test/java/org/springframework/cloud/kubernetes/client/config/KubernetesClientBootstrapConfigurationTests.java
```diff
@@ -158,7 +157,7 @@ public void bothPresent() {
 	// tests that @ConditionalOnCloudPlatform(CloudPlatform.KUBERNETES) has the desired
 	// effect, meaning when it is disabled, no property source bean is present
 	@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT, classes = Application.class,
-			properties = { "kubernetes.informer.enabled=false" })
+			properties = { "kubernetes.manifests.enabled=false" })
```
wind57, line 160: I will only pin point the important (somehow) changes in this PR. This test was failing, so had to change this setting

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/ConditionalOnKubernetesEnabled.java
```diff
@@ -1,41 +0,0 @@
-/*
- * Copyright 2019-2020 the original author or authors.
```
wind57, line 2: this was removed

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/profile/AbstractKubernetesProfileEnvironmentPostProcessor.java
```diff
@@ -58,12 +58,17 @@ public void postProcessEnvironment(ConfigurableEnvironment environment, SpringAp
 
 		application.addInitializers(ctx -> LOG.replayTo(AbstractKubernetesProfileEnvironmentPostProcessor.class));
 
-		boolean kubernetesEnabled = environment.getProperty("spring.cloud.kubernetes.enabled", Boolean.class, true);
-		if (!kubernetesEnabled) {
-			return;
+		boolean serviceHost = StringUtils
```
wind57, line 61: this is another important change, we need to replicate what `@ConditionalOnCloudPlatform(KUBERNETES)` is doing

README.adoc
```diff
@@ -1096,11 +1096,16 @@ as the code of the project relies on the
 https://github.com/fabric8io/kubernetes-client[Fabric8 Kubernetes Java client], which is a fluent DSL that can
 communicate by using `http` protocol to the REST API of the Kubernetes Server.
 
-To disable the integration with Kubernetes you can set `spring.cloud.kubernetes.enabled` to `false`. Please be aware that when `spring-cloud-kubernetes-config` is on the classpath,
```
wind57, line 1099: what we use to say : 

1a) if you wan to disable : use `spring.cloud.kubernetes.enabled=false`
2a) if you have config on the classpath, put `spring.cloud.kubernetes.enabled` into bootstrap
3a) if you want to disable `EnvironmentPostProcessor`, you need to do set `-DSPRING_CLOUD_KUBERNETES_ENABLED=false` 

What we say now:

1b) awareness is based on `ConditionalOnCloudPlatform`. It is possible to override
this via `spring.main.cloud-platform`. In this way we preserve (1a): You want to _disable_ integration? use `spring.main.cloud-platform`
2b) same as (1b) pretty much
3b) we do not mention  `-DSPRING_CLOUD_KUBERNETES_ENABLED=false`, because now, by default, it is _not_ enabled; and it was the case in (3a). As a matter of fact, the only reason we put this in the documentation was because it was enabled by default and users had to disable it. This paradigm shifts now in the reverse order. I mean how many people will have env variables `KUBERNETES_SERVICE_HOST` and `KUBERNETES_SERVICE_PORT` set? On the other hand, _everyone_ started with `spring.cloud.kubernetes.enabled=true`, by default, because of that `matchIfMissing=true`.

docs/src/main/asciidoc/kubernetes-awareness.adoc
```diff
@@ -1,18 +1,29 @@
...
+that setting via `spring.main.cloud-platform`.
+
+For example, if you need to test some features, but do not want to deploy to a cluster, it is enough to set the:
+`spring.main.cloud-platform=KUBERNETES`. This will make `spring-cloud-kubernetes` act as-if it is deployed in a real cluster.
+Be aware that when `spring-cloud-kubernetes-config` is on the classpath, `spring.main.cloud-platform` should be set in `bootstrap.{properties|yml}`
+(or the profile specific one),  otherwise it should be in `application.{properties|yml}` (or the profile specific one).
+Also note that these properties: `spring.cloud.kubernetes.config.enabled` and `spring.cloud.kubernetes.secrets.enabled`
+only take effect when set in `bootstrap.{properties|yml}`.
+
+=== Functional changes since 2.1.x version
```
wind57, line 22: a light note about the changes, mainly here because I bet we will need to reference it in the future questions/issues that will come, here or on stackoverflow or any other channel.

wind57:
just to make sure: 


```
grep -r "@ConditionalOnKubernetes\$" -i . | wc -l # print 0
```

```
grep -r "spring.cloud.kubernetes.enabled\$" -i . | wc -l # prints 0
```

docs/src/main/asciidoc/_configprops.adoc
```diff
@@ -73,7 +73,7 @@
 |spring.cloud.kubernetes.discovery.service-labels |  | If set, then only the services matching these labels will be fetched from the Kubernetes API server.
 |spring.cloud.kubernetes.discovery.service-name | `unknown` | The service name of the local instance.
 |spring.cloud.kubernetes.discovery.wait-cache-ready | `true` | 
-|spring.cloud.kubernetes.enabled | `true` | Whether to enable Kubernetes integration.
+|spring.main.cloud-platform" | `KUBERNETES` | Whether to enable Kubernetes integration (this will force it and is not recommended)
```
ryanjbaxter, line 76: I dont think we should document this property here, its documented as part of Boot

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/profile/AbstractKubernetesProfileEnvironmentPostProcessor.java
ryanjbaxter: Why?

wind57: this is an environment post processor here. We used to have : 

```
boolean kubernetesEnabled = environment.getProperty("spring.cloud.kubernetes.enabled", Boolean.class, true);
```

because there is simply no other way to check if kubernetes was enabled or not, as this is an environment post processor declared in `spring.factories`. 

We need to have the same logic here, but for `@ConditionalOnCloudPlatform(KUBERNETES)`.

docs/src/main/asciidoc/_configprops.adoc
wind57: ok, I will remove this

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/profile/AbstractKubernetesProfileEnvironmentPostProcessor.java
ryanjbaxter: What if you check spring.main.platform?

wind57: what if users don't specify that, but they are still running under kubernetes? meaning `KUBERNETES_SERVICE_HOST` and `KUBERNETES_SERVICE_PORT` are present? Wouldn't we miss adding `kubernetes` profile?

ryanjbaxter: I am assuming Boot would set the value correctly based on the platform, is that not the case?

wind57: does not look like it, see [here](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/cloud/CloudPlatform.java#L161..L164).

ryanjbaxter: Could we use the CloudPlatform class instead or replicating the logic?

wind57: OMG! I am an idiot, thx a lot.

ryanjbaxter: No worries :)

ryanjbaxter (approved):

PR #910, opened 2021-11-11
fix 909 (Drop @Deprecated things in some modules)

wind57:


wind57:
this simply removes `@Deprecated` stuff that I have touched and know about, some tests are changed also because of that

wind57:
not for myself, this drops deprecated things from : 

 - commons
 - fabric8 autoconfig/config
 - k8s-native autoconfig/config

PR #911, opened 2021-11-12
Change the log level to prevent false-positives

osvso:
The `WARN` log level is being abused here which can flood the production logs (usually you are not interested in any lower log levels there). It is totally fine to not define a separate configMap for each profile that is active in the application therefore it shouldn't be a warning. The same applies to any other form of externalized configuration.


osvso:
@wind57 @ryanjbaxter Good morning guys! As requested, I've opened another PR targeting 2.0.x branch.

ryanjbaxter (approved):

PR #915, opened 2021-11-14
Fix reading a wrong secret

wind57:


spring-cloud-kubernetes-fabric8-config/src/test/java/org/springframework/cloud/kubernetes/fabric8/config/secrets_with_labels/SecretsWithLabelsTests.java
```diff
@@ -0,0 +1,96 @@
...
+		createSecret("secret-three", three);
+
+	}
+
+	private static void createSecret(String name, Map<String, String> data) {
+		mockClient.secrets().inNamespace("spring-k8s")
+				.create(new SecretBuilder().withNewMetadata().withName(name).endMetadata().addToData(data).build());
+	}
+
+	/**
```
wind57, line 81: here is why this is a defect

spring-cloud-kubernetes-fabric8-config/src/main/java/org/springframework/cloud/kubernetes/fabric8/config/Fabric8SecretsPropertySource.java
```diff
@@ -63,8 +62,11 @@ public Fabric8SecretsPropertySource(KubernetesClient client, String name, String
 				putDataFromSecret(secret, result, namespace);
 			}
 
-			client.secrets().inNamespace(namespace).withLabels(labels).list().getItems()
-					.forEach(s -> putDataFromSecret(s, result, namespace));
+			if (labels != null && !labels.isEmpty()) {
```
wind57, line 65: while working on a diff PR, I had a test failure that kind of threw me off. After some debugging, I realized that fabric8 client is doing something not very correct, at least imo. [here is the explanation for it](https://github.com/spring-cloud/spring-cloud-kubernetes/pull/915/files#r748789603).

Basically, when doing something like: `client.secrets().inNamespace(namespace).withLabels(labels).list().getItems()` and passing an empty `labels` - this will match _all_ labels in that namespace. Without this fix, the unit test that I put in, fails. It fails because `secret-three` (that should have never been read anyway) overrides the value from `secret-two`. This is a defect. I am totally the one to be blamed for it, since I refactored the logic here in this manner. But boy, this was un-expected. I will raise a question in the fabric8 client also, I can't wrap my head around why this is the way it is, let's hope I miss something (I usually do)

wind57: [here](https://github.com/fabric8io/kubernetes-client/issues/3590) is the fabric8 link to the question.

wind57:
Not for myself, if [this](https://github.com/spring-cloud/spring-cloud-kubernetes/pull/917) is accepted and goes in, this PR is not needed. The only thing that should go from it, is the unit test.

spring-cloud-kubernetes-fabric8-config/src/main/java/org/springframework/cloud/kubernetes/fabric8/config/Fabric8SecretsPropertySource.java
ryanjbaxter: Maybe add a comment to the code explaining the behavior and the check

ryanjbaxter (approved):
The bug is only in main?  Or is it in 2.1.x as well?

wind57:
imho, we should not merge this; I should close it. The secrets implementation is really complicated to follow and has lots of issues. I was waiting when it would be OK to introduce breaking changes and clean it up, in a major way. 

I want to take that other PR I still have open and talk a lot more other there. I might need a test from here, so let's keep it open. I'll move this to a draft state, so that we do not merge it by accident.

wind57:
I am closing this, since in main, this is already fixed.

Issue #794, opened 2021-05-17
Keep consistent with SpringCloud default behavior


carrypann:
**Is your feature request related to a problem? Please describe.**
Bootstrap is no longer enabled by default after SpringCloud 2020.0+,  while SpringCloud Kubernetes still requires it by including `spring-cloud-starter-bootstrap`. Why not keep consistent with SpringCloud's default behavior?

SpringBoot 2.4+ brings the new syntax `spring.connfig.import` to import external config data, means that we don't need `bootstrap.yaml` any more.  Considering that `spring-cloud-consul-config` has been supported this syntax, so we removed `bootstrap.yaml`, and put `spring.application.name` to `application.yaml`, which results in `spring-cloud-kubernetes-config` not working.

ryanjbaxter:
It is needed because of the way Spring Cloud Kubernetes Config loads ConfigMaps and Secrets as property sources.  It needs to happen in the bootstrap phase.  We suggest though that you mount these as volumes and then leverage `spring.config.import` to add them as property sources instead of using the config module.

carrypann:
Thanks for your reply. I know that I can mount config files as volumes, but I want to specific difference config name to import, like `spring.cloud.kubernetes.config.sources[0].name=xxx`, however, `spring.config.import=configtree` does not support this feature.

Since `spring.config.import=configtree` can import external config without bootstrap context, Spring Cloud Kubernetes Config can also loads ConfigMaps and Secrets by providing an k8s implementation of `org.springframework.boot.context.config.ConfigDataLoader`, is this right?

ryanjbaxter:
This is kind of interesting, something I have not really thought about.

So what is missing with `spring.config.import` is the ability to look up files based on `spring.application.name`?

As a side note have you looked into using `spring.cloud.kubernetes.config.paths`?

spring-cloud-issues:
If you would like us to look at this issue, please provide the requested information. If the information is not provided within the next 7 days this issue will be closed.

spring-cloud-issues:
Closing due to lack of requested feedback. If you would like us to look at this issue, please provide the requested information and we will re-open the issue.

PR #917, opened 2021-11-15
Major refactor of configmaps and secrets implementation

wind57:


spring-cloud-kubernetes-fabric8-config/src/main/java/org/springframework/cloud/kubernetes/fabric8/config/Fabric8SecretsPropertySourceLocator.java
```diff
@@ -69,11 +68,10 @@ public Fabric8SecretsPropertySourceLocator(KubernetesClient client, SecretsConfi
 	@Override
 	protected MapPropertySource getPropertySource(ConfigurableEnvironment environment,
 			SecretsConfigProperties.NormalizedSource normalizedSource, String configurationTarget) {
-		String secretName = getApplicationName(environment, normalizedSource.getName(), configurationTarget);
```
wind57, line 72: this has move to the caller. It does not really matter here, since `getApplicationName` is in commons _anyway_; but for the fix of the issue this matters a lot.

spring-cloud-kubernetes-fabric8-config/src/main/java/org/springframework/cloud/kubernetes/fabric8/config/Fabric8SecretsPropertySourceLocator.java
```diff
@@ -69,11 +68,10 @@ public Fabric8SecretsPropertySourceLocator(KubernetesClient client, SecretsConfi
...
 		String secretNamespace = getApplicationNamespace(this.client, normalizedSource.getNamespace(),
 				configurationTarget, provider);
 		Map<String, String> labels = normalizedSource.getLabels();
-		return new Fabric8SecretsPropertySource(this.client, secretName, secretNamespace, labels,
+		return new Fabric8SecretsPropertySource(this.client, normalizedSource.getName(), secretNamespace, labels,
```
wind57, line 74: proper secret name is now computed in `determineSources`, instead of here

spring-cloud-kubernetes-client-config/src/main/java/org/springframework/cloud/kubernetes/client/config/KubernetesClientSecretsPropertySource.java
```diff
@@ -51,10 +51,9 @@ public KubernetesClientSecretsPropertySource(CoreV1Api coreV1Api, String name, S
 			Map<String, String> labels, boolean failFast) {
 		Map<String, Object> result = new HashMap<>();
 
-		LOG.info("Loading Secret with name '" + name + "' or with labels [" + labels + "] in namespace '" + namespace
```
wind57, line 54: slightly improve logging to get a better picture of what is going on

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/SecretsConfigProperties.java
```diff
@@ -158,13 +162,30 @@ public boolean isEmpty() {
...
+				Map<String, String> defaultLabels, Environment environment) {
+
+			Stream.Builder<NormalizedSource> normalizedSources = Stream.builder();
+
 			String normalizedName = StringUtils.hasLength(this.name) ? this.name : defaultName;
 			String normalizedNamespace = StringUtils.hasLength(this.namespace) ? this.namespace : defaultNamespace;
 			Map<String, String> normalizedLabels = this.labels.isEmpty() ? defaultLabels : this.labels;
 
-			return new SecretsConfigProperties.NormalizedSource(normalizedName, normalizedNamespace, normalizedLabels);
+			// if users do not specify a name for the secret (normalizedName), we still
```
wind57, line 174: this is the crox of the change/fix. 

What we used to have:

```
spring:
  cloud:
    kubernetes:
      secrets:
         namespace: spring-k8s
         name: some-secret
         labels:
            a: b 
```

With a configuration as above, we used to read : 

- a secret with name `some-secret` 
- all secrets with lables `a=b`. Notice that in this case, we did not care about the name of the secret. 

Same happens after this change, but this time, we have separated such a configuration in _two_ normalized sources (as opposed to only one before). This allows us to : 

- fix the actual issue of reading the a secret only once (you can look at the bug description for more details + I will add a few unit test to enforce this soon). This holds because we make a distinction now for sources that are "name based" and "labels based"; and if we do not care about `lables` in the case of "name based secrets" (and in reverse), we can always do:
 
```
// named-base source has Collections.emptyMap(), which denotes "we do not care about lables"
new NormalizedSource(secretName, normalizedNamespace, Collections.emptyMap());
```

and 

```
// lables-base source has null first argument, which denotes "we do not care about secret name"
new NormalizedSource(null, normalizedNamespace, labels);
```

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/SecretsConfigProperties.java
```diff
@@ -93,14 +97,13 @@ public String getConfigurationTarget() {
 	 * These are the actual name/namespace pairs that are used to create a
 	 * SecretsPropertySource
 	 */
-	public List<SecretsConfigProperties.NormalizedSource> determineSources() {
+	public List<NormalizedSource> determineSources(Environment environment) {
 		if (this.sources.isEmpty()) {
-			return Collections
-					.singletonList(new SecretsConfigProperties.NormalizedSource(SecretsConfigProperties.this.name,
-							SecretsConfigProperties.this.namespace, SecretsConfigProperties.this.labels));
+			return Collections.singletonList(new NormalizedSource(
```
wind57, line 102: `getApplicationName` has moved in here, just like in `normalize` method

spring-cloud-kubernetes-fabric8-config/src/main/java/org/springframework/cloud/kubernetes/fabric8/config/Fabric8SecretsPropertySource.java
```diff
@@ -49,22 +50,27 @@ public Fabric8SecretsPropertySource(KubernetesClient client, String name, String
...
-			Secret secret = client.secrets().inNamespace(namespace).withName(name).get();
-
-			// the API is documented that it might return null
-			if (secret == null) {
-				LOG.warn("secret with name : " + name + " in namespace : " + namespace + " not found");
-			}
-			else {
-				putDataFromSecret(secret, result, namespace);
+			// can come as null or empty, totally valid scenario
+			if (StringUtils.hasText(name)) {
```
ryanjbaxter, line 56: I think I would prefer we switch the logic here

if there are labels ignore the name and load the secrets that match the label
if there are no label check if there is a name (there always should be but worth a check) and load that if present

Then you could pass the secret name and it would still work as expected.

ryanjbaxter (changes requested):
Any reason why this can't be in 2.x?

spring-cloud-kubernetes-fabric8-config/src/main/java/org/springframework/cloud/kubernetes/fabric8/config/Fabric8SecretsPropertySource.java
wind57: I was actually torn apart, to either do what you suggest here, or what I did.

The problem with your suggestion is that it is a breaking change, someone that configured this with both a name and labels will get un-expected results, right?

ryanjbaxter: Can we have two separate constructors, one for name based and one for labels based?

wind57: If you are thinking to have that _in general_, then yeah, why not. If you are thinking to have that here, in this PR, then I'd say no. This will complicate things a lot as this was will require many changes. Especially if we want to implement [this filters](https://github.com/spring-cloud/spring-cloud-kubernetes/pull/818), the the idea of "sources" will become less relevant, as searching for things will be done in a different way.

wind57:
you're right, there is no reason for this not to go in `2.0.x`, but if I change the base now, I will bring 215 files in :) 

So when we agree on common ground here, I will close this one, open a new PR against `2.0.x` where I will cherry-pick commits from here.

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/SecretsConfigProperties.java
ryanjbaxter: Sorry for the confusion, I meant a constructor for NormalizedSource for the name case and the labels case.  I kind of dislike passing `null` and an empty `Map` to signify these different use cases.

wind57: every time I want to fix these kind of issues, this problem pops up again and again for me. I really want to follow your advise, and the changes I just added confirm that. But, as you can see - this is involved. Do I want to keep them the way I have them now? Of course I do - clear separation, easy to add new features, maintainable, etc. Will you agree with such breaking changes? Well, I think we both know the answer :) At least for sure not in `2.0.x`. Be it my will, I would leave this defect open (its just an extra read of a secret, potentially for some users, not that of a big deal) and instead take this re-factoring in. It's not just about the fact that I wrote it, I don't really care about that; but it puts our code base in a much better shape and it fixes many issues/inconsistencies we currently have.

> I kind of dislike passing null and an empty Map to signify these different use cases.

YES! You make perfect sense, but as soon as you go into the details, things complicate badly, as this updated PR shows. 

This PR will of course fail, but I've added the changes (in fabric8 only), so that you could have an understanding where I would take this.

wind57: I've pushed an implementation for fabric8 that works and passes all the tests, I will not do the same for k8s-client until I get some feedback. Otherwise I might invest a lot of time into something that will end up nowhere. thank you for looking into this!

wind57: the idea is not complicated at all.

- create a `NormalizedSource`, from it create `NamedSecretNormalizedSource` and `LabeledSecretNormalizedSource`.
- define a type called `NormalizedSourceType`
- instead of passing lots of arguments around, create a `Fabric8ConfigContext`
- With the above create such a strategy pattern: 

```
EnumMap<NormalizedSourceType, Function<Fabric8ConfigContext, Map.Entry<String, Map<String, Object>>>> STRATEGIES = ...
```

apply each strategy type for each `NormalizedSourceType`. It's doing the same exact things it did (with some fixes that are not transparent), but in a way that makes the code easier to extend. 

If you ever agree that this can go in, in some release, same changes should be done for config maps, to stay on par - though it is not mandatory, as config maps do not have labels selector currently.

spring-cloud-kubernetes-fabric8-config/src/main/java/org/springframework/cloud/kubernetes/fabric8/config/Fabric8SecretsPropertySource.java
```diff
@@ -37,47 +45,99 @@
...
+		STRATEGIES.put(NormalizedSourceType.LABELED_SECRET, labeledSecret());
+	}
+
 	private static final Log LOG = LogFactory.getLog(Fabric8SecretsPropertySource.class);
 
-	public Fabric8SecretsPropertySource(KubernetesClient client, String name, String namespace,
-			Map<String, String> labels, boolean failFast) {
-		super(getSourceName(name, getApplicationNamespace(client, namespace, "Secret", null)), getSourceData(client,
-				name, getApplicationNamespace(client, namespace, "Secret", null), labels, failFast));
+	public Fabric8SecretsPropertySource(Fabric8ConfigContext context) {
```
ryanjbaxter, line 58: Breaking change?

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/SecretsPropertySource.java
```diff
@@ -29,13 +31,12 @@
  */
 public class SecretsPropertySource extends MapPropertySource {
 
-	public SecretsPropertySource(String name, Map<String, Object> source) {
-		super(name, source);
+	public SecretsPropertySource(Map.Entry<String, Map<String, Object>> entry) {
```
ryanjbaxter, line 34: Breaking change?

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/SecretsConfigProperties.java
```diff
@@ -158,64 +169,25 @@ public boolean isEmpty() {
 			return !StringUtils.hasLength(this.name) && !StringUtils.hasLength(this.namespace);
 		}
 
-		public SecretsConfigProperties.NormalizedSource normalize(String defaultName, String defaultNamespace,
-				Map<String, String> defaultLabels) {
+		public Stream<NormalizedSource> normalize(String defaultName, String defaultNamespace,
```
ryanjbaxter, line 172: Breaking change?

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/SecretsConfigProperties.java
```diff
@@ -93,14 +97,20 @@ public String getConfigurationTarget() {
 	 * These are the actual name/namespace pairs that are used to create a
 	 * SecretsPropertySource
 	 */
-	public List<SecretsConfigProperties.NormalizedSource> determineSources() {
+	public List<NormalizedSource> determineSources(Environment environment) {
```
ryanjbaxter, line 100: Breaking change?

ryanjbaxter (commented):
Overall I like this much better!!!!

I know it can be super frustrating when trying to make things better without breaking anything, and I greatly appreciate all your effort.

There looks like there are some breaking changes here.  If you think its better to live with the bug at the moment, and wait to make these changes till we can break things I am fine with that.  I do think we should keep these in main for now, if we do move forward.

wind57:
yeah, I am totally fine with this bug until we can refactor. As said, it's just a read of the _same_ secret multiple times, and only for users that have more than one sources with labels, chances are slim... 

I will try to work a bit on moving everything to Kind and brush the pipeline and tests a bit, and then re-run all these builds, so that they are in a clean state (that might take a while as I have an M1 Mac and it seems there are many issues I am facing).

Thank you for your feedback!

ryanjbaxter:
Thanks appreciate it!

wind57:
Well... the name of the PR is wrong, to begin with.

What this actually does is refactor the secrets implementation, in a major way and has a lot of breaking changes. If you agree on the overall path of the code base, then imho, this should include config maps changes also (that are not here). 

To put it very simple: I want to clean-up/refactor both config maps and secrets implementation. This will bring the code base in a far more manageable way, but will make it harder to review. I was actually preparing for this kind or work for the past year or so, that is why I wanted to remove the deprecations, move to boot API for conditional on kubernetes, etc. 

Proper, rigorous unit tests will be put in place, of course and as usual, I will support these changes for potential issues...

ryanjbaxter:
OK that makes sense.  Something else we need to do since we are ripping things apart is removing the dependency on the bootstrap configuration as we want to eventually move away from that in all Spring Cloud projects.  If at all possible, I would prefer we break these changes up into individual PRs.  I understand that might not always be possible but something to keep in mind.  Can you move this to draft state for now?

wind57:
the bootstrap issue has really sparked my interest when I first saw [this](https://github.com/spring-cloud/spring-cloud-kubernetes/issues/794). I have looked a bit at the replacements APIs and have an idea or two, it is on my radar, FYI

ryanjbaxter:
Yeah I'm going to toss some ideas around with the team and come up with a proposal.

wind57:
I am still working on these refactorings. fabric8 implementation is done. I was thinking that it is complicated to review these things, so I am always available for some 1-1 calls over zoom/meet/etc to get over this changes. Just throwing this idea in here

ryanjbaxter:
Awesome let me know when it's ready and I will let you know!

wind57:
Here is an overview of what I tried to do in this PR.

- it touched only config maps and secrets implementations and refactors that specific code.
- Everything starts from a `NormalizedSource`. Such a source is computed from user provided properties _only_. It has a `name`, `namespace` (both can be `null` at this point) and `failSafe` (`false` by default). 
- It has 3 known implementations (at least one more will appear shortly: `LabeledConfigMap`). For example, one change here is to move "configTarget" field into each of the sources and remove it from properties, which to me makes a lot more sense. 
- Each also provides its "type", via `NormalizedSourceType`. This is later needed to compute the strategies on how to find the actual property source data. 
- It also has an option to be "frozen". This makes sense if you think about the transition from a `NormalizedSource` to a `Fabric8ConfigContext`, specifically the `namespace` property. In case users do not specify the namespace - which means it can be null, we find it out elsewhere and then place it in `Fabric8ConfigContext`. This means that calling `NormalizedSource::namespace` should not be allowed at this point in time, as the namespace is already present in `Fabric8ConfigContext::namespace` and obviously is not null anymore. 
- It help to think about a `NormalizedSource` as a transition phase to a specific context, like `Fabric8ConfigContext`, which is the only thing needed to find out the property sources data. 

---

Once we have our `NormalizedSource`, we pass it to `Fabric8SecretsPropertySourceLocator::getPropertySource`. From it (+ some more data), we get a `Fabric8ConfigContext`, that we pass to `Fabric8SecretsPropertySource`. 

This is the place where strategies come in place:

```
EnumMap<NormalizedSourceType, Fabric8ContextToSourceData> STRATEGIES = new EnumMap<>(
			NormalizedSourceType.class);
```

`NormalizedSourceType` is easy - it denotes the type of the Key in this map. `Fabric8ContextToSourceData` is a shorthand from `Function<Fabric8ConfigContext, SourceData>`, where `SourceData` is `record SourceData(String sourceName, Map<String, Object> sourceData)`.

This should somehow make sense : once we find out our entire "context" (`Fabric8ConfigContext`), i.e: all the data needed to get the source data, all we need is a `Function` that will be applied for that context. 

This is where `NamedSecretContextToSourceDataProvider` (values in that strategy map comes from), for example. These keys take a few `Function/BiFunction` as input, these are just references to methods from `SecretsPropertySource`. I need them as method references because they are protected, and must stay that way actually. 

---

This PR also tightens the visibility and things we expose, as well as tries to delegate as much as possible to common code - so that fabric8 and k8s native are as close as possible. It only fixes a few defects along the way. Now the code, to me, is a lot easier to read and maintain - also add functionality. 

--- 

@ryanjbaxter as said before, still available for calls over any known channels if there are questions. thank you.

ryanjbaxter:
I'm sorry we are just so busy appreciate your patience.  I am on vacation this coming week.  It will be on my list when I get back

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/NormalizedSource.java
```diff
@@ -0,0 +1,80 @@
...
+	 */
+	public abstract NormalizedSourceType type();
+
+	public abstract String target();
+
+	/**
+	 * calling this method will result in an Exceptions being thrown when calling some
+	 * getters. At the moment, this is done for the namespace only.
+	 */
+	public final void freeze() {
```
ryanjbaxter, line 70: Not sure I understand what this does....I must be missing something :)

spring-cloud-kubernetes-fabric8-config/src/main/java/org/springframework/cloud/kubernetes/fabric8/config/Fabric8ContextToSourceData.java
```diff
@@ -0,0 +1,30 @@
...
+import java.util.function.Function;
+
+import org.springframework.cloud.kubernetes.commons.config.SourceData;
+
+/**
+ * A more succinct way to define a Function from Fabric8ConfigContext to SourceData.
+ *
+ * @author wind57
+ */
+interface Fabric8ContextToSourceData extends Function<Fabric8ConfigContext, SourceData> {
```
ryanjbaxter, line 28: Any reason to not make these public?

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/NormalizedSource.java
wind57: you're not. I did not plan for this, initially. I only added it after making the mistake a couple of times and then not understanding why tests are failing. 

So:

- users can specify a namespace, this will be translated in our codebase into a field in the `NormalizedSource`
- `NormalizedSource` is a field in `Fabric8ConfigContext`
- In `Fabric8SecretsPropertySourceLocator` we need to find out the "real" namespace (if `NormalizedSource` does not contain it), as such:

```
String namespace = getApplicationNamespace(client, normalizedSource.namespace(), normalizedSource.target(),
				provider);
		normalizedSource.freeze();
		Fabric8ConfigContext context = new Fabric8ConfigContext(client, normalizedSource, namespace, environment);
``` 

Now think about for the classes that get a `Fabric8ConfigContext` as input. They need the namespace to be able to make the actual requests. They have a choice:

- get it from `NormalizedSource::namespace`
- get it directly as a field in `Fabric8ConfigContext`

which one is correct? The second one is, because the first one might be missing entirely. 

In order for the implementations not to make this mistake (this is where I failed a couple of times), I "freeze" the NormalizedSource, since the namespace computation (the final one - the one that we will be using to actually read the configmap/secret) was done, so reading the namespace from it, is actually incorrect. 

Sorry for the long wording.

spring-cloud-kubernetes-fabric8-config/src/main/java/org/springframework/cloud/kubernetes/fabric8/config/Fabric8ContextToSourceData.java
wind57: not really :) I was just going nuts to hide everything :D not sure if I am correct in this case or not. I'll listen to your advise

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/NormalizedSource.java
ryanjbaxter: Thanks for the explanation!

I think it might be better to encapsulate the logic behind where to get the namespace from in a method in the `ConfigContext`.

wind57: I thought about that, and that is not that easy. It will again take a lot of words explaining why, though. If you are against this freezing (which I totally get), we can drop it and just enforce the fact that reading the namespace should be done from `Fabric8ConfigContext`, via unit tests. 

This freezing, in general, is "kind of" a protection. You can easily _not_ call `NormalizedSource::freeze` and of course can call `NormalizedSource::namespace` afterwards, so I am more than OK dropping this freezing.

ryanjbaxter: Yes lets drop it.

wind57: will do! thx. I just can't build this currently because of [this issue](https://github.com/spring-cloud/spring-cloud-kubernetes/pull/985). Just FYI

PR #924, opened 2021-11-23
CircleCi Test Enhancements

ryanjbaxter:


wind57:
you could use http://www.yamllint.com/ to see if lint is fine...

.circleci/config.yml
```diff
@@ -46,85 +43,105 @@ jobs:
...
-            echo "Kubernetes cluster launched"
+                echo "Kubernetes cluster launched"
 
-            # Allow intra-pod communication
-            sudo iptables -P FORWARD ACCEPT
+                # Allow intra-pod communication
+                sudo iptables -P FORWARD ACCEPT
 
-            echo n |/snap/bin/microk8s.enable dns registry istio
+                echo n |/snap/bin/microk8s.enable dns registry istio
```
wind57, line 77: this `istio` profile is where I currently struggle (working on it), we have a single integration test that works with istio and to move it to kind take a while. I am working on it, but since you made this PR anyway, I will ask. We want to do that, still, correct? that is : move istio test to kind also

ryanjbaxter: yeah i think so

wind57: np at all! thx for the feedback

wind57:
I am almost ready with the PR that drops micro-k8s, but I will wait for this one to go in first and then I will merge my changes against that. FYI @ryanjbaxter happy Thanksgiving!

ryanjbaxter:
You too!

I hit a snag where I am getting a classnotfoundexcepion when one of the containers is starting for some reason.  I will look at it again next week

.circleci/config.yml
```diff
@@ -125,6 +134,40 @@ jobs:
...
+          command: |
+            ./mvnw -s .settings.xml clean install -Dservice.occurence=${_SERVICE_OCCURENCE} -DskipTests
+      - run:
+          name: save docker images
+          command: |
+            TAG=$(./mvnw -q -Dexec.executable=echo -Dexec.args='${project.version}' --non-recursive exec:exec)
+            echo $TAG
+            mkdir docker-images
+            docker save -o docker-images/spring-cloud-kubernetes-configuration-watcher.tar docker.io/springcloud/spring-cloud-kubernetes-configuration-watcher:${TAG}
+      - persist_to_workspace:
```
wind57, line 164: I understand what this steps does, but I don't understand why we need it :( are we saving some builds in this manner? thank you

spring-cloud-kubernetes-integration-tests/run.sh
```diff
@@ -116,17 +116,33 @@ main() {
...
 	# running tests..
-	for p in "${INTEGRATION_PROJECTS[@]}"; do
+	if [[ $CIRCLECI ]]; then
+		#This splits projects across all circleci instances, it returns a list of projects separated by a space
+		SPLIT_PROJECTS=$(printf "%s\n" "${INTEGRATION_PROJECTS[@]}" | circleci tests split)
+		SPLIT_PROJECTS=$(echo $SPLIT_PROJECTS | sed 's/ /,/g')
+		echo "split tests $SPLIT_PROJECTS"
+		#This splits the projects back into an array so we can iterate over them
+		IFS=',' read -ra PROJECTS <<< "$SPLIT_PROJECTS"
+		echo "projects $PROJECTS"
```
wind57, line 126: should this be `echo "projects ${PROJECTS[@]}"`? I mean `echo "projects $PROJECTS"` will show only the first one in the array.

But why the need for `IFS=',' read -ra PROJECTS <<< "$SPLIT_PROJECTS"`? We already have this array as `INTEGRATION_PROJECTS`. I mean if you do : 

```
	echo "${PROJECTS[@]}"
	echo "********"
	echo "${INTEGRATION_PROJECTS[@]}"

```

you will get the same output in both arrays. Unless I am missing something, we can drop creating the `PROJECTS` array.

wind57: As an addendum, I am also wondering why we would not split this by timings? `--split-by=timings --timings-type=classname`... I guess it's because there is no place to save the results to, like via that `store_test_results`, right?

wind57:
sorry for missing so much, I changed countries (even continents) and time has become sparse, but I am still active, just FYI; just slow at times.

.circleci/config.yml
```diff
@@ -17,31 +17,34 @@
...
+  build-then-test:
+    jobs:
+      - build
+      - test:
+          requires:
+            - build
 jobs:
-  build:
+  test:
+    parallelism: 5 # parallel containers to split the tests among
```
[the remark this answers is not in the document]
ryanjbaxter: This sounds good.

.circleci/config.yml
```diff
@@ -17,31 +17,34 @@
...
-            ./mvnw -s .settings.xml -U --fail-never dependency:go-offline -Dservice.occurence=${_SERVICE_OCCURENCE} || true
-      - save_cache:
-          paths:
-            - ~/.m2
-          key: spring-cloud-kubernetes-{{ .Branch }}-{{ checksum "pom.xml" }}
       - run:
           name: Run regular tests
           command: |
-            ./mvnw -s .settings.xml clean install -Dservice.occurence=${_SERVICE_OCCURENCE} #org.jacoco:jacoco-maven-plugin:prepare-agent install -U -P sonar -nsu --batch-mode -Dmaven.test.redirectTestOutputToFile=true -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn
+            CLASSNAMES=$(circleci tests glob "**/src/test/**/**.java" | grep -v 'spring-cloud-kubernetes-integration-tests' \
```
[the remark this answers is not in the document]
ryanjbaxter: I don't think including those files hurts anything but yeah we can check to see if there is an `@Test` in the file too

.circleci/config.yml
```diff
@@ -125,6 +134,40 @@ jobs:
       - store_test_results:
           path: ~/junit/
           destination: testartifacts
+  build:
+    machine: true
+    environment:
```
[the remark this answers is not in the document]
ryanjbaxter: yup

.circleci/config.yml
ryanjbaxter: Its a bit of a time hack.  Basically we are building these images during the build stage, so to avoid building them again for the integration tests, I save them off and load them again in the test stage.

spring-cloud-kubernetes-integration-tests/run.sh
ryanjbaxter: `PROJECTS` are the split `INTEGRATION_PROJECTS` its a subset.
We could use timing data, but there are not a large amount of integration tests here and its almost 1-2 tests per instance at the moment, so splitting by timing is probably not going to buy us anything.

.circleci/config.yml
wind57: nice! makes total sense

PR #935, opened 2021-12-13
CircleCi Cleanup

ryanjbaxter:


ryanjbaxter:
@wind57 new PR based on your suggestions

wind57:
hmm, neither the `-Xmx/Xms` and `large` did not work? :(

ryanjbaxter:
There was some kind of error with the java options
https://app.circleci.com/pipelines/github/spring-cloud/spring-cloud-kubernetes/1355/workflows/ba94f2bf-d458-434b-8ac0-becf3a98fcf5/jobs/2572

And I am not sure about the machine size at the moment...all i know is that the NGINX ingress is failing to start since making these changes...

wind57:
ah! it's probably the way java processes do not give memory back to the OS, so there is little memory for kind to operate against. thank you for pointing me to the build

ryanjbaxter:
@wind57 finally passed, any objections to this?

wind57:
None, sir!

wind57:
There are two more problems that we need to fix, one is easy, one not so much. 

First one is something that I faced just now and it is present in this PR also.

If you look at the "Run regular tests" phase, you will see that it starts with "No timing found for..." and the suggestion [from here](https://discuss.circleci.com/t/test-timings-cant-be-found/33702/18) is to drop that `--timings-type=classname`. I am over-simplifying this, but that is the crox of that page. 

My testing showed that this is indeed correct and after I drop `--timings-type=classname`, I see "Autodetected classname timings.", which is exactly what we want. 

---

The other problem is that `@Nested` tests do not work as we want them to, that is, I see lines like this in one of my PRs (where I integrate your changes with parallel splitting):

```
Autodetected classname timings.
No timing found for "org.springframework.cloud.kubernetes.client.KubernetesClientActuatorTests"
No timing found for "org.springframework.cloud.kubernetes.client.config.KubernetesClientBootstrapConfigurationTests"
.....
```

The problem is slightly bigger here: we run the tests via `mvn -Dtest = .....`. If, for example, you do: 

```
cd spring-cloud-kubernetes-fabric8-config
mvn clean test -Dtest=org.springframework.cloud.kubernetes.fabric8.config.Fabric8BootstrapConfigurationTests
```

build will fail, saying that "No tests were executed!", which means that none of the tests where we have `@Nested` configuration is actually going to run. :| 

If I do `ls -R **/src/test/**/*.java | grep -v 'spring-cloud-kubernetes-integration-tests' | xargs grep -l '@Test'  | xargs grep -l "@Nested"`, I get that we have these classes that use `@Nested`: 

```
org.springframework.cloud.kubernetes.client.KubernetesClientActuatorTests
org.springframework.cloud.kubernetes.client.default_api.KubernetesClientDefaultApiClientTests
org.springframework.cloud.kubernetes.client.config.KubernetesClientBootstrapConfigurationTests
org.springframework.cloud.kubernetes.client.config.KubernetesClientConfigMapPropertySourceLocatorRetryTests
org.springframework.cloud.kubernetes.client.config.KubernetesClientSecretsPropertySourceLocatorRetryTests
org.springframework.cloud.kubernetes.commons.config.KubernetesBootstrapConfigurationTests
org.springframework.cloud.kubernetes.configserver.KubernetesConfigServerAutoConfigurationTests
org.springframework.cloud.kubernetes.fabric8.Fabric8ClientUserAgentTests
org.springframework.cloud.kubernetes.fabric8.config.Fabric8ActuatorTests
org.springframework.cloud.kubernetes.fabric8.config.Fabric8BootstrapConfigurationTests
org.springframework.cloud.kubernetes.fabric8.config.Fabric8ConfigMapPropertySourceLocatorRetryTests
org.springframework.cloud.kubernetes.fabric8.config.Fabric8SecretsPropertySourceLocatorRetryTests
org.springframework.cloud.kubernetes.fabric8.istio.IstioAutoConfigurationTests
```

After I investigated a bit, I found that this is an issue in `surefire`, see [here](https://issues.apache.org/jira/projects/SUREFIRE/issues/SUREFIRE-1633?filter=allopenissues) 

As of now, I don't see any other option then refactoring these tests...

wind57:
The above worked, so to recap:

- first we need to drop `--timings-type=classname`, this will ensure that all non-nested classes will have their timing saved, so that circleci can analyze them. That is, with this change, we will see : "Autodetected classname timings".
- second, we need to add : 

```
<!-- workaround for https://issues.apache.org/jira/projects/SUREFIRE/issues/SUREFIRE-1633?filter=allopenissues -->
<!-- we run tests with mvn -DtestsToRun=.... in the pipeline-->
<includes>
    <include>${testsToRun}</include>
</includes>
```

in surefire plugin (in two places in root `pom.xml`, the other one is for the `sonar` profile) and change `config.yaml` to use : `-DtestsToRun=$TEST_ARG` instead of `-Dtest=$TEST_ARG`.

This change will make sure that `@Nested` are actually run (unlike what we currently have).

---

We will still be left with a problem, though. We would still _not_ be saving timing data for the classes that have `@Nested` classes inside them. That is, though `@Nested` are going to run, we would see entries like:

```
No timing found for "org.springframework.cloud.kubernetes.client.KubernetesClientActuatorTests"
```

in logs.

I am going to try and fix this one also.

ryanjbaxter:
@wind57 thanks!  if `testsToRun` is not specified what happens?  for example if I run `./mvnw clean install` are all the tests still run?  I would think so...

ryanjbaxter:
When I run `./mvnw clean install` locally with these changes its trying to run the integration tests, and those obviously fail

wind57:
indeed, I just saw that my PR where these changes are present fails too because integration tests are running. This is fixable though:

- add a property in root `pom.xml`: `<excludeITTests>**/*IT.java</excludeITTests>`
- reference that property in surefire plugin (in two places, for sonar also):

```
<excludes>
    <exclude>${excludeITTests}</exclude>
</excludes>
```

change `run.sh` : 

```
"${KIND}" load docker-image docker.io/springcloud/$p:${PROJECT_VERSION}
# empty excludeITTests, so that integration tests will run
${MVN} clean install -DexcludeITTests=
cd ..
```

With these changes, my build passes : https://app.circleci.com/pipelines/github/spring-cloud/spring-cloud-kubernetes/1377/workflows/a11bcb26-753a-4155-9cfa-79fc79628378/jobs/2619

---

The only thing left is to treat `@Nested` correctly, so that timings are saved. I will try to fix this also in the upcoming days

ryanjbaxter:
Thanks that is what I was thinking but I had to wrap up our release yesterday so didnt have a chance to investigate.

>The only thing left is to treat @Nested correctly, so that timings are saved. I will try to fix this also in the upcoming days

Im going to merge this now as I am going on vacation for a few weeks and I dont want to leave it hanging.  If you figure out the `@Nested` tests just create a PR and we will get that merged.  Thanks!

wind57:
I'm jealous, but enjoy your vacation to the fullest!

PR #960, opened 2022-01-25
Fix build

wind57:


spring-cloud-kubernetes-integration-tests/simple-core/pom.xml
```diff
@@ -71,6 +71,19 @@
 			<artifactId>rest-assured</artifactId>
 			<scope>test</scope>
 		</dependency>
+
+		<dependency>
```
wind57, line 75: we only need this for integration tests on micro-k8s. @ryanjbaxter

wind57:
closing, as these changes are already picked-up in the other PR

wind57:
oh, that makes sense! every change like this makes me start the day with an extra coffee... I should really apply for a job at pivotal one day, then I'll know all these changes. probably. :) again, thank you for pointing this out to me!

ryanjbaxter:
Yeah I realize I have a bit of an inside track here.  I saw the build break last night when I made the commit and then saw why an knew it had to be connected to a Boot change because the Boot team told us they added back in rest-assurred earlier in the day.  No worries at all, just wanted to connect the dots for you.

PR #962, opened 2022-01-25
Nested split

wind57:


wind57:
this literally just removes all `@Nested` tests that we have and creates separate ones. It might be difficult to review, but I have not changed the flow, the logic, etc - not a tiny bit.

wind57:
I have a question if there is interested in this one, in general, or I should close it. Either way is fine with me, I just want to know if I should periodically update it and keep in fresh with the latest main. Nothing more, nothing less. Thank you for your time @ryanjbaxter

PR #967, opened 2022-02-05
clean-up

wind57:


wind57:
we merged the branch that removed `@Nested` in our code-base. That branch was open for a while and I was merging main into it from time to time. Unfortunately while I was bringing changes from `main`, some classes were brought in...

Basically this PR takes care to remove the left-overs of `@Nested` classes that came in when I was merging `main`. 
It's not like we remove tests - these classes were already refactored and tests from them are already present in our code base as separated classes.

I hope this makes sense.

wind57:
@ryanjbaxter minor left-overs clean-up

PR #980, opened 2022-02-27
log with warn fix 978

wind57:


osvso:
@wind57 Why you changed the log level to warn again? it's such a logging abuse. 
https://github.com/spring-cloud/spring-cloud-kubernetes/pull/907

wind57:
I don't remember my exact reasonings now. You can change the log level as you want via a property.

osvso:
Yes, I know. I would prefer to not be forced to exclude logging from a library that plays an important role.

wind57:
agreed. we can probably move it to even `debug`, as it rarely matters at all. 

In the `3.0.x` releases, it's not even there, at all. wdyt about moving it to debug?

ryanjbaxter:
Done https://github.com/spring-cloud/spring-cloud-kubernetes/commit/fa013d31803537b094e226f17e139d2d98c56907

PR #982, opened 2022-03-09
change to least privileged role rules for leader-election

mohsen0:


wind57:
I don't know what to say about this PR, to be honest. I understand your changes and they do make sense, least privileged indeed makes sense here. At the same time this is just an example obviously, so anyone looking at this examples should make the correct choices based on their needs. 

That being said, we all know that the "copy/paste" mentality is a big one :), so I do agree to have your changes incorporated. Thank you for making this PR, I'll wait to see what @ryanjbaxter thinks.

ryanjbaxter:
Probably better to be more restrictive than not

mohsen0:
> I don't know what to say about this PR, to be honest. I understand your changes and they do make sense, least privileged indeed makes sense here. At the same time this is just an example obviously, so anyone looking at this examples should make the correct choices based on their needs.
> 
> That being said, we all know that the "copy/paste" mentality is a big one :), so I do agree to have your changes incorporated. Thank you for making this PR, I'll wait to see what @ryanjbaxter thinks.

You would expect the documentation to promote best practices, I thought about adding this to the readme. but since non of the other projects had such examples and even this policy still is too permissive toward pods. I thought I add this here, for now, to just start the conversation.

Thanks

PR #984, opened 2022-03-11
spring-cloud-kubernetes-client-config-it move to test-containers

wind57:


spring-cloud-kubernetes-integration-tests/pom.xml
```diff
@@ -13,22 +13,12 @@
...
 
 	<name>Spring Cloud Kubernetes :: Integration Tests</name>
-	<description>Integration tests where SCK applications are run inside a Kubernetes
-		cluster
-	</description>
+	<description>Integration tests where SCK applications are run inside a Kubernetes cluster</description>
 
 	<properties>
 		<java.version>17</java.version>
-		<arquillian-cube.version>1.18.2</arquillian-cube.version>
```
wind57, line 22: minor clean-up of things we are not using anymore

spring-cloud-kubernetes-integration-tests/run.sh
```diff
@@ -179,6 +181,10 @@ main() {
     # teardown will happen automatically on exit
 }
 
+MOVED_TO_TEST_CONTAINERS=(
```
wind57, line 184: I'd like to move one module at a time to testcontainers, so this just tracks what project is already moved. When all of them will be moved, I will drop this, obviously.

spring-cloud-kubernetes-integration-tests/spring-cloud-kubernetes-client-config-it/src/test/java/org/springframework/cloud/kubernetes/client/config/it/ConfigMapAndSecretIT.java
```diff
@@ -78,17 +74,35 @@
 
 	private static K8SUtils k8SUtils;
 
+	private static final K3sContainer K3S = new K3sContainer(DockerImageName.parse("rancher/k3s:v1.21.10-k3s1"))
```
wind57, line 77: - start an instance of `K3sContainer` with image `rancher/k3s:v1.21.10-k3s1` as base
- bind `/tmp/images` volume. before when we invoke this piece of code, `docker save ....` already happened, so the image is there
- `withCommand("server")` so that we get `traefik` installed
- `.withReuse(true)` to speed-up the process (we also need `echo "testcontainers.reuse.enable=true" > ~/.testcontainers.properties` for that to work)

spring-cloud-kubernetes-integration-tests/spring-cloud-kubernetes-client-config-it/src/test/java/org/springframework/cloud/kubernetes/client/config/it/ConfigMapAndSecretIT.java
```diff
@@ -78,17 +74,35 @@
...
+	static void setup() throws Exception {
+		K3S.start();
+		K3S.execInContainer("ctr", "i", "import", "/tmp/images/spring-cloud-kubernetes-client-config-it.tar");
+		createApiClient(K3S.getKubeConfigYaml());
 		api = new CoreV1Api();
 		appsApi = new AppsV1Api();
 		networkingApi = new NetworkingV1Api();
 		k8SUtils = new K8SUtils(api, appsApi);
+
+		RbacAuthorizationV1Api rbacApi = new RbacAuthorizationV1Api();
```
wind57, line 92: we used to that via `run.sh`, but now, since we are moving to testcontainers, this is done in each project, also moved all the needed files for that.

spring-cloud-kubernetes-test-support/src/main/java/org/springframework/cloud/kubernetes/integration/tests/commons/K8SUtils.java
```diff
@@ -218,7 +269,8 @@ public boolean isDeploymentReady(String deploymentName, String namespace) throws
 		}
 		V1Deployment deployment = deployments.getItems().get(0);
 		Integer availableReplicas = deployment.getStatus().getAvailableReplicas();
-		log.info("Available replicas for " + deploymentName + ": " + availableReplicas);
```
wind57, line 221: I was really getting annoyed but that `null` for available replicas logs. not a big deal, but I feel better now

spring-cloud-kubernetes-test-support/src/main/java/org/springframework/cloud/kubernetes/integration/tests/commons/K8SUtils.java
```diff
@@ -195,6 +211,41 @@ public void waitForDeployment(String deploymentName, String namespace) {
 				.until(() -> isDeploymentReady(deploymentName, namespace));
 	}
 
+	public void waitForIngress(String ingressName, String namespace) {
```
wind57, line 214: instead of retrying hitting an endpoint, before ingress is up and running, imho: the idiomatic way to do it is to wait for ingress to be ready and only then hit an endpoint. 

This method does exactly that : waits for ingress

spring-cloud-kubernetes-test-support/src/main/java/org/springframework/cloud/kubernetes/integration/tests/commons/K8SUtils.java
```diff
@@ -97,9 +104,18 @@ public static ApiClient createApiClient(boolean debug, Duration readTimeout) thr
 		return client;
 	}
 
+	public static ApiClient createApiClient(String configFile) throws IOException {
```
wind57, line 107: we need to create a client from a config file

spring-cloud-kubernetes-integration-tests/spring-cloud-kubernetes-client-config-it/src/test/java/org/springframework/cloud/kubernetes/client/config/it/ConfigMapAndSecretIT.java
```diff
@@ -78,17 +74,35 @@
...
+		api.createNamespacedServiceAccount(NAMESPACE, getConfigK8sClientItServiceAccount(), null, null, null);
+		rbacApi.createNamespacedRoleBinding(NAMESPACE, getConfigK8sClientItRoleBinding(), null, null, null);
+		rbacApi.createNamespacedRole(NAMESPACE, getConfigK8sClientItRole(), null, null, null);
+	}
+
+	@AfterAll
+	static void afterAll() throws Exception {
```
wind57, line 99: unload the image, once we are done. this is needed because we enable `withReuse` and tests might step on each other.

spring-cloud-kubernetes-integration-tests/spring-cloud-kubernetes-client-config-it/src/test/java/org/springframework/cloud/kubernetes/client/config/it/ConfigMapAndSecretIT.java
```diff
@@ -155,21 +120,56 @@ public void testConfigMapAndSecretWatchRefresh() throws Exception {
...
 		deployConfigK8sClientPollingIt();
 
 		// Check to make sure the controller deployment is ready
 		k8SUtils.waitForDeployment(SPRING_CLOUD_CLIENT_CONFIG_IT_DEPLOYMENT_NAME, NAMESPACE);
 		testConfigMapAndSecretRefresh();
 	}
 
+	void testConfigMapAndSecretRefresh() throws Exception {
+
+		String propertyURL = "localhost:" + K3S.getMappedPort(80) + "/myProperty";
```
wind57, line 133: I have slightly changed the code in this test, by using `WebClient`, or waiting for ingress, etc. Semantically, it checks the same things

spring-cloud-kubernetes-integration-tests/spring-cloud-kubernetes-client-config-it/src/test/resources/logback-test.xml
```diff
@@ -0,0 +1,14 @@
+<configuration>
```
wind57, line 1: without this, there will be _a lot_ of logs coming from testcontainers

wind57:
@ryanjbaxter if you do not mind, I'd like us to move to testcontainers one module at a time.

spring-cloud-kubernetes-integration-tests/spring-cloud-kubernetes-fabric8-client-loadbalancer/src/test/resources/wiremock-deployment.yaml
```diff
@@ -13,7 +13,7 @@ spec:
     spec:
       containers:
         - name: servicea-wiremock
-          image: rodolpheche/wiremock:2.27.2
+          image: wiremock/wiremock:2.32.0
```
wind57, line 16: the image that we use is deprecated, so just update to a newer, supported one (that has arm64 support also!)

.circleci/config.yml
```diff
@@ -43,6 +43,11 @@ jobs:
             sudo update-alternatives --set java /usr/lib/jvm/temurin-17-jdk-amd64/bin/java
             sudo update-alternatives --set javac /usr/lib/jvm/temurin-17-jdk-amd64/bin/javac
             java -version
+      - run:
```
wind57, line 46: needed for `.withReuse(true)` to work for testcontainers

spring-cloud-kubernetes-integration-tests/spring-cloud-kubernetes-client-loadbalancer-it/src/main/java/org/springframework/cloud/kubernetes/client/loadbalancer/it/KubernetesClientLoadBalancerApplicationIt.java
```diff
@@ -49,13 +54,15 @@ public static void main(String[] args) {
 
 	@Bean
 	@LoadBalanced
-	RestTemplate restTemplate() {
-		return new RestTemplateBuilder().build();
```
wind57, line 53: use reactive client, instead of rest template

spring-cloud-kubernetes-integration-tests/spring-cloud-kubernetes-client-loadbalancer-it/src/main/java/org/springframework/cloud/kubernetes/client/loadbalancer/it/KubernetesClientLoadBalancerApplicationIt.java
```diff
@@ -49,13 +54,15 @@ public static void main(String[] args) {
...
+	WebClient.Builder client() {
+		return WebClient.builder();
 	}
 
-	@GetMapping("/servicea")
+	@GetMapping("/loadbalancer-it/servicea")
```
wind57, line 61: I can't seem to find a way for traefik to re-write the url, like nginx used to do before. But that is not the point of the test anyway, so no big deal, imo.

spring-cloud-kubernetes-integration-tests/spring-cloud-kubernetes-client-reactive-discovery-client-it/src/main/resources/application.yaml
```diff
@@ -1,3 +1,7 @@
+spring:
```
wind57, line 1: so that `/actuator/heath` and `/services` are accessible on the same base. This is needed because traefik ingress does not seem to have a way to re-write the URL. As said before, this is not a big deal, since the test is supposed to test something else, not ingress rules

ryanjbaxter (commented):
This is great!  

A couple of high level thoughts/comments:

1. Do we still need run.sh in order to run these tests?  It appears so....
It would be ideal if we could get to a point where we could just run these integration tests individually, for example right inside an IDE.  Do you think we could ever get to that point?  Or will we always need run.sh in order to pull images that we need?

2. I think it would be useful to add a README to the integration test parent module describing how these integration tests work.

wind57:
> for example right inside an IDE...

yes, in the long run this is my goal. For the time being I want to first move to testcontainers for all modules - once that is done, I'll take a closer look on how to automate everything. If you do not mind, I'll take this effort in separate issue, because otherwise the files and work piles up badly... As such, run.sh is going to be here for a while.

> I think it would be useful to add a README to the integration test parent module describing how these integration tests work.

absolutely. I did not think about this, but this is indeed a good idea.

P.S. Sorry for answering so late, I am from rep. of Moldova (and now everybody knows about us), and I am a little involved in the refugee crisis, time is sparse lately.

.circleci/config.yml
```diff
@@ -69,6 +74,23 @@ jobs:
             docker load -i ./docker-images/spring-cloud-kubernetes-configuration-watcher.tar
             docker load -i ./docker-images/spring-cloud-kubernetes-discoveryserver.tar
             docker load -i ./docker-images/spring-cloud-kubernetes-configserver.tar
+      - run:
```
wind57, line 77: so it seems that there is a way to drop run.sh file that we currently have... I'll see what happens with the rest of modules.

spring-cloud-kubernetes-test-support/src/main/java/org/springframework/cloud/kubernetes/integration/tests/commons/Commons.java
```diff
@@ -0,0 +1,110 @@
...
+
+import static org.springframework.cloud.kubernetes.integration.tests.commons.K8SUtils.getPomVersion;
+
+/**
+ * A few commons things that can be re-used across clients. This is meant to be used for
+ * testing purposes only.
+ *
+ * @author wind57
+ */
+public final class Commons {
```
wind57, line 39: this is where the main magic happens for test-containers @ryanjbaxter

spring-cloud-kubernetes-test-support/src/main/java/org/springframework/cloud/kubernetes/integration/tests/commons/Commons.java
```diff
@@ -0,0 +1,138 @@
...
+		List<Image> images = CONTAINER.getDockerClient().listImagesCmd().exec();
+		images.stream()
+				.filter(x -> Arrays.stream(x.getRepoTags() == null ? new String[] {} : x.getRepoTags())
+						.anyMatch(y -> y.contains(image)))
+				.findFirst().orElseThrow(() -> new IllegalArgumentException("Image : " + image + " not build locally. "
+						+ "You need to build it first, and then run the test"));
+	}
+
+	/**
+	 * A K3sContainer, but with fixed port mappings. This is needed because of the nature
```
wind57, line 118: this is a little hacky-ish, but because of the nature of our tests, we need it.

spring-cloud-kubernetes-integration-tests/spring-cloud-kubernetes-configuration-watcher-it/src/test/java/org/springframework/cloud/kubernetes/configuration/watcher/ActuatorRefreshIT.java
```diff
@@ -49,66 +50,79 @@
 /**
  * @author Ryan Baxter
  */
-public class ActuatorRefreshIT {
-
-	private static final String CONFIG_WATCHER_WIREMOCK_DEPLOYMENT_NAME = "config-watcher-wiremock-deployment";
-
-	private static final String CONFIG_WATCHER_WIREMOCK_APP_NAME = "config-watcher-wiremock";
```
wind57, line 56: this test has slightly eaten my soul :) it took a long time to understand how it is _really_ set-up and understand that we need to bind static ports to testscontainers, basically like docker does it via : `docker ... -p 8080:8080`. The default mode in which testcontainers functions is that it assigns a random port to what is exposed. So if you expose something on 8080 (then tell testcontainers about it via `.withExposedPorts`), testcontainers will create a random port (55555 for example), that you can find out about via : `....getMappedPort(8080); // will return 55555` and _then_ use that `55555`.

Because of the way our tests are set-up, this is not an option for us, so we need to statically bind these ports.

spring-cloud-kubernetes-integration-tests/spring-cloud-kubernetes-discovery-client-it/src/test/resources/client/spring-cloud-kubernetes-discoveryclient-it-deployment.yaml
```diff
@@ -19,10 +19,10 @@ spec:
           readinessProbe:
             httpGet:
               port: 8080
-              path: /actuator/health/readiness
+              path: /discoveryclient-it/actuator/health/readiness
```
wind57, line 22: I have to re-write some of the paths like this... it's because traefik does not have a simple way to do what nginx does via "re-write target", so I have to dance around this links a little. It's really not a big deal, since we are testing entirely un-related things, just a bit annoying. :)

spring-cloud-kubernetes-integration-tests/spring-cloud-kubernetes-client-reactive-discovery-client-it/src/test/resources/spring-cloud-kubernetes-client-reactive-discovery-it-service.yaml
```diff
@@ -1,14 +0,0 @@
-apiVersion: v1
```
wind57, line 1: I removed this project, entirely. There is already such a project that does the same exact tests... I do not know (or remember) what happened and how we ended up in this situation, but this needs to be cleaned up.

.circleci/config.yml
```diff
@@ -69,6 +74,35 @@ jobs:
...
+          name: testcontainers integration tests for kubernetes-native client
+          command: |
+            # our integration tests use static binding of ports (see FixedPortsK3sContainer)
```
wind57, line 80: leaving the comment here, otherwise this info will be lost and it is important

wind57:
@ryanjbaxter this is ready to be reviewed. I'd like to split this work in two phases : k8s-native client and fabric8. If you look closely in `run.sh`, there will not be any k8s-native projects, everything is in the new phase in `config.yaml`. Which means : 

- we will no longer need `run.sh` once I move fabric8 projects to testcontainers
- you have the option to run everything locally, as you requested (I loved the idea)

The number of files is big, but the core changes are not that big. Mainly move away from RestTemplate, re-write some paths because of the change to traefik instead of nginx and a bunch of testcontainer related changes. I also slightly (where I could) moved yaml files to separate folders, to make this easier to read and maintain.

spring-cloud-kubernetes-integration-tests/run.sh
```diff
@@ -1,199 +0,0 @@
-#!/bin/bash
```
wind57, line 1: no more `run.sh`!

.circleci/config.yml
```diff
@@ -70,10 +75,25 @@ jobs:
             docker load -i ./docker-images/spring-cloud-kubernetes-discoveryserver.tar
             docker load -i ./docker-images/spring-cloud-kubernetes-configserver.tar
       - run:
-          name: Run Kind Integration Tests
+          name: testcontainers integration tests
           command: |
+            # we need sort, because the last project calls K3S::stop
```
ryanjbaxter, line 80: Can you elaborate on this?  Why does the last project call `stop`?

.circleci/config.yml
```diff
@@ -70,10 +75,25 @@ jobs:
...
+
             cd spring-cloud-kubernetes-integration-tests
-            ./run.sh
+            PROJECT_VERSION=$(mvn help:evaluate -Dexpression=project.version -q -DforceStdout)
+            for p in "${INTEGRATION_PROJECTS[@]}"; do
+              cd "${p}"
+              mvn spring-boot:build-image \
+              	  -Dspring-boot.build-image.imageName=docker.io/springcloud/$p:${PROJECT_VERSION} \
+                  -Dspring-boot.build-image.builder=paketobuildpacks/builder
+              mvn clean verify -DexcludeITTests=
```
ryanjbaxter, line 94: What is this doing?

.circleci/config.yml
```diff
@@ -70,10 +75,25 @@ jobs:
             docker load -i ./docker-images/spring-cloud-kubernetes-discoveryserver.tar
             docker load -i ./docker-images/spring-cloud-kubernetes-configserver.tar
       - run:
-          name: Run Kind Integration Tests
+          name: testcontainers integration tests
```
ryanjbaxter, line 78: Do you think it would be possible to run the integration tests as part of the regular build?  So ultimately we could just run `./mvnw clean install` and everything gets run?

.circleci/config.yml
wind57: we don't, you are correct. I removed this part

.circleci/config.yml
wind57: you are correct again, as a matter of fact, we now run those tests with simply `mvn clean install`. I'll comment where it needs your eyes on how this is done

.circleci/config.yml
wind57: YES! initially I did not believe this is feasible, but took my time (cause I really liked this idea of yours) and it's possible.

spring-cloud-kubernetes-integration-tests/spring-cloud-kubernetes-client-config-it/pom.xml
```diff
@@ -60,6 +77,56 @@
 				<filtering>true</filtering>
 			</resource>
 		</resources>
+
+		<plugins>
+			<!-- build image in the 'package' phase, and ignore plain tests -->
```
wind57, line 82: @ryanjbaxter every integration test has these 3 parts: 

- build the image in the `package` phase
- skip plain tests with `skipTests` in `maven-surefire-plugin`
- run integration tests in the `maven-failsafe-plugin`

In this manner, when running `mvn clean install`, we will first build the image and then run the tests

wind57:
I can't believe this took so much, but in the end, there is a PR... let me know if you have any questions what-so-ever

wind57:
I am converting this back to a Draft as it seems that spring-boot 3.0.0-SNAPSHOT introduces some checks that are now failures instead of warning. When I figure out what is going on and how to fix it - this shall stay a draft.

ryanjbaxter:
I saw some failure in the build on Friday, I was going to try and look at it on Monday

wind57:
yeah... I am struggling with this one for a week so far, I am getting closer and closer to understand what is going on, I think :)

spring-cloud-kubernetes-controllers/spring-cloud-kubernetes-configuration-watcher/src/main/resources/application-bus-amqp.yaml
```diff
@@ -0,0 +1,23 @@
+spring:
```
wind57, line 1: instead of having `activate.on-profile` - I've added these 2 files, one for each profile: `bus-amqp` and `bus-kafka` - this fixed the build here, but strangely enough - not in `main`. This this PR is entirely about a different thing - I am marking it as ready to review, for now.

spring-cloud-kubernetes-integration-tests/spring-cloud-kubernetes-fabric8-client-configmap/src/main/resources/application.yaml
```diff
@@ -0,0 +1,5 @@
+spring:
```
wind57, line 1: this test was reading from a configmap and asserting the value there. so why not move this one to config data? it would be the first integration test we have that has moved to config data

wind57:
@ryanjbaxter this one has been updated to the latest `main` also, and as such is ready for review too

ryanjbaxter:
@wind57 I spent some time messing around with this branch.  

First off this is great stuff!  Thanks so much!

There was a small problem with the fabric8 configmap test you converted to config data, we need to remove the bootstrap dependency.  I made a fix and committed that.

The other issue I ran into is the istio test.  It assumes that you have the istio cli installed locally at `/tmp/istio/istio-bin` right?

ryanjbaxter:
It's not just running locally but we do builds and releases from Jenkins internally, so we need it to run there as well.

I mean we already managing the dependency by installing a specific version of Istio ourselves, now we are just keeping the cli along with the code

wind57:
@ryanjbaxter thank you for fixing that issue, I was blind and could not realize to use `getAbsolutePath` :) I've merge `main` into this branch

PR #994, opened 2022-03-17
take useNameAsPrefix into consideration when there are no sources defined, in case of ConfigMap

wind57:


spring-cloud-kubernetes-fabric8-config/src/main/java/org/springframework/cloud/kubernetes/fabric8/config/NamedConfigMapContextToSourceDataProvider.java
```diff
@@ -128,8 +126,4 @@ public Fabric8ContextToSourceData get() {
 
 	}
```
wind57, line 128: this has moved to `ConfigMapConfigProperties`

wind57:
@ryanjbaxter I am closing this one for the time being. We sort discussed about something very similar and your thought process (which I get more and more inclined towards), is to not implement this very edge cases, which we have not received any complains about. If there will be, I will take care of them

PR #1013, opened 2022-05-12
Fix watcher

wind57:


wind57:
@ryanjbaxter I understand what is going on, but I am inclined to say this is a bug in spring-boot. 

The first issue is that something like this:

```
org.springframework.boot.context.config.InvalidConfigDataPropertyException: Property 'spring.profiles' imported from location 'class path resource [application.yml]' is invalid and should be replaced with 'spring.config.activate.on-profile' [origin: class path resource [application.yml] from app.jar - 34:13]
```

but this one is easy to fix.

Second issue, is that we have in config-watcher such a `application.yaml` (after fixing the above):

```
spring:
  cloud:
    bus:
      enabled: true
    stream:
      default-binder: rabbit
  config:
    activate:
      on-profile: bus-amqp
management:
  health:
    rabbit:
      enabled: true
---
spring:
  config:
    activate:
      on-profile: bus-kafka
  cloud:
    bus:
      enabled: true
    stream:
      default-binder: kafka
```

it turns out that `spring.config.activate.on-profile` does not work as expected. For example, I am running an integration test : `ActuatorRefreshRabbitMQIT` - which must use `default-binder: rabbit`, instead it uses `default-binder: kafka`. This means `spring.config.activate.on-profile` does not work correctly. Trying to figure out if this is known or an existing (recent) issue...

wind57:
this gets more confusing now, as the same fix works for me on the PR with test-containers (instead of a single application.yaml with `spring.config.activate.on-profile`, I have 3 : `application.yaml`, `application-bus-amqp.yaml`, `application-bus-kafka.yaml` _without_ `spring.config.activate.on-profile`), but fails here. Still trying to figure out

ryanjbaxter:
Hum this is odd...I wonder if it could be the `-` in the profile name causing an issue since that is a profile delimiter?  Just a guess really.

spring-cloud-kubernetes-controllers/spring-cloud-kubernetes-configuration-watcher/pom.xml
```diff
@@ -3,9 +3,9 @@
 		 xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
 		 xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
 	<parent>
```
wind57, line 5: slight formatting, nothing changed here

wind57:
Its not this :( I understood the root cause, but then build fails here and works for me locally :) I'll keep investigating. The good news is that test containers PR is ready and works

ryanjbaxter:
I made a commit to simplify the configuration...it should be additive, meaning we won't have to duplicate the configuration in the profile specific configuration

wind57:
indeed, that simplifies it, thank you. The issue is that this fails sometimes, I can not explain myself why. Same thing happens locally - it just randomly fails, without any code changes. :| 

The only good news is that once we remove these tests and move to test containers, this is something I will not have to think about anymore :)

PR #1017, opened 2022-05-23
Labeled secrets prefix support

wind57:
This PR is based on one that is in review currently, as such it makes no sense to look at, until that one is going to be merged.


spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/ConfigUtils.java
```diff
@@ -130,4 +149,43 @@ public static String sourceName(String target, String applicationName, String na
 		return target + PROPERTY_SOURCE_NAME_SEPARATOR + applicationName + PROPERTY_SOURCE_NAME_SEPARATOR + namespace;
 	}
 
+	public static final class Prefix {
```
wind57, line 152: this is a core change when adding support for prefix for labeled sources. First of all it is a nested class on purpose, so that I could access its `private` methods inside the parent class, but not expose them, not even to the package. 
This is needed so that any external interaction is only via its predefined instances.

The need for such a class comes when you think about sources that we find by labels. Let's take a secret for example. Users can specify searching for this secret by some labels, at the same time specifying that they want "useNameAsPrefix". Well, we will only know the name, once we find this source - at this point in the flow, we simply have no idea what the name will be. As such, I call this a "DELAYED" prefix, meaning - I know for sure that I need such a prefix, but I do not know its name _currently_. 

This change, in turn, triggers many tests to be updated, unfortunately.

spring-cloud-kubernetes-fabric8-config/src/main/java/org/springframework/cloud/kubernetes/fabric8/config/LabeledSecretContextToSourceDataProvider.java
```diff
@@ -77,9 +81,43 @@ public Fabric8ContextToSourceData get() {
 						.getItems();
 
 				if (!secrets.isEmpty()) {
-					secrets.forEach(secret -> result.putAll(dataFromSecret(secret, namespace)));
-					sourceName = secrets.stream().map(Secret::getMetadata).map(ObjectMeta::getName)
+
```
wind57, line 84: this is the main change for fabric8 implementation.

spring-cloud-kubernetes-fabric8-config/src/main/java/org/springframework/cloud/kubernetes/fabric8/config/LabeledSecretContextToSourceDataProvider.java
```diff
@@ -77,9 +81,43 @@ public Fabric8ContextToSourceData get() {
...
+					for (Secret secret : secrets) {
+						// we support prefix per source, not per secret. This means that
```
wind57, line 86: when we search "by labels" - we can find more than one secret, these secrets we find can easily have the same property names - and they will get overridden. This is expected and more fine-grained labels would be the solution here.

wind57:
@ryanjbaxter support for prefix for labeled secrets + documentation added as talked in the previous PR

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/NamedSecretNormalizedSource.java
```diff
@@ -25,14 +25,14 @@
  */
 public final class NamedSecretNormalizedSource extends NormalizedSource {
 
-	private final String prefix;
+	private final ConfigUtils.Prefix prefix;
 
-	public NamedSecretNormalizedSource(String name, String namespace, boolean failFast, String prefix) {
+	public NamedSecretNormalizedSource(String name, String namespace, boolean failFast, ConfigUtils.Prefix prefix) {
```
ryanjbaxter, line 30: Would it be worth having a constructor that doesn't take a `Prefix`?

There are several places where we create a new instance but are just passing `UNSET` as the prefix, seems easier to create an instance without specifying a prefix we just automatically set it to `UNSET`.  

Also using/specifying a prefix is optional so it makes sense to have a constructor that doesn't take one.

(This also applies to the Config Map version as well)

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/ConfigUtils.java
```diff
@@ -130,4 +149,43 @@ public static String sourceName(String target, String applicationName, String na
...
+
+		/**
+		 * prefix has not been provided.
+		 */
+		public static final Prefix UNSET = new Prefix(() -> {
```
ryanjbaxter, line 157: its not only that we are not providing one, its that we don't want one either right?

Would it make sense to just use `null`?  It almost seems like it being "unset" implies something is wrong, when in fact that is ok.

Or I guess another option we to set it to the empty string

ryanjbaxter (changes requested):

docs/src/main/asciidoc/property-source-config.adoc
```diff
@@ -409,6 +409,41 @@ will result in three properties being generated:
...
+              letter: c
+          - labels:
+              letter: d
+            useNameAsPrefix: true
+          - name: my-secret
+----
+====
+
+The same processing rules apply when generating property source as for config maps. The only difference is that
+potentially, looking up secrets by labels can mean that we find more than one source. In such a case prefix (if specified via `useNameAsPrefix`)
```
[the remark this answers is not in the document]
wind57: right. we already have a defect opened where people are requesting such functionality, and I will be adding it soon. It's on my TODO list

spring-cloud-kubernetes-client-config/src/main/java/org/springframework/cloud/kubernetes/client/config/LabeledSecretContextToSourceDataProvider.java
```diff
@@ -77,10 +81,43 @@ public KubernetesClientContextToSourceData get() {
 						createLabelsSelector(labels), null, null, null, null, null).getItems();
 
 				if (!secrets.isEmpty()) {
-					sourceName = secrets.stream().map(V1Secret::getMetadata).map(V1ObjectMeta::getName)
+
+					for (V1Secret secret : secrets) {
```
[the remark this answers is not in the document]
wind57: I have no idea how you do it, but you always ping places where there is a bug. this is awesome! I have realized that I have a small issue here: since the order is undefined, we still need a defined order for the prefix, otherwise users will not be able to wire the correct property all the time...

I've added `sorted` when computing the prefix where we find multiple secrets. For example, if based on label: `{color:blue}` we find two secrets `secret-a` and `secret-b`, prefix is going to be `secret-a.secret-b` (sorted names).

thank you for bringing this up.

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/ConfigUtils.java
wind57: > its not only that we are not providing one, its that we don't want one either right?

entirely correct. 

I _think_ this is how you see it, right?

```
/**
 * prefix has not been provided.
 */
public static final Prefix UNSET = new Prefix(() -> "");
```

If so, this is a very valid point, agreed.

spring-cloud-kubernetes-client-config/src/main/java/org/springframework/cloud/kubernetes/client/config/LabeledSecretContextToSourceDataProvider.java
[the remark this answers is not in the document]
ryanjbaxter: HAHAHA you are giving more credit than I deserve!

This is still probably worth adding to the docs IMO

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/ConfigUtils.java
ryanjbaxter: Yes but maybe its best to name it `DEFAULT`?

ryanjbaxter (changes requested):

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/NamedSecretNormalizedSource.java
wind57: done

spring-cloud-kubernetes-client-config/src/main/java/org/springframework/cloud/kubernetes/client/config/LabeledSecretContextToSourceDataProvider.java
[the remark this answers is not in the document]
wind57: done.

PR #1020, opened 2022-05-25
fix

wind57:


wind57:
@ryanjbaxter  minor thing... should be obvious from the PR what is going on

PR #1044, opened 2022-07-13
remove spotify docker plugin

wind57:


wind57:
I am not sure if this is used or not internally Ryan, but if you say it's not - we should remove it. wdyt?

ryanjbaxter:
It is used internally, that is how we push the image to dockerhub

PR #1045, opened 2022-07-13
bump fabric8 version

wind57:


wind57:
not a big deal, but worth upgrading, imho.

PR #1049, opened 2022-07-15
use branch name in the cache key

wind57:


.github/workflows/maven.yaml
```diff
@@ -21,13 +21,31 @@ jobs:
         with:
           distribution: 'temurin'
           java-version: '17'
+
+      - name: Extract branch name
```
wind57, line 25: we do the same thing in circleci and this fixes the build failure we have. to be fair, I am not yet 100% sure exactly why it happens to begin with, and of course I can't reproduce it locally (yet). in the meanwhile, this fixes the problem...

wind57:
@ryanjbaxter small PR that should address the build failure. next one I will look at making parallel jobs for github

PR #1052, opened 2022-07-18
proper order fix

wind57:


spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/ConfigUtils.java
```diff
@@ -165,25 +165,32 @@ public static String sourceName(String target, String applicationName, String na
...
-					// see if data is a single yaml/properties file and if it needs
-					// decoding
-					Map<String, String> rawData = foundSource.data();
-					if (decode) {
-						rawData = decodeData(rawData);
-					}
-					data.putAll(SourceDataEntriesProcessor.processAllEntries(rawData == null ? Map.of() : rawData,
-							environment));
-				});
+		// this is an ordered stream, and it means that non-profile based sources will be
```
wind57, line 176: this is the idea of this PR. In plain Spring, every if we have `application.yaml` and `application-dev.yaml`, properties from the latter will override the ones from the former. I think that users using our project would expect the same: `configmap-dev` will override properties from `configmap`. Me, personally, would expect that. 

This PR enforces that order, also adds tests to make sure this is respected.

ryanjbaxter (approved):

wind57:
merged latest `main` into this one, ready to go.

PR #1054, opened 2022-07-26
fix github actions build failure

wind57:


.github/workflows/maven.yaml
```diff
@@ -188,11 +184,17 @@ jobs:
 
           diff=$((right_bound - left_bound))
           sliced_array=("${CLASSNAMES[@]:$left_bound:$diff}")
-          TEST_ARG=$(echo $sliced_array | sed 's/ /,/g')
+          TEST_ARG=$(echo ${sliced_array[@]} | sed 's/ /,/g')
```
wind57, line 187: I had a bug here, I was only running the first test from a `sliced_array`, fixed.

wind57:
just found one more problem here with `abstract` classes... it is also present in circleci, so need to fix that also.

spring-cloud-kubernetes-client-config/src/test/java/org/springframework/cloud/kubernetes/client/config/KubernetesClientConfigMapIncludeProfileSpecificSourcesTests.java
```diff
@@ -1,107 +0,0 @@
-/*
- * Copyright 2013-2020 the original author or authors.
```
wind57, line 2: these two files were not used, I noticed this by accident

.github/workflows/maven.yaml
```diff
@@ -118,26 +113,63 @@ jobs:
...
-          				 | sed 's@/@.@g' \
-          				 | sed 's/.\{5\}$//' \
-                         | sort \
-          				 | tr '\n' ' '))
-
-          number_of_tests=${#CLASSNAMES[@]}
+          # - find all tests
+          # - exclude Fabric8IstionIT
+          # - only take classes that have @Test inside them
+          # - ignore the ones that have 'abstract class'. we do this because otherwise we would pass
```
wind57, line 119: this small comments explain the problem, I noticed this by accident and could reproduce locally too... thought appropriate to fix it in this PR

.circleci/config.yml
```diff
@@ -43,6 +43,11 @@ jobs:
             # needed for .withReuse(true) to work
             echo "testcontainers.reuse.enable=true" > ~/.testcontainers.properties
       - checkout
+      - restore_cache:
```
wind57, line 46: restore cache for tests

wind57:
this get funner and funner, it seems with 653 tests we are hitting [this issue](https://github.com/CircleCI-Public/circleci-cli/pull/441)

wind57:
I'm sorry for spamming with so many commits, but circleci test splitting does not have support for proper debugging, so this was my only hope. It seems that I have fixed the issue within circleci (I'll be sure when the build passes), after I confirm this I'll move to fix the github actions too.

.circleci/config.yml
```diff
@@ -60,19 +65,74 @@ jobs:
...
+              replaced=$(echo ${DERIVED_FROM_ABSTRACT_CLASSES_COMMAND/replace_me/"$class_name"})
+              result=($(eval $replaced))
+              PLAIN_TEST_CLASSNAMES+=(${result[@]})
+            done < <(eval $ABSTRACT_TEST_CLASSNAMES_COMMAND)
+
+            IFS=$'\n'
+            SORTED_TEST_CLASSNAMES=( $(sort \<<< "${PLAIN_TEST_CLASSNAMES[*]}" | uniq -u) )
+            unset IFS
+
+            printf "%s\n" "${SORTED_TEST_CLASSNAMES[@]}" > file.txt
```
wind57, line 127: I have not found a way to make circleci happy, without writing tests to a file first. At least it works like this

.github/workflows/maven.yaml
```diff
@@ -15,44 +15,44 @@ jobs:
...
         name: cache hit callback
         continue-on-error: false
         run: echo 'cache hit'
 
       - name: Build with skip tests
-        run: ./mvnw -s .settings.xml clean install -U -DskipTests
+        run: |
+          ./mvnw -s .settings.xml clean install -U -DskipTests
+
+          # build a lightweight project so that dependencies for the
```
wind57, line 50: this is important, and should make that timeout go away

.github/workflows/maven.yaml
```diff
@@ -104,40 +107,80 @@ jobs:
           distribution: 'temurin'
           java-version: '17'
 
+      - name: set branch name env variable
```
wind57, line 110: we repeat these steps, "set env variable", "checkout", "download cache". Unfortunately the closest thing to make these steps "workflow-wide" accessible is called "composite actions", but what we want is not supported. So, we will live with repeating the same thing 3 times.

wind57:
Finally. So Ive refactored and fixed a few things here and there. I still would like to understand why github is so slow on some things ( it does not have support for shared directory, so once build, images need to be uploaded in order to be shared across steps ) + tests are really imbalanced, some matrix step takes 7 min, some 20. And of course what happens to jacoco here? Some locking? Why does it block builds this way? I'll try to figure all this out.

spring-cloud-kubernetes-integration-tests/spring-cloud-kubernetes-fabric8-istio-it/pom.xml
```diff
@@ -61,6 +61,7 @@
 				<configuration>
 					<imageName>docker.io/springcloud/${project.artifactId}:${project.version}</imageName>
 					<imageBuilder>paketobuildpacks/builder</imageBuilder>
+					<skip>${skip.build.image}</skip>
```
wind57, line 64: I am skipping images builds for some steps, so need this one across every project that has image builds

.github/workflows/composites/upload-docker-images/action.yaml
```diff
@@ -0,0 +1,15 @@
+# we are not going to use upload/download actions, since it is slower than cache usage
```
wind57, line 1: I am using github composite actions, as it keeps the code much more understandable and manageable.

.github/workflows/composites/load-docker-images/action.yaml
```diff
@@ -0,0 +1,27 @@
+name: install docker images
+description: load docker images from tar
+runs:
+  using: "composite"
+  steps:
+    - name: load docker images
+      shell: bash
+      run: |
+        docker load -i /tmp/docker/images/spring-cloud-kubernetes-client-config-it.tar
```
wind57, line 9: this looks scary and I thought initially a a simple fact: why load every image in all steps, if I know that some steps either do not requires any images at all (there are no integration tests in this index) or they might need 1 or 2 images at all. But it takes only 1 minute to load, so far too little to invest any additional time here, imho

.github/workflows/composites/upload-docker-images/action.yaml
```diff
@@ -0,0 +1,15 @@
...
+# as key to the cache we are going to use DOCKER_IMAGES_KEY, which is the same as GITHUB_RUN_ID
+
+name: upload docker images
+description: upload docker images
+runs:
+  using: "composite"
+  steps:
```
wind57, line 8: circleci, unlike github actions, has a common path for the entire workflow, so for example: 

- I can put images into `/tmp/docker/images` in one step
- reference these images via `docker load -i` in any step I would like

github actions does not have such support, they recommend to use upload/download actions. It turns out that uploading our images, currently takes 15 minutes, far too much.

So instead, I am using cache with a slight hack - using `DOCKER_IMAGES_KEY` (`echo "DOCKER_IMAGES_KEY=$(echo $GITHUB_RUN_ID)" >> $GITHUB_ENV`) that I set in `env-variables` composite action and use that. Since  `GITHUB_RUN_ID` is unique per workflow, this achieves the "upload/download" without respective actions, but using `cache` action instead. It shrinks this step to 1 minute.

wind57:
@ryanjbaxter this is ready to review. thank you

wind57:
I've merged latest main also.

spring-cloud-kubernetes-integration-tests/spring-cloud-kubernetes-fabric8-istio-it/pom.xml
ryanjbaxter: Because the image is already built in the cache?

ryanjbaxter (changes requested):
Not sure I grok everything here, but it does look cleaner with the actions.

My main concern is what needs to be done when we add a new integration test?  What files need to be modified?  Do certain things need to be in the POM?  Seems like a README with details on how to structure the integration test and what other files need to be updated to ensure it runs properly would be good.

spring-cloud-kubernetes-integration-tests/spring-cloud-kubernetes-fabric8-istio-it/pom.xml
wind57: So first I do:

```
- name: build with skip tests and skip images
run: ./mvnw -T 1C -s .settings.xml clean install -U -DskipTests -Dskip.build.image=true
```
build entire project, but with `-DskipTests -Dskip.build.image=true`, thus no images anywhere yet.

then, I do:

```
- name: build controllers project
uses:  ./.github/workflows/composites/build-controllers-project
```

that does:

```
cd spring-cloud-kubernetes-controllers
.././mvnw -T 1C clean install
cd ..
```

As such, this builds the controllers images. 

Same thing in the next step for integration projects:

```
cd spring-cloud-kubernetes-integration-tests
# build the images, but dont run the tests
.././mvnw -T 1C clean install -DskipTests
 cd ..
```

As a result, we have all images built. Then I tar images:


```
        mkdir -p /tmp/docker/images
        TAG=$(./mvnw help:evaluate -Dexpression=project.version -q -DforceStdout)
        cd spring-cloud-kubernetes-integration-tests
        while read integ_test; do
          docker save -o /tmp/docker/images/${integ_test}.tar docker.io/springcloud/${integ_test}:$TAG
        done < <(mvn -Dexec.executable='echo' -Dexec.args='${project.artifactId}' exec:exec -q | grep -v 'spring-cloud-kubernetes-integration-tests')
        cd ..
```

This is a while loop, that in plain english does: "get all modules that are present in `spring-cloud-kubernetes-integration-tests`, but skip `spring-cloud-kubernetes-integration-tests` itself. Then for each entry tar it and save that tar in /tmp/docker/images". 

Only now, I can actually create a "cache per workflow" to share these images with other steps. 

---
So to answer your question (I hope it makes sense), I skip building images in some initial steps on purpose. If some plain test fails, it does not matter if I build the image or not - the build overall will fail, so why build it pre-maturely? Especially, since it takes so much time. So this skipping acts as an optimization and as an aggregate step, of building images for a subsequent step (cache per workflow).

wind57:
> My main concern is what needs to be done when we add a new integration test? What files need to be modified? Do certain things need to be in the POM? Seems like a README with details on how to structure the integration test and what other files need to be updated to ensure it runs properly would be good.

Actually you don't need to do anything, nada. :) just follow the convention of the other integration tests... But overall may be you are right, I'll add a README in this directory.

.github/workflows/composites/load-docker-images/action.yaml
ryanjbaxter: What about here?  If I add a new integration tests wouldnt I need to add the docker image here?

spring-cloud-kubernetes-integration-tests/spring-cloud-kubernetes-fabric8-istio-it/pom.xml
ryanjbaxter: makes sense

.github/workflows/composites/load-docker-images/action.yaml
wind57: OMG!!! I am an idiot, really sorry. This was never intended to load images like this, but instead iterate and find proper names. This gain proves my theory that you have an eye for where I am minor issues in my PRs. I love it!

spring-cloud-kubernetes-integration-tests/spring-cloud-kubernetes-fabric8-istio-it/pom.xml
wind57: as soon as I added the README, build started failing - it turns out it passed until now, only by pure luck. trying to understand how to fix this in github actions

wind57:
I am still in the blur what is going on here. The test indeed fails (`ActuatorRefreshKafkaIT`), for me locally too, so I started to debug it a bit. The test itself does a few things:

- starts configuration-watcher 
- starts configuration-watcher-it (the test itself)
- starts an instance of kafka

So we start the application, then create a configmap with the same name and a proper label, this should trigger the informer (id does), in turn this will call `applicationEventPublisher::publishEvent` that is of type `RefreshRemoteApplicationEvent` (this happens too). In my understanding, at this point in time, I should see a message in `springCloudBus` topic, but when I have a lister: 

```
bin/kafka-console-consumer.sh --topic springCloudBus --from-beginning --bootstrap-server localhost:9092
```

I don't see anything there. Weird. Still trying to figure out what is going on

spring-cloud-kubernetes-integration-tests/spring-cloud-kubernetes-configuration-watcher-it/src/test/java/org/springframework/cloud/kubernetes/configuration/watcher/ActuatorRefreshKafkaIT.java
```diff
@@ -140,6 +141,8 @@ void after() throws Exception {
 		k8SUtils.waitForDeploymentToBeDeleted(SPRING_CLOUD_K8S_CONFIG_WATCHER_IT_DEPLOYMENT_NAME, NAMESPACE);
 	}
 
+	// TODO figure out why this one fails on bus-starter-4.0.0-SNAPSHOT
+	@Disabled
```
wind57, line 145: I am disabling it for now. I've spent a few good days trying to figure out what is going on here, but my limited understanding of bus/binders/functions/kafka/etc did not help. I'll try to spent some more time in the future, as I don't like disabling this at all :(

ryanjbaxter: So you can reproduce this locally?  What do I need to do to reproduce it?

wind57: It's going to be a bit of mouthful to explain how to reproduce this locally, but in the meanwhile I am on to something (I think). May be you know this off the top of your head.

So in `2.1.x`we get a Bean from `BusRefreshAutoConfiguration` called : `RefreshListener`:


```
    @Bean
    @ConditionalOnProperty(
        value = {"spring.cloud.bus.refresh.enabled"},
        matchIfMissing = true
    )
    @ConditionalOnBean({ContextRefresher.class})
    public RefreshListener refreshListener(ContextRefresher contextRefresher, ServiceMatcher serviceMatcher) {
        return new RefreshListener(contextRefresher, serviceMatcher);
    }
```

After this one is invoked, the message ends up in the kafka queue.

What is interested is that this Bean is present in `main` - I can get it from the context, but it's not present in the "applicationListeners" of the app. It's almost like there should be some code that registers it, but it's not. I'm sorry if this is obscure explanation, but I am still trying to figure out...

ryanjbaxter: Weird, the autoconfiguration runs in main though?

wind57: right, it does:

```
2022-08-09T18:25:56.563+03:00 TRACE 14798 --- [           main] o.s.b.a.condition.OnPropertyCondition    : Condition OnPropertyCondition on org.springframework.cloud.bus.BusRefreshAutoConfiguration#refreshListener matched due to @ConditionalOnProperty (spring.cloud.bus.refresh.enabled) matched
```

when I start the app from `main`

wind57: Exactly. It is present as a Bean, but never called. This has to be either a race somewhere or a recent change in boot? snapshots, because this test used to work, in this PR.

ryanjbaxter: OK if you can recreate something that reproduces the problem fairly easily I can track down the right people to figure it out.

wind57: Let me show you how I reproduce it and let's see if it works.

- start zookeeper locally: 

```
docker run --rm -it -p 2181:2181 --name=zookeeper -e ZOOKEEPER_CLIENT_PORT=2181 -e ZOOKEEPER_ID=1 -e ZOOKEEPER_SERVER_1=zookeeper zookeeper:3.6.2
```

- find it's IP:

```
docker inspect zookeeper --format='{{ .NetworkSettings.IPAddress }}'
```

- start kafka locally:

```
docker run --rm -it -p 9092:9092 -e KAFKA_ADVERTISED_PORT=9092 -e KAFKA_ADVERTISED_HOST_NAME=localhost -e KAFKA_ZOOKEEPER_CONNECT='172.17.0.2:2181'  wurstmeister/kafka:2.13-2.6.0
```

`172.17.0.2` comes from the previous command.

- start a local kind cluster:

```
kind create cluster
```

- build `spring-cloud-kubernetes-commons` project:

```
cd spring-cloud-kubernetes-commons/
mvn clean install -Dskip.build.image=true -DskipTests -U
```

- build `spring-cloud-kubernetes-client-config`:

```
cd spring-cloud-kubernetes-client-config
mvn clean install -Dskip.build.image=true -DskipTests -U
```

Everything is ready so that our app can start.  Run `ConfigMapWatcherChangeDetector` (from `spring-cloud-kubernetes-controllers/spring-cloud-kubernetes-configuration-watcher`) with 3 env variables:

```
SPRING_MAIN_CLOUDPLATFORM=kubernetes;SPRING_PROFILES_ACTIVE=bus-kafka;SPRING_CLOUD_KUBERNETES_CLIENT_NAMESPACE=default
```

App is now ready to react to changes, so let's create a configmap: 

```
apiVersion: v1
data:
    key: value
kind: ConfigMap
metadata:
  name: spring-cloud-kubernetes-configuration-watcher
  labels:
    spring.cloud.kubernetes.config: "true"
```

This in turn, should trigger a kafka message in `springCloudBus` topic. To see if it's really there:

```
docker ps
// get the hash of the kafka image
docker exec -it <PREVIOUS_HASH> bash
/op/kafka/bin/kafka-console-consumer.sh --topic springCloudBus --from-beginning --bootstrap-server localhost:9092
```

wind57: do you think its safe to merge this the way it is and to track the test failure in an individual issue? this PR was supposed to be much simpler and grew to things I never expected. :)

ryanjbaxter: Im just looking now and noticing our jenkins builds are failing as well.  We are going to have to look into this some more.

wind57: if you find out what is going on, please ping me also. I am dying to find out

ryanjbaxter: I'm confused by your example here to reproduce the problem.  The name of the configmap is `spring-cloud-kubernetes-configuration-watcher` wouldn't that then target a refresh of the config watcher itself?  Shoudn't we be targeting a refresh of another app?

What I see happening in your example is that because the rerfesh is for the config watcher itself it shuts down the configwatcher context.  That is not what is happening in the integration test.

wind57: My bad! Its an incorrect name, it must have the `-it` at the end: `spring-cloud-kubernetes-configuration-watcher-it`. Sorry, I had too many changes in motion.

wind57: I edited the comment. Sorry again.

wind57: Btw, it does not even matter if that name is wrong. You would still not see a message in the kafka queue, and that is the main problem. If you look at my example, I do not even have a consumer ( there is no `spring-cloud-kubernetes-configuration-watcher-it` deployed ).

wind57: I also do not know if you noticed, but the problem is for sure in a SNAPSHOT version. While this test was failing, the circleci one was passing - and we had a bug there where the cache was not updating with the latest snapshots.

ryanjbaxter: I was talking to some people yesterday and there were some changes in Kafka recently that seems to correlate with the timeframe in which we saw these failures.  I am looking into it some more.  Hopefully will have an update next week.

wind57: Thank you. Its actually good that we started this task, we found out about at least 3 issues, which I will fix.

ryanjbaxter: OK so I finally nailed down that was happening after I was able to reproduce the problem using a simple Spring Cloud Bus + Kafka app.  This exception was being thrown when we called `triggerRefresh`

```
java.lang.NoSuchMethodError: 'org.springframework.util.concurrent.ListenableFuture org.springframework.kafka.core.KafkaTemplate.send(org.apache.kafka.clients.producer.ProducerRecord)'
	at org.springframework.integration.kafka.outbound.KafkaProducerMessageHandler.handleRequestMessage(KafkaProducerMessageHandler.java:513) ~[spring-integration-kafka-6.0.0-M4.jar:6.0.0-M4]
```
The reason why we didn't see it is because the ExecutorService was swallowing it 🤦‍♂️.  

I added a try/catch with a log statement around that call so we can catch anything like this in the future.

https://github.com/spring-cloud/spring-cloud-kubernetes/blob/17b61c22d32d51b7c554060f179a1447630414a4/spring-cloud-kubernetes-controllers/spring-cloud-kubernetes-configuration-watcher/src/main/java/org/springframework/cloud/kubernetes/configuration/watcher/ConfigMapWatcherChangeDetector.java#L69

The reason for the NoSuchMethodError was because there was a chance in Spring Core to `ListenableFuture` but the Spring Boot snapshots were using Spring Integration Milestone 4 which didn't have the corresponding change to support the change to `ListenableFuture`.  I got the Boot team to move Spring Integration to snapshots and that fixed the underlying problem.

So we should be able to remove the `@Ignore` from the test now and that build should pass.

wind57: NICE!! Indeed it builds just fine now :) take a look when you have time

PR #1057, opened 2022-08-09
Fix circleci flow

wind57:


wind57:
When `ActuatorRefreshKafkaIT` started failing in github workflow, I could not understand why does it passes in circleci? It meant two things:

- circleci was caching SNAPSHOT dependencies, and we were never properly using snapshots. This was indeed the case, and while looking more closely and debugging a bit, I found out that the cache we were using was never being updated with latest snapshots. I fixed that in this PR. 
- and second, it meant that `ActuatorRefreshKafkaIT` was broken because of a recent snapshot version of some library: the test was passing in circleci (where we had an older snapshot version and we were never using the newest one), and was failing in github workflow - where we were using the most recent snapshot version.

This PR also cleans up a bit the pipeline making it more readable. In essence, the flow/logic stays the same; but it's more readable now, imho. 

And of course now, because we fixed this issue, `ActuatorRefreshKafkaIT` fails - which is normal and expected. When we figure out what is going on and fix the underlying problem, a rebuild here will pass the test. 

---

This PR can go in, only after the github actions one does.

.circleci/config.yml
```diff
@@ -29,35 +143,35 @@ workflows:
...
             - fabric8_istio
 jobs:
+
+  build:
+    machine:
+      image: ubuntu-2204:2022.04.1
+    environment:
+      _JAVA_OPTIONS: "-Xms2g -Xmx2g"
+    steps:
+      - restore_main_cache
```
wind57, line 153: there are _two_ caches here involved. the first one is the old `.m2` cache you are thinking of and we save it and restore it in the usual way. The second one is a cache with latest dependencies, and its key is `spring-cloud-kubernetes-per-run-based-cache-{{ .Environment.CIRCLE_WORKFLOW_ID }}` - thus unique per this run. When I need the cache to be restored in some step, I always restore this one, with latest snapshots. 

This is why that `ActuatorRefreshKafkaIT` was not failing in circleci - which was a problem/bug.

.circleci/config.yml
```diff
@@ -29,35 +143,35 @@ workflows:
...
+  fabric8_istio:
+    machine:
+      image: ubuntu-2204:2022.04.1
+    steps:
+      - checkout-code-then-restore-cache-then-attach-workspace
+      - build-fabric8-istio-integration-test
+
   test:
-    parallelism: 5 # parallel containers to split the tests among
+    parallelism: 10 # parallel containers to split the tests among
```
wind57, line 166: I've changed this also, because we have fixed the incorrect caching problem - it brought an additional time to the build. Adding 5 more instances, shaves around 8-10 min from the build.

wind57:
@ryanjbaxter  this is the next one in the list for you to review when you get a chance. thank you

PR #1070, opened 2022-09-07
change switch to jdk-17

wind57:


spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/reload/ConfigReloadAutoConfiguration.java
```diff
@@ -64,22 +64,20 @@ public TaskSchedulerWrapper<TaskScheduler> taskScheduler() {
 	public ConfigurationUpdateStrategy configurationUpdateStrategy(ConfigReloadProperties properties,
 			ConfigurableApplicationContext ctx, Optional<RestartEndpoint> restarter, ContextRefresher refresher) {
 		String strategyName = properties.getStrategy().name();
-		switch (properties.getStrategy()) {
```
wind57, line 67: this simply changes to the new way switch statements work in java-17. This PR must wait for the other one in review - because it fixes a test failure

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/reload/ConfigReloadAutoConfiguration.java
```diff
@@ -64,22 +64,20 @@ public TaskSchedulerWrapper<TaskScheduler> taskScheduler() {
...
+					restarter.get().restart();
+				});
+			}
+			case REFRESH -> new ConfigurationUpdateStrategy(strategyName, refresher::refresh);
+			case SHUTDOWN -> new ConfigurationUpdateStrategy(strategyName, () -> {
 				wait(properties);
 				ctx.close();
 			});
-		}
-		throw new IllegalStateException("Unsupported configuration update strategy: " + strategyName);
```
wind57, line 82: in general, you do not need a default case if all the enum values are covered. We throw here an `IllegalStateException` and say what `strategyName` to fail with. If we remove this line, we could, in theory, get an `IncompatibleClassChangeError` (that is the `default` that `javac` will add).

In our case, that is simply impossible, because `ReloadStrategy` is in the same maven module as this class, and you can't compile `ReloadStrategy` as `version-1`, then compile this class with the same `version-1`; only after to compile `ReloadStrategy` with `version-2` and and run this class. This simply makes no sense, as such drop this line over here

wind57:
@ryanjbaxter if you are comfortable with this PR, for example, then we can merge it - and it should fix the build failure. wdyt?

ryanjbaxter:
I dont have any problem with this change, but what build failure are you addressing?  This build just passed
https://github.com/spring-cloud/spring-cloud-kubernetes/actions/runs/3023390649

wind57:
The one from the previous merge you made - main branch, that is.

wind57:
I see, you re-triggered. sorry, was not aware.

ryanjbaxter:
So this is not fixing any build failure then?

wind57:
Right.

PR #1071, opened 2022-09-07
fix skip image place

wind57:


spring-cloud-kubernetes-controllers/spring-cloud-kubernetes-configserver/pom.xml
```diff
@@ -66,10 +66,12 @@
 						<name>${env.IMAGE}</name>
 					</image>
 					<goal>build-image</goal>
-					<skip>${skip.build.image}</skip>
 				</configuration>
 				<executions>
 					<execution>
+						<configuration>
```
wind57, line 72: this is the only change in this PR. I've realized that the place where we say "skip image building" is not correct. It does not impact us, but of course this clearly declares the intention.

wind57:
@ryanjbaxter ready to be looked at too. thank you

PR #1072, opened 2022-09-08
remove jib from controllers + some old kind related stuff

wind57:


wind57:
@ryanjbaxter last one before that fix...

PR #1075, opened 2022-09-15
allow to repackage when image is not build

wind57:


spring-cloud-kubernetes-controllers/spring-cloud-kubernetes-configserver/pom.xml
```diff
@@ -75,6 +76,13 @@
 							<goal>build-image</goal>
 						</goals>
 					</execution>
+					<execution>
```
wind57, line 79: while working on a different fix, I've realized that that my previous fix did not fix much :( if we disable a single execution of a plugin, we disable the entire plugin.

Instead I want a very different thing: I want to be able to say : "don't build an image, but still build a Spring jar" and this set-up achieves that. Yeah, it's my bug, for sure.

Again, it dose not affect us, but it's incorrect nevertheless.

wind57:
@ryanjbaxter brain fart from my side in one of the previous PRs...

ryanjbaxter (approved):

wind57:
I'm checking the build failure on `main`...

PR #1076, opened 2022-09-15
fix build failure

wind57:


wind57:
@ryanjbaxter I guess the problem was fixed - and this PR while it does not fix anything per-se, is still OK to be merged, imho, as it simplifies a tiny bit things. thank you

PR #1081, opened 2022-09-23
Add possibility to disable configmap creation

ugwun:
As configmaps are sometimes managed by different parties (in our case it is a Helmchart that takes care of creating and destroying configmaps), it would be helpful to have a possibility to disable leader configmap creation on the library's side.

### DIFF
--- a/spring-cloud-kubernetes-fabric8-leader/src/test/java/org/springframework/cloud/kubernetes/fabric8/leader/Fabric8LeadershipControllerTest.java
+++ b/spring-cloud-kubernetes-fabric8-leader/src/test/java/org/springframework/cloud/kubernetes/fabric8/leader/Fabric8LeadershipControllerTest.java
@@ -16,18 +16,32 @@
 
 package org.springframework.cloud.kubernetes.fabric8.leader;
 
+import ch.qos.logback.classic.Level;
+import ch.qos.logback.classic.Logger;
+import ch.qos.logback.classic.spi.ILoggingEvent;
+import ch.qos.logback.core.read.ListAppender;
 import io.fabric8.kubernetes.client.KubernetesClient;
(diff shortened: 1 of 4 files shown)


ryanjbaxter (changes requested):
Can you provide a test?

ugwun:
I added unit test org.springframework.cloud.kubernetes.fabric8.leader.Fabric8LeadershipControllerTest#whenNonExistentConfigmapAndCreationNotAllowedStopLeadershipAcquire

ryanjbaxter:
Thanks, it would be good to add some documentation about the property as well.

spring-cloud-kubernetes-fabric8-leader/src/test/java/org/springframework/cloud/kubernetes/fabric8/leader/Fabric8LeadershipControllerTest.java
```diff
@@ -60,4 +74,55 @@ public void shouldGetEmptyLocalLeader() {
 		assertThat(this.fabric8LeadershipController.getLocalLeader().isPresent()).isFalse();
 	}
 
+	@Test
+	void whenNonExistentConfigmapAndCreationNotAllowedStopLeadershipAcquire() {
+
+		class MemoryAppender extends ListAppender<ILoggingEvent> {
```
wind57, line 80: so this is supposed to capture a log and assert it, right? If so, can't we use `OutputCaptureExtension' here? wdyt?

ugwun: I didn't know this possibility existed. I updated the code with OutputCaptureExtension. Thanks.

ugwun:
@ryanjbaxter I added documentation to `_configprops.adoc`

PR #1082, opened 2022-09-23
clean-up in discovery implementation - part 1

wind57:


spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/discovery/KubernetesDiscoveryConstants.java
```diff
@@ -0,0 +1,50 @@
...
+ */
+
+package org.springframework.cloud.kubernetes.commons.discovery;
+
+/**
+ * Constants that are to be used across discovery implementations.
+ *
+ * @author wind57
+ */
+public final class KubernetesDiscoveryConstants {
```
wind57, line 24: reuse some constants across both implementations.

I will have more PRs like these, as I start to delve into discovery implementation...

wind57:
@ryanjbaxter thank you for the feedback on all other open questions, I will take a look at those. Can you look at this one in the meanwhile, seems like an easy one, imho.

PR #1083, opened 2022-09-24
simplify discovery starters

wind57:


spring-cloud-starter-kubernetes-client/pom.xml
```diff
@@ -28,19 +12,10 @@
 	<artifactId>spring-cloud-starter-kubernetes-client</artifactId>
 
 	<dependencies>
-		<dependency>
```
wind57, line 31: this will be more of a question here... `spring-cloud-kubernetes-commons` and `spring-cloud-kubernetes-client-autoconfig` are already transitive dependencies of `spring-cloud-kubernetes-client-discovery`. So should we still provide them? Seems like un-necessary?

wind57: @ryanjbaxter would you be able to take a look here and provide some feedback please? thank you

ryanjbaxter: Seems to make sense but I would just make sure by creating a simple project and make sure we aren't missing any dependencies.

wind57: hmm, but why? these starters are supposed to expose only the `spring-cloud-kubernetes-fabric8-discovery` - discovery functionality that is and all else must be an implementation detail. 

Unless you have something different in mind and I don't get the full picture :)

PR #766, opened 2021-03-22
Initialize DiscoveryClient

ryanjbaxter:
Fixes #517

PR #1088, opened 2022-09-26
clean-up in discovery implementation - part 2

wind57:


spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/discovery/KubernetesDiscoveryClientHealthIndicatorInitializer.java
```diff
@@ -16,15 +16,21 @@
...
+
 import org.springframework.cloud.client.discovery.event.InstanceRegisteredEvent;
 import org.springframework.cloud.kubernetes.commons.PodUtils;
 import org.springframework.context.ApplicationEventPublisher;
+import org.springframework.core.log.LogAccessor;
 
 /**
  * @author Ryan Baxter
  */
-public final class KubernetesDiscoveryClientHealthIndicatorInitializer implements InitializingBean {
```
wind57, line 27: few things happened here:

- I dropped `InitializingBean` in favor of `@PostConstruct` and also wrote a unite test for it. 
- I think that we should publish an event only when we are inside kubernetes, not always. So something like:

```
if (podUtils.isInsideKubernetes()) {
   // publish event
} else {
   // don't publish the event
}
```

- we used to publish the Supplier instead of the actual Pod `podUtils.currentPod()`, but needs to be `podUtils.currentPod().get()`.

spring-cloud-kubernetes-commons/src/test/java/org/springframework/cloud/kubernetes/commons/discovery/KubernetesDiscoveryClientHealthIndicatorInitializerTests.java
```diff
@@ -0,0 +1,131 @@
...
+
+/**
+ * @author wind57
+ *
+ * Tests the
+ * {@link org.springframework.cloud.kubernetes.commons.discovery.KubernetesDiscoveryClientHealthIndicatorInitializer}
+ * with the fabric8 client.
+ *
+ */
+class KubernetesDiscoveryClientHealthIndicatorInitializerTests {
```
wind57, line 42: a test that proves that when we are inside k8s, we publish the event and we do not otherwise

wind57:
@ryanjbaxter what do you think about this one?

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/discovery/KubernetesDiscoveryClientHealthIndicatorInitializer.java
ryanjbaxter: I actually would rather publish it when we are running outside of k8s as well.  I think people will be developing/testing outside of kubernetes and we should still publish this event.

wind57: good point. The configuration providing this bean is anyway annotated with `@ConditionalOnCloudPlatform`

wind57: actually, there was a test that failed after I made the change, and I don't think that it's correct to post an event when we are outside k8s.

The thing is we used to do : 

```
publishEvent(new InstanceRegisteredEvent<>(podUtils.currentPod(), null))
```

`podUtils.currentPod()` returns a `Supplier`, which is obviously a minor slip when the initial code was written.

So when we were outside the k8s we would post an event where the source was a `Supplier`. Now that we fixed the code, and do : `podUtils.currentPod().get()` - this returns a null when there is no pod, so `InstanceRegisteredEvent` takes a null as input - and fails; which is should. 

To me, this means that we should not sent the event while we are outside k8s, simply because there is no pod, as such no source for the event.

wdyt?

ryanjbaxter: I mean the point it really to initialize the health indicator properly so its status shows as `UP` otherwise it will be `UNKNOWN`.  I dont even think the Pod is really used for anything, its just the event that needs to be published.
https://github.com/spring-cloud/spring-cloud-commons/blob/main/spring-cloud-commons/src/main/java/org/springframework/cloud/client/discovery/health/DiscoveryClientHealthIndicator.java#L58

Basically we could probably publish the event without the pod if its null.

Also there are a few other implementations that repeat this pattern (my bad)
https://github.com/spring-cloud/spring-cloud-kubernetes/pull/766

wind57: I see, thank you for being patient with me!

ryanjbaxter: I'm the one who introduced the bug in the first place, so thank you!

wind57: ready. I've added a small record as a source

PR #1097, opened 2022-10-10
do not create properties for configmap/secrets when these are disabled (for configdata only cases)

wind57:


spring-cloud-kubernetes-fabric8-config/src/test/java/org/springframework/cloud/kubernetes/fabric8/config/locator_retry/fail_fast_enabled_retry_disabled/ConfigFailFastEnabledButRetryDisabled.java
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package org.springframework.cloud.kubernetes.fabric8.config.locator_retry;
+package org.springframework.cloud.kubernetes.fabric8.config.locator_retry.fail_fast_enabled_retry_disabled;
```
wind57, line 17: I will do this more, that is will move abstract class and its related bootstrap and configdata test classes to a common package. This is so much easier to navigate to when debugging code, imho. I've done that before in various places too

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/KubernetesConfigDataResource.java
```diff
@@ -62,14 +62,6 @@ public KubernetesClientProperties getProperties() {
 		return this.properties;
 	}
 
-	public ConfigMapConfigProperties getConfigMapProperties() {
```
wind57, line 65: these were not used, so I just decided to remove them

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/KubernetesConfigDataLocationResolver.java
```diff
@@ -188,12 +196,21 @@ protected PropertyHolder loadProperties(ConfigDataLocationResolverContext contex
 		}
 		kubernetesClientProperties.setNamespace(namespace);
 
-		ConfigMapConfigProperties configMapConfigProperties = binder
-				.bind(ConfigMapConfigProperties.PREFIX, ConfigMapConfigProperties.class)
-				.orElseGet(ConfigMapConfigProperties::new);
-		SecretsConfigProperties secretsProperties = binder
-				.bind(SecretsConfigProperties.PREFIX, SecretsConfigProperties.class)
-				.orElseGet(SecretsConfigProperties::new);
+		boolean configEnabled = binder.bind("spring.cloud.kubernetes.config.enabled", boolean.class).orElse(true);
```
wind57, line 199: this is the idea of this PR. I noticed this small thing while working on a different thing, really. 

Instead of providing the configmap and secrets properties all the time (even if configmap/secret could be disabled), we can conditionally provide them. If you think about it, if you have set : `spring.cloud.kubernetes.secrets.enabled=false`, then you obviously don't care about `ConfigMapConfigProperties`...

spring-cloud-kubernetes-fabric8-config/src/test/java/org/springframework/cloud/kubernetes/fabric8/config/locator_retry/fail_fast_enabled_retry_disabled/ConfigDataConfigFailFastEnabledButRetryDisabled.java
```diff
@@ -49,7 +51,10 @@ class ConfigDataConfigFailFastEnabledButRetryDisabled extends ConfigFailFastEnab
 	private static KubernetesClient mockClient;
 
 	@MockBean
-	KubernetesNamespaceProvider kubernetesNamespaceProvider;
+	private KubernetesNamespaceProvider kubernetesNamespaceProvider;
+
+	@SpyBean
```
wind57, line 56: we need this `@SpyBean` because the test itself has this:

```
@SpyBean
private Fabric8ConfigMapPropertySourceLocator propertySourceLocator;
```

and this one has a dependency on `ConfigMapConfigProperties` in its constructor, but since we have changed the code to not provide these properties if configmap is disabled (which we do in this test), we need to provide it.

wind57:
@ryanjbaxter this is ready to be looked at. It's a minor improvement, but needs your eyes.

PR #1105, opened 2022-10-13
move configmap source to record

wind57:


spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/ConfigMapConfigProperties.java
```diff
@@ -109,100 +109,16 @@ public List<NormalizedSource> determineSources(Environment environment) {
...
-		}
-
-		public void setLabels(Map<String, String> labels) {
-			this.labels = labels;
-		}
-
-		public boolean isEmpty() {
-			return !StringUtils.hasLength(this.name) && !StringUtils.hasLength(this.namespace);
-		}
+	public record Source(String name, String namespace, @DefaultValue Map<String, String> labels, String explicitPrefix,
```
wind57, line 120: same idea here, move to record

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/ConfigMapConfigProperties.java
```diff
@@ -235,23 +151,6 @@ private Stream<NormalizedSource> normalize(String defaultName, String defaultNam
 
 		}
 
-		@Override
```
wind57, line 238: now this in theory is dangerous, because equals/hashCode is based on different parameters in records, while we require only on two of them. It turns out that we never required equals/hashCode, because we get them all as a `List` and transform them into a `List<NormalizedSource>`. And later use wrap `List<NormalizedSource>` into a `Set<NormalizedSource>`, but now all implementations of `NormalizedSource` do have proper hashCode and equals on the fields that we care. So removing these two methods makes sense.

We also have unit tests that validate this behavior.

wind57:
@ryanjbaxter this is ready to be looked at. thank you.

ryanjbaxter (approved):

PR #1112, opened 2022-10-16
move subsets to a record + add default instance for metadata

wind57:


spring-cloud-kubernetes-fabric8-discovery/src/main/java/org/springframework/cloud/kubernetes/fabric8/discovery/EndpointSubsetNS.java
```diff
@@ -16,44 +16,16 @@
...
-
-	@Override
-	public boolean equals(Object o) {
-		return this.endpointSubset.equals(o);
-	}
-
-	@Override
-	public int hashCode() {
-		return this.endpointSubset.hashCode();
+record EndpointSubsetNS(String namespace, List<EndpointSubset> endpointSubset) {
```
wind57, line 26: I will sound like a broken record a little more, but this moves to a record.

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/discovery/KubernetesDiscoveryProperties.java
```diff
@@ -188,15 +187,10 @@ public record Metadata(@DefaultValue("true") boolean addLabels, String labelsPre
...
-		}
-
-		// needed in order to get the defaults for some fields
-		public Metadata() {
-			this(true, null, true, null, true, "port.");
-		}
+		/**
+		 * Default instance.
+		 */
+		public static final Metadata DEFAULT = new Metadata(true, null, true, null, true, "port.");
```
wind57, line 193: having a default instance, is easier to manage and understand and I did this in one of the PRs before.

spring-cloud-kubernetes-fabric8-discovery/src/main/java/org/springframework/cloud/kubernetes/fabric8/discovery/EndpointSubsetNS.java
wind57: it took a little bit of time until I felt comfortable with : 

```
endpointSubset = endpointSubset == null ? List.of() : endpointSubset;
```

reading the bytecode was the only way for me to "get it". In essence this can be read like this:

```
List<EndpointSubset> local = endpointSubset == null ? List.of() : endpointSubset;
this.endpointSubset = local;
```

`this.endpointSubset = local` will be inserted by the compiler in a compact constructor in a record. All fields in a record are "committed" in a compact constructor at the very end and this is done by `javac`. 

---

This is not very different from a plain class constructor, for example:


```
public class Mine {

	private final List<String> endpointSubset;

	public Mine(List<String> endpointSubset) {
		if (endpointSubset == null) {
			endpointSubset = List.of();
		}
		this.endpointSubset = endpointSubset;
	}
	
}
```

But it was confusing in the beginning.

ryanjbaxter (approved):

Issue #1000, opened 2022-04-04
Namespace filter for service discovery


vchegwidden:
This request is to add the ability to filter/limit namespaces to use for service discovery in the `DiscoveryClient` through a property

When working as a tenant on a shared cluster we don't have permissions to view resources in all namespaces. We also don't have permissions to create cluster roles and cluster rolebindings.
In these situations you can't use the `spring.cloud.kubernetes.discovery.all-namespaces=true` property since it will throw an exception when trying to get resources in namespaces you don't have access to.
Not having the option to list namespaces to take into consideration means you would be limited to a single namespace for all your deployments, which isn't great...

I've been trying to run Spring boot admin in a cluster where my access is limited to only the namespaces I own. When trying to use the `DiscoveryClient` and setting `all-namespaces=true` the following exception is thrown since the service account doesn't have the correct permissions

```
"stack_trace":"i.f.k.c.KubernetesClientException: Failure executing: GET at: https://<nodeIP>/api/v1/namespaces/some-namespace/services. 
    Message: Forbidden!Configured service account doesn't have access. Service account may have been revoked. services is forbidden: 
    User \"system:serviceaccount:sba:spring-boot-admin\" cannot list resource \"services\" in API group \"\" in the namespace \"some-namespace\".
```

The work-around I've used is to implement my own `DiscoveryClient` that takes a list of namespaces from a property, then overriding the behaviour of `getServices` and `getEndpointList` to only look in the namespaces supplied (Note I've only tested this with the fabric8 client)

It would however be great if this was included out of the box.

```
@Value("#{'${spring.cloud.kubernetes.discovery.filter-namespaces:}'.split(',')}")
private List<String> filteredNamespaces;

@Override
public List<String> getServices(Predicate<Service> filter) {
    if (filteredNamespaces.isEmpty()) {
        return super.getServices(filter);
    }

    List<String> services = new ArrayList<>();
    for (String ns : filteredNamespaces) {
        services.addAll(
            this.client.services()
                .inNamespace(ns).list().getItems().stream().filter(filter).map(s -> s.getMetadata().getName().toList());
    }
}

@Override
public List<Endpoints> getEndPointsList(String serviceId) {
    if (filteredNamespaces.isEmpty()) {
        return super.getEndpointList(serviceId);
    }

    List<Endpoints> endpoints = new ArrayList<>();
    for (String ns : filteredNamespaces) {
        endpoints.addAll(
            this.client.endpoints()
[...]

PR #1113, opened 2022-10-17
Namespace filter for service discovery (issue #1000). 

mbialkowski1:


spring-cloud-kubernetes-integration-tests/spring-cloud-kubernetes-fabric8-client-discovery/src/test/java/org/springframework/cloud/kubernetes/fabric8/configmap/Fabric8DiscoveryNamespaceFilterIT.java
```diff
@@ -0,0 +1,248 @@
...
+import org.springframework.cloud.kubernetes.integration.tests.commons.K8SUtils;
+import org.springframework.core.ParameterizedTypeReference;
+import org.springframework.http.HttpMethod;
+import org.springframework.http.client.reactive.ReactorClientHttpConnector;
+import org.springframework.web.reactive.function.client.WebClient;
+
+/**
+ * @author mbialkowski1
+ */
+class Fabric8DiscoveryNamespaceFilterIT {
```
wind57, line 54: this is fabulous! I'd like to ask for an integration test in k8s-client also, the reason for that is that eventually we will move away from fabric8 for discovery (just like we did for reload for example). again, this is awesome and thank you!

ryanjbaxter:
Done!

wind57:
I'l leaving a note here for myself mainly, we need this change in `KubernetesCatalogWatch` too. The fabric8 one + when we add a k8s-native implementation of catalog watcher, too.

mbialkowski1:
> @mbialkowski1 tell us when this is done and I will gladly take a look (most probably Ryan too) thank you

I think I'm done, I added an integration test for `org.springframework.cloud.kubernetes.discovery.KubernetesDiscoveryClient`.
I also found the problem in the method `org.springframework.cloud.kubernetes.client.discovery.KubernetesInformerDiscoveryClient#getInstances` 
We use only one service:
```
V1Service service = properties.allNamespaces() ? this.serviceLister.list().stream()
				.filter(svc -> serviceId.equals(svc.getMetadata().getName())).findFirst().orElse(null)
```

If a service with the same name is in several different namespaces then we lost information about it.
I corrected it, can you see if it is ok? (https://github.com/spring-cloud/spring-cloud-kubernetes/pull/1113/commits/fed3e4f74a581ad1455096a61849e76ba9974c26)
What else should I do/improve in this PR?

wind57:
I will take a look at this in the upcoming weeks, I just want to work (and clarify some things first), if you do not mind.

ryanjbaxter (commented):
Do we still retain the functionality that is `all-namespace` is false and no specific namespace is specified we still use the current namespace?

mbialkowski1:
> Do we still retain the functionality that is `all-namespace` is false and no specific namespace is specified we still use the current namespace?

Yes, it should work like that.

PR #1126, opened 2022-11-08
Add tests for KubernetesConfigDataLoader

wind57:


spring-cloud-kubernetes-commons/src/test/java/org/springframework/cloud/kubernetes/commons/config/KubernetesConfigDataLoaderTests.java
```diff
@@ -0,0 +1,144 @@
...
+import org.springframework.mock.env.MockEnvironment;
+import org.springframework.mock.env.MockPropertySource;
+
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+/**
+ * @author wind57
+ */
+class KubernetesConfigDataLoaderTests {
```
wind57, line 41: not a lot to see here, just added some basic tests for `KubernetesConfigDataLoader`. In one of my previous PRs, when I moved a class to a record, I had a typo when renaming a method; as a result an integration test was failing, what is worse: I lost around 1 day figuring out the problem. Then I said that I will add proper unit tests, so that we would not be in such a position again. I will add more tests like these...

PR #1142, opened 2022-11-19
Remove redundant dependencies

weihubeats:
There are two `spring-boot-starter-logging dependencies` in `spring-cloud-kubernetes-commons`，We could remove one

### DIFF
--- a/spring-cloud-kubernetes-commons/pom.xml
+++ b/spring-cloud-kubernetes-commons/pom.xml
@@ -63,7 +63,6 @@
 		<dependency>
 			<groupId>org.springframework.boot</groupId>
 			<artifactId>spring-boot-starter-logging</artifactId>
-			<optional>true</optional>
 		</dependency>
 		<dependency>
 			<groupId>org.springframework.retry</groupId>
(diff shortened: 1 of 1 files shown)


wind57:
shouldn't we drop the other one, that is `optional`? you are removing it from `commons`, so all project that use it should inherit it, while they in return should have it as `optional`. wdyt?

weihubeats:
Now that I've removed the optional dependency, do I need to add
```xml
                <dependency>
			<artifactId>spring-boot-starter-logging</artifactId>
			<optional>true</optional>
		</dependency>
```
 to other projects that depend on the `spring-cloud-kubernetes-commons` module?

wind57:
You will need to analyze each dependency individually and see if it makes sense. For example, `autoconfigure` dependency  depends on `commons` and is used in `config` and `discovery`... So it's not that trivial. For the time being, what you have here might be enough, as the purpose of this is to simply drop the duplicated dependency. Let's wait for the maintainers to see what they think.

weihubeats:
ok

wind57:
@ryanjbaxter can you glance over here? minor thing, but still worth it, imo. thank you

ryanjbaxter (approved):

PR #1147, opened 2022-11-24
Add integration test for fabric8 catalog watcher

wind57:

PR #1149, opened 2022-11-25
Use endpoint slices for fabric8 catalog watcher

wind57:


wind57:
only started work here. depends on [this](https://github.com/spring-cloud/spring-cloud-kubernetes/pull/1147)

spring-cloud-kubernetes-test-support/src/main/resources/setup/role.yaml
```diff
@@ -4,6 +4,6 @@ metadata:
   namespace: default
   name: namespace-reader
 rules:
-  - apiGroups: ["", "extensions", "apps"]
-    resources: ["configmaps", "pods", "services", "endpoints", "secrets"]
+  - apiGroups: ["", "extensions", "apps", "discovery.k8s.io"]
```
wind57, line 7: @ryanjbaxter before I even start documentation and explanation on my thought process in this PR, I need your supervisor advice. 

Kubernetes recommends using EndpointSlices instead of Endpoints (think of it a @Deprecated). Now, we can use EndpointSlices as this PR and integration tests prove. The problem is that by switching to EndpointSlices, users would need to change the roles, as this file shows.

So, one hand we have the official k8s recommendation, on the other hand if we do follow it, we kind of introduce a "breaking" change, as users need to adjust to it. 

Of course, we could let users opt-in into EndpointSlices and we could explain in the docs the steps needed for that, but then we would kind of go against what k8s says to do. 

As a side note, there is nothing wrong in using Endpoints, the code still works... wdyt is the correct path forward here? Thank you.

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/discovery/KubernetesDiscoveryProperties.java
```diff
@@ -51,14 +52,15 @@ public record KubernetesDiscoveryProperties(
 		@DefaultValue({"443", "8443"}) Set<Integer> knownSecurePorts,
 		@DefaultValue Map<String, String> serviceLabels, String primaryPortName,
 		@DefaultValue Metadata metadata,
-		@DefaultValue("" + DEFAULT_ORDER) int order) {
+		@DefaultValue("" + DEFAULT_ORDER) int order,
+		@DefaultValue("true") boolean useEndpointSlices) {
```
wind57, line 56: 1. we enable using `EndpointSlices` by default. Users can opt-out of this one, and use Endpoints only - for example if their cluster is on a version where `EndpointSlices` are not supported. 

 - Let's assume you start the app with a version of k8s where EndpointSlices are _not_ supported : we will default to using `Endpoints` in such case (even if `useEndpointSlices=true`).
 - If you start the app and EndpointSlices are supported, we will use it.

So the question might come up of why then do we need such a flag, if we will default anyway? Well, my thought process is this:

- for EndpointSlices to work, you need proper `Role` settings and that might not always be an option. So I thought that if changing the Role si not an option, you can disable EndpointSlices, even if the cluster supports them.

spring-cloud-kubernetes-fabric8-discovery/src/main/java/org/springframework/cloud/kubernetes/fabric8/discovery/KubernetesCatalogWatch.java
```diff
@@ -41,23 +40,23 @@
...
+	private static final String ENDPOINT_SLICE = "EndpointSlice";
 
-	private final KubernetesClient kubernetesClient;
+	private static final LogAccessor LOG = new LogAccessor(LogFactory.getLog(KubernetesCatalogWatch.class));
 
-	private final KubernetesDiscoveryProperties properties;
+	private final Fabric8CatalogWatchContext context;
 
-	private final KubernetesNamespaceProvider namespaceProvider;
+	private Function<Fabric8CatalogWatchContext, List<EndpointNameAndNamespace>> stateGenerator;
```
wind57, line 51: we separate the logic now into two logical pieces: into Endpoints and EndpointSlices. Each of those (`Fabric8EndpointsCatalogWatch` and `Fabric8EndpointSliceV1CatalogWatch`) is actually a `Function`. This function takes as input a `Fabric8CatalogWatchContext` (which has members as : `KubernetesClient`, `KubernetesDiscoveryProperties` and `KubernetesNamespaceProvider` ) and returns a `List<EndpointNameAndNamespace>`.

spring-cloud-kubernetes-fabric8-discovery/src/main/java/org/springframework/cloud/kubernetes/fabric8/discovery/KubernetesCatalogWatch.java
```diff
@@ -113,4 +82,34 @@ public void catalogServicesWatch() {
 		}
 	}
 
+	@PostConstruct
+	void postConstruct() {
+
+		if (context.properties().useEndpointSlices()) {
```
wind57, line 88: we need to make a distinction of when to use Endpoints vs EndpointSlices, because the latter might be not support on the k8s cluster. As such, we try to find out if it is supported or not and based on that return the proper `stateGenerator` instance

spring-cloud-kubernetes-fabric8-discovery/src/main/java/org/springframework/cloud/kubernetes/fabric8/discovery/Fabric8CatalogWatchContext.java
```diff
@@ -0,0 +1,45 @@
...
+import org.springframework.cloud.kubernetes.commons.discovery.EndpointNameAndNamespace;
+import org.springframework.cloud.kubernetes.commons.discovery.KubernetesDiscoveryProperties;
+
+/**
+ * A simple holder for some instances needed for either Endpoints or EndpointSlice catalog
+ * implementations.
+ *
+ * @author wind57
+ */
+record Fabric8CatalogWatchContext(KubernetesClient kubernetesClient, KubernetesDiscoveryProperties properties,
```
wind57, line 37: a record that acts as a Tuple for all the parameters needed in order to compute the `List<EndpointNameAndNamespace>`. Please notice both Endpoints and EndpointSlices implementations will delegate to its `state` method here. 

This method takes as input a `Stream<ObjectReference>` because both Endpoints and EndpointSlices can be narrowed to this, meaning : 

```
Stream<ObjectReference> references = endpointSlices.stream().map(EndpointSlice::getEndpoints)
				.flatMap(List::stream).map(Endpoint::getTargetRef);
```

and 

```
Stream<ObjectReference> references = endpoints.stream().map(Endpoints::getSubsets).filter(Objects::nonNull)
				.flatMap(List::stream).map(EndpointSubset::getAddresses).filter(Objects::nonNull).flatMap(List::stream)
				.map(EndpointAddress::getTargetRef);
```

I hope this makes sense.

spring-cloud-kubernetes-fabric8-discovery/src/main/java/org/springframework/cloud/kubernetes/fabric8/discovery/Fabric8EndpointSliceV1CatalogWatch.java
```diff
@@ -0,0 +1,73 @@
...
+import org.springframework.cloud.kubernetes.commons.discovery.EndpointNameAndNamespace;
+import org.springframework.cloud.kubernetes.fabric8.Fabric8Utils;
+import org.springframework.core.log.LogAccessor;
+
+/**
+ * Implementation that is based on EndpointSlice V1.
+ *
+ * @author wind57
+ */
+final class Fabric8EndpointSliceV1CatalogWatch
```
wind57, line 38: implementation that uses `EndpointSlices`

spring-cloud-kubernetes-fabric8-discovery/src/main/java/org/springframework/cloud/kubernetes/fabric8/discovery/Fabric8EndpointsCatalogWatch.java
```diff
@@ -0,0 +1,85 @@
...
+import org.springframework.cloud.kubernetes.commons.discovery.EndpointNameAndNamespace;
+import org.springframework.cloud.kubernetes.fabric8.Fabric8Utils;
+import org.springframework.core.log.LogAccessor;
+
+/**
+ * Implementation that is based on Endpoints.
+ *
+ * @author wind57
+ */
+final class Fabric8EndpointsCatalogWatch
```
wind57, line 40: implementation that uses `Endpoints` - this is what we had until now. Nothing has changed logically here

spring-cloud-kubernetes-fabric8-discovery/src/main/java/org/springframework/cloud/kubernetes/fabric8/discovery/Fabric8EndpointSliceV1CatalogWatch.java
```diff
@@ -0,0 +1,73 @@
...
+
+	@Override
+	public List<EndpointNameAndNamespace> apply(Fabric8CatalogWatchContext context) {
+		// take only pods that have endpoints
+		List<EndpointSlice> endpointSlices;
+		if (context.properties().allNamespaces()) {
+			LOG.debug(() -> "discovering endpoints in all namespaces");
+
+			try (KubernetesClient client = context.kubernetesClient()) {
+				endpointSlices = client.discovery().v1().endpointSlices().inAnyNamespace()
```
wind57, line 51: I am using `client.discovery().v1()` here. There is also an option to use `v1beta1` for EndpointSlices, but since its `beta`, I left it out... If you disagree, please let me know.

spring-cloud-kubernetes-fabric8-discovery/src/test/java/org/springframework/cloud/kubernetes/fabric8/discovery/Fabric8KubernetesCatalogWatchEndpointSlicesTests.java
```diff
@@ -0,0 +1,325 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+/**
+ * Some tests that use the fabric8 mock client, using EndpointSlices
+ *
+ * @author wind57
+ */
+@EnableKubernetesMockClient
```
wind57, line 66: fabric tests for EndpointSlices

spring-cloud-kubernetes-integration-tests/spring-cloud-kubernetes-fabric8-client-catalog-watcher/src/test/resources/app/watcher-endpoints-deployment.yaml
```diff
@@ -0,0 +1,33 @@
...
+          livenessProbe:
+            httpGet:
+              port: 8080
+              path: /actuator/health/liveness
+          ports:
+            - containerPort: 8080
+          env:
+            - name: LOGGING_LEVEL_ORG_SPRINGFRAMEWORK_CLOUD_KUBERNETES_FABRIC8_DISCOVERY
+              value: DEBUG
+            - name: SPRING_CLOUD_KUBERNETES_DISCOVERY_USE_ENDPOINT_SLICES
```
wind57, line 32: property that enabled using EnpointSlices

spring-cloud-kubernetes-test-support/src/main/resources/setup/role.yaml
ryanjbaxter: As far as I understand the Endpoints API is going to continue to be supported and is not planned to be removed at the moment, correct?

Also it seems like Endpoint Slices is only officially supporting in k8s 1.2.1 right?  If you are using a version prior to that it was still in beta, correct?

I think if we move forward with this, I would make it an opt in feature.  Document how to opt in and the additional configuration that is needed to use it.

wind57: > As far as I understand the Endpoints API is going to continue to be supported and is not planned to be removed at the moment, correct?

Correct.

> Also it seems like Endpoint Slices is only officially supporting in k8s 1.2.1 right? If you are using a version prior to that it was still in beta, correct?

Yeah, they had alpha support in 1.16, then beta, then stable. 

> I think if we move forward with this, I would make it an opt in feature. Document how to opt in and the additional configuration that is needed to use it.

Excellent! I'll update the PR accordingly.

wind57:
@ryanjbaxter this is ready to be looked at too. thank you

spring-cloud-kubernetes-test-support/src/main/resources/setup/role.yaml
wind57: Thank you for your suggestions. I've updated the PR as suggested. Have a good trip next week!

spring-cloud-kubernetes-fabric8-discovery/src/main/java/org/springframework/cloud/kubernetes/fabric8/discovery/KubernetesCatalogWatch.java
ryanjbaxter: No problem!

PR #1159, opened 2022-12-02
fix

wind57:


wind57:
Two tests are stepping on each others toes when creating a cluster role binding with the same name. `Fabric8DiscoveryNamespaceFilterIT` and `DiscoveryClientFilterNamespaceIT`.

I would expect for this to happen on every PR, not only in `main` and not only on circleci. The explanation of why this happens is interesting too:

- on github actions, in logs, I see such entries for `Fabric8DiscoveryNamespaceFilterIT` and `DiscoveryClientFilterNamespaceIT`, which means they run in different containers. That is : these two tests run in different set-ups, they don't step on each other.

```
2:10:32.754 [main] INFO  🐳 [rancher/k3s:v1.25.4-k3s1] - Creating container for image: rancher/k3s:v1.25.4-k3s1
```

while on circleci:

```
04:31:09.385 [main] INFO  🐳 [rancher/k3s:v1.25.4-k3s1] - Reusing container with ID: 350ef33d687b16e2c8fb25b17f30611d9ef9d307ab6c9dd451606829ea56498d and hash: 4d34febe25e368b3f31dda74a9da096fc0273187
```

And _both_ `Fabric8DiscoveryNamespaceFilterIT` and `DiscoveryClientFilterNamespaceIT` run in the same container (https://app.circleci.com/pipelines/github/spring-cloud/spring-cloud-kubernetes/2644/workflows/4924500d-069b-46ed-9af7-04e97dc44004/jobs/5337), so they try to create a cluster role binding while running in the same container. Thus it breaks.

---

This PR does minor clean-up and renames cluster role binding.

ryanjbaxter (approved):
Its weird that this does not happen when building the PR on CircleCI.

wind57:
> Its weird that this does not happen when building the PR on CircleCI. 

There is a logical explanation to that too - i updated my comment, right after you commented.

PR #1163, opened 2022-12-06
Bump version to 16.0.2

ryanjbaxter:


wind57:
Is this change related to this one? https://github.com/kubernetes-client/java/pull/2487#issuecomment-1340084450

ryanjbaxter:
Somewhat ;)

I would like to be using the lates major release because once we are released we are stuck with that major.

Also there was a PR that was merged that provided hints for native support that we are likely to need at some point
https://github.com/kubernetes-client/java/pull/2402

Issue #1162, opened 2022-12-06
Add Labels And Annotations From Pods To ServiceInstance Metadata


weihubeats:
Is there a good way to get the pod's label if I want to. I might want to go through different pod tags to do rule routing

I see that the label in the metadata of the ServiceInstance I get seems to be the label of the kubernetes service

weihubeats:
<img width="1134" alt="image" src="https://user-images.githubusercontent.com/42484192/205862891-8ef44d8b-0d64-45df-a418-ca4ce2c75d98.png">
I want to get the tags of each pod, but I found that method
```java
// KubernetesInformerDiscoveryClient
public List<ServiceInstance> getInstances(String serviceId)
```
 does not have these tags, the tags of his original data are the tags of the service in kubernetes and not the tags of the pods

wind57:
it starts to make sense, but what is the point in you having the labels of the pods? this is what I don't get

weihubeats:
Because I might want to do a grayscale release, selecting the pods with grayscale labels from multiple pods in a service by labels them

wind57:
Seems like you want to get also the pods and their metadata from an underlying `ServiceInstance`. This is doable, I guess, but I need also @ryanjbaxter thoughts here.

ryanjbaxter:
I can possibly see the use case.  If we do this though, I wonder what the impact would be to the existing labels and annotations (could they conflict?).  Also we would have to correlates the endpoints to the pods, which i am not particularly clear how to do, not to say its impossible.

weihubeats:
We now want to use spring cloud kubernetes to implement grayscale or swimlane features, is there any good advice from the community?

weihubeats:
I can submit a pr if needed, I think it is available through V1PodList

ryanjbaxter:
A PR would be welcome

wind57:
`DefaultKubernetesServiceInstance` has a field `Map<String, Map<String, String>> podMetadata`.

All 3 clients return such metadata, if it is requested. This can be closed @ryanjbaxter

PR #1170, opened 2022-12-09
Add Labels And Annotations From Pods To ServiceInstance Metadata

weihubeats:
https://github.com/spring-cloud/spring-cloud-kubernetes/issues/1162

### DIFF
--- a/spring-cloud-kubernetes-client-discovery/src/main/java/org/springframework/cloud/kubernetes/client/discovery/KubernetesInformerDiscoveryClient.java
+++ b/spring-cloud-kubernetes-client-discovery/src/main/java/org/springframework/cloud/kubernetes/client/discovery/KubernetesInformerDiscoveryClient.java
@@ -17,13 +17,6 @@
 package org.springframework.cloud.kubernetes.client.discovery;
 
 
-import static org.springframework.cloud.kubernetes.commons.discovery.KubernetesDiscoveryConstants.HTTP;
-import static org.springframework.cloud.kubernetes.commons.discovery.KubernetesDiscoveryConstants.HTTPS;
-import static org.springframework.cloud.kubernetes.commons.discovery.KubernetesDiscoveryConstants.UNSET_PORT_NAME;
-
-import org.apache.commons.logging.Log;
-import org.apache.commons.logging.LogFactory;
-
 import java.time.Duration;
 import java.util.ArrayList;
 import java.util.HashMap;
@@ -36,16 +29,7 @@ import java.util.function.Supplier;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
-import org.springframework.beans.factory.InitializingBean;
-import org.springframework.cloud.client.ServiceInstance;
-import org.springframework.cloud.client.discovery.DiscoveryClient;
-import org.springframework.cloud.kubernetes.commons.discovery.DefaultKubernetesServiceInstance;
-import org.springframework.cloud.kubernetes.commons.discovery.KubernetesDiscoveryProperties;
-import org.springframework.util.Assert;
-import org.springframework.util.CollectionUtils;
-import org.springframework.util.ObjectUtils;
-import org.springframework.util.StringUtils;
-
+import io.kubernetes.client.extended.wait.Wait;
 import io.kubernetes.client.informer.SharedInformer;
 import io.kubernetes.client.informer.SharedInformerFactory;
 import io.kubernetes.client.informer.cache.Lister;
@@ -56,7 +40,23 @@ import io.kubernetes.client.openapi.models.V1ObjectMeta;
 import io.kubernetes.client.openapi.models.V1ObjectReference;
 import io.kubernetes.client.openapi.models.V1Pod;
 import io.kubernetes.client.openapi.models.V1Service;
-import io.kubernetes.client.util.wait.Wait;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+import org.springframework.beans.factory.InitializingBean;
+import org.springframework.cloud.client.ServiceInstance;
+import org.springframework.cloud.client.discovery.DiscoveryClient;
+import org.springframework.cloud.kubernetes.commons.discovery.DefaultKubernetesServiceInstance;
+import org.springframework.cloud.kubernetes.commons.discovery.KubernetesDiscoveryProperties;
+import org.springframework.util.Assert;
+import org.springframework.util.CollectionUtils;
+import org.springframework.util.ObjectUtils;
+import org.springframework.util.StringUtils;
+
+import static org.springframework.cloud.kubernetes.commons.discovery.KubernetesDiscoveryConstants.HTTP;
+import static org.springframework.cloud.kubernetes.commons.discovery.KubernetesDiscoveryConstants.HTTPS;
+import static org.springframework.cloud.kubernetes.commons.discovery.KubernetesDiscoveryConstants.UNSET_PORT_NAME;
+
 
 /**
  * @author Min Kim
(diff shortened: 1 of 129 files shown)


ryanjbaxter (changes requested):
I prefer if we take the code which collects the service metadata and the pod metadata and separate that out into its own method.  This is getting a bit complex with all the code in one method.  Also you want to respect whether the add labels and annotations preference is set.

ryanjbaxter:
Also please fix the checkstyle errors

weihubeats:
ok

spring-cloud-kubernetes-client-discovery/src/main/java/org/springframework/cloud/kubernetes/client/discovery/KubernetesInformerDiscoveryClient.java
```diff
@@ -145,39 +159,61 @@ private Stream<ServiceInstance> getServiceInstanceDetails(V1Service service, Str
...
+	private Map<String, String> initPodMetaData(V1EndpointAddress addr, Map<String, String> metadata, Map<String, V1ObjectMeta> podMap) {
+		Map<String, String> newMetaData = new HashMap<>(metadata);
+		if (!CollectionUtils.isEmpty(podMap)) {
+			V1ObjectReference targetRef = addr.getTargetRef();
+			if (!ObjectUtils.isEmpty(targetRef)) {
+				String name = targetRef.getName();
+				if (!ObjectUtils.isEmpty(name)) {
+					Optional.ofNullable(podMap.get(name)).ifPresent(podMetaData -> {
+						if (!CollectionUtils.isEmpty(podMetaData.getLabels())) {
+							newMetaData.putAll(podMetaData.getLabels());
```
ryanjbaxter, line 210: is this just adding labels, what about annotations?

spring-cloud-kubernetes-client-discovery/src/main/java/org/springframework/cloud/kubernetes/client/discovery/KubernetesInformerDiscoveryClient.java
```diff
@@ -145,39 +159,61 @@ private Stream<ServiceInstance> getServiceInstanceDetails(V1Service service, Str
...
+				if (this.properties.includeNotReadyAddresses()
+					&& !CollectionUtils.isEmpty(subset.getNotReadyAddresses())) {
+					addresses.addAll(subset.getNotReadyAddresses());
+				}
+
+				final int port = findEndpointPort(endpointPorts, primaryPortName, serviceId);
+				return addresses.stream()
+					.map(addr -> new DefaultKubernetesServiceInstance(
+						addr.getTargetRef() != null ? addr.getTargetRef().getUid() : "", serviceId,
+						addr.getIp(), port, properties.initPodMetaData() ? initPodMetaData(addr, metadata, finalPodMap) : metadata, secured, service.getMetadata()
```
ryanjbaxter, line 192: IDK that we need a separate property, we already have a property to enable adding annotations and labels

spring-cloud-kubernetes-client-discovery/src/main/java/org/springframework/cloud/kubernetes/client/discovery/KubernetesInformerDiscoveryClient.java
weihubeats: The metadata obtained here is the metadata of the pod, which should belong to metadata by definition

spring-cloud-kubernetes-client-discovery/src/main/java/org/springframework/cloud/kubernetes/client/discovery/KubernetesInformerDiscoveryClient.java
weihubeats: The metadata obtained here is the metadata of the pod, which should belong to metadata by definition

spring-cloud-kubernetes-client-discovery/src/main/java/org/springframework/cloud/kubernetes/client/discovery/KubernetesInformerDiscoveryClient.java
weihubeats: done

spring-cloud-kubernetes-client-discovery/src/main/java/org/springframework/cloud/kubernetes/client/discovery/reactive/KubernetesInformerReactiveDiscoveryClient.java
```diff
@@ -36,15 +37,15 @@
...
-	private KubernetesInformerDiscoveryClient kubernetesDiscoveryClient;
+	private final KubernetesInformerDiscoveryClient kubernetesDiscoveryClient;
 
 	public KubernetesInformerReactiveDiscoveryClient(KubernetesNamespaceProvider kubernetesNamespaceProvider,
-			SharedInformerFactory sharedInformerFactory, Lister<V1Service> serviceLister,
-			Lister<V1Endpoints> endpointsLister, SharedInformer<V1Service> serviceInformer,
-			SharedInformer<V1Endpoints> endpointsInformer, KubernetesDiscoveryProperties properties) {
+		SharedInformerFactory sharedInformerFactory, Lister<V1Service> serviceLister,
+		Lister<V1Endpoints> endpointsLister, SharedInformer<V1Service> serviceInformer, Lister<V1Pod> podLister,
+		SharedInformer<V1Endpoints> endpointsInformer, KubernetesDiscoveryProperties properties) {
```
ryanjbaxter, line 45: Same here we need a new constructor

ryanjbaxter (commented):
A couple of other things:

1.) We would need to add the same functionality to the Fabric8 implementation
2.) I am still unsure of what will happen when labels/annotations from the pod and service conflict.  Seems like they will override either other, last one added will win.  I am not sure how we could resolve this.

weihubeats:
> A couple of other things:
> 
> 1.) We would need to add the same functionality to the Fabric8 implementation 2.) I am still unsure of what will happen when labels/annotations from the pod and service conflict. Seems like they will override either other, last one added will win. I am not sure how we could resolve this.

I think there is no good way to solve the conflict, but only through the user to control themselves, because the metadata in the ServiceInstance is just a Map<String, String>

ryanjbaxter:
We still need documentation

wind57:
correlating the the address to the pod is not complicated, for example:

```
kubectl get -o json endpoints spring-cloud-kubernetes-client-configmap-event-reload
{
    "apiVersion": "v1",
    "kind": "Endpoints",
    "metadata": {
        "annotations": {
            "endpoints.kubernetes.io/last-change-trigger-time": "2023-02-21T14:01:05Z"
        },
        "creationTimestamp": "2023-02-21T14:00:55Z",
        "labels": {
            "app": "spring-cloud-kubernetes-client-configmap-event-reload"
        },
        "name": "spring-cloud-kubernetes-client-configmap-event-reload",
        "namespace": "default",
        "resourceVersion": "613",
        "uid": "2d017d3a-a18a-4ddb-ae2f-05c3f61ab117"
    },
    "subsets": [
        {
            "addresses": [
                {
                    "ip": "10.42.0.7",
                    "nodeName": "c577996ade89",
                    "targetRef": {
                        "kind": "Pod",
                        "name": "spring-cloud-kubernetes-client-configmap-deployment-event-jw5jz",
                        "namespace": "default",
                        "uid": "a499484e-0d8c-4668-be9c-f4df52c550d3"
                    }
                }
            ],
            "ports": [
                {
                    "name": "http",
                    "port": 8080,
                    "protocol": "TCP"
                }
            ]
        }
    ]
}
/ #
```

notice that each address (if it has a `targetRef`), has : 


```
                    "targetRef": {
                        "kind": "Pod",
                        "name": "spring-cloud-kubernetes-client-configmap-deployment-event-jw5jz",
                        "namespace": "default",
                        "uid": "a499484e-0d8c-4668-be9c-f4df52c550d3"
                    }
```

As to returning the metadata from pods, we could introduce a new record in `DefaultKubernetesServiceInstance`, for example `PodMetadata` that would contain things like pod-name and pod-labels. Of course you have to opt it, to get such data.

@ryanjbaxter if you agree, I would continue this at some point in time.

wind57:
I've added such functionality to the fabric8 client, pending review [here](https://github.com/spring-cloud/spring-cloud-kubernetes/pull/1254). I will, of course, add such an option to the k8s-client too later.

ryanjbaxter:
Addressed in https://github.com/spring-cloud/spring-cloud-kubernetes/pull/1254

PR #1177, opened 2022-12-14
Fix 1042

wind57:

PR #1181, opened 2022-12-16
Simplify integration tests

wind57:


wind57:
this is made on top of [this one](https://github.com/spring-cloud/spring-cloud-kubernetes/pull/1177), so it will wait.

While writing lately integration tests (+ the fast that last year I finished the year with that refactor to k3s in integration tests too and I was making a recap of the year :) ),I realized that reading the integration tests is just way too complicated. It's not supposed to be like this, especially since the code is not very much different from test to test, at least on the set-up side. So I am proposing some changes in this PR so start making some changes : easier to read and understand what is going on in an integration test. If you agree, I will slowly move all of them.

wind57:
After writing some integration tests I've realized that we go quite a lot of boilerplate code in each of them. In this PR, I've tried to simplify and hide all the complexity of an integration test set-up to a single class `Util`. I've also clean-ed up tests quite a lot, moved wiremock/kafka/rabbitmq/zookeeper related code to a single package. imho, now the integration test reads far simpler and it's easier to reason on what is going on.

spring-cloud-kubernetes-integration-tests/spring-cloud-kubernetes-fabric8-client-simple-core/src/test/java/org/springframework/cloud/kubernetes/fabric8/core/SimpleCoreIT.java
```diff
@@ -91,64 +85,25 @@ void test() {
 		Assertions.assertEquals("Hello from k8s profile", result);
 	}
 
-	private static void deleteManifests() {
+	private static void manifests(Phase phase) {
 
-		try {
```
wind57, line 96: @ryanjbaxter here is the idea of my PR, for example. In almost all of our integration tests, we create a deployment,service,ingress (and may be configmaps and secrets), and they do the exact same thing. First create them, do the testing, then deleted them. So why can't we simplify this and hide all the complexity of waiting for the resources to be created and deleted via proper methods? In such a case, reading the code of the integration test becomes far simpler to reason about. I've done this in all the tests + did some clean-up here and there.

wind57:
hey, welcome back. It's a simple reason, I hoped that the k8s watcher implementation would go in first, before this one

ryanjbaxter:
Ah OK, I missed that other PR.  I just merged it, so you should now be able to resolve the conflicts here

wind57:
@ryanjbaxter ready to be looked at. thank you

ryanjbaxter:
This is great!  Thank you!

PR #1183, opened 2022-12-21
Clean fabric8 discovery 1

wind57:


spring-cloud-kubernetes-fabric8-discovery/src/main/java/org/springframework/cloud/kubernetes/fabric8/discovery/KubernetesDiscoveryClientAutoConfiguration.java
```diff
@@ -55,23 +58,18 @@
...
-			}
-		}
-		else {
-			if (properties.allNamespaces()) {
-				return (client) -> client.services().inAnyNamespace().withLabels(properties.serviceLabels());
-			}
-			else {
-				return (client) -> client.services().withLabels(properties.serviceLabels());
-			}
+	public KubernetesClientServicesFunction servicesFunction(KubernetesDiscoveryProperties properties,
```
wind57, line 61: this simplifies the code a bit with less branches. You can pass `withLabels` all the time, without the need to check if they are present or not, simply because the default value for `properties.serviceLabels()` is an empty list

spring-cloud-kubernetes-fabric8-discovery/src/main/java/org/springframework/cloud/kubernetes/fabric8/discovery/Fabric8DiscoveryServicesAdapter.java
```diff
@@ -0,0 +1,87 @@
...
+ * @author wind57
+ */
+final class Fabric8DiscoveryServicesAdapter implements Function<KubernetesClient, List<Service>> {
+
+	private static final SpelExpressionParser PARSER = new SpelExpressionParser();
+
+	private static final SimpleEvaluationContext EVALUATION_CONTEXT = SimpleEvaluationContext.forReadOnlyDataBinding()
+			.withInstanceMethods().build();
+
+	private final KubernetesClientServicesFunction function;
```
wind57, line 47: the idea of having the `KubernetesClientServicesFunction` was probably good in the beginning when the code was only taking shape, now though, with the introduction of only some namespaces, is not the case anymore. 

Instead of removing it, I though of creating this "adapter" that allows to still use it, but at the same time allow the computation of services based on multiple namespaces.

spring-cloud-kubernetes-fabric8-discovery/src/main/java/org/springframework/cloud/kubernetes/fabric8/discovery/KubernetesDiscoveryClient.java
```diff
@@ -291,35 +287,13 @@ private Map<String, String> getMapWithPrefixedKeys(Map<String, String> map, Stri
...
-				return include;
-			};
-		}
-		return getServices(filteredServices);
+		return adapter.apply(client).stream().map(s -> s.getMetadata().getName()).toList();
 	}
 
+	@Deprecated(forRemoval = true)
 	public List<String> getServices(Predicate<Service> filter) {
-		if (properties.namespaces().isEmpty()) {
```
wind57, line 313: this method should have never been `public`, as it is called from within this class only. But it's here and we can only drop it in the next major release, as such I need to "dance" a little. `Fabric8DiscoveryServicesAdapter` has to take a `Predicate<Service>` in, though if this method was `private` that would have not been needed. Because of this, I also need to adapt some tests, by passing `null`

wind57: At the same time, I have a PR that drops the deprecated method and adjusted the code because of that. For example `Fabric8DiscoveryServicesAdapter` accepts a `Predicate<Service> filter` as a constructor parameter, though this is only needed because of the above public method. That PR takes care to properly clean things, not simply delete this public method, but to properly adjust the code everywhere because of that.

wind57:
@ryanjbaxter this one is ready to be looked at too... thank you.

PR #1188, opened 2022-12-28
Simplify config data functionality

wind57:


spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/KubernetesConfigDataLocationResolver.java
```diff
@@ -130,7 +130,7 @@ protected final boolean isRetryEnabledForSecrets(SecretsConfigProperties secrets
 				&& secretsProperties.failFast();
 	}
 
-	protected KubernetesNamespaceProvider kubernetesNamespaceProvider(Environment environment) {
```
wind57, line 133: this sits in commons package as `protected`, but implementations: `KubernetesClientConfigDataLocationResolver` and `Fabric8ConfigDataLocationResolver` don't override them in any meaningful way. Most probably just a left over of the initial though process when implementing.

spring-cloud-kubernetes-fabric8-config/src/main/java/org/springframework/cloud/kubernetes/fabric8/config/Fabric8ConfigDataLocationResolver.java
```diff
@@ -84,19 +83,12 @@ protected void registerBeans(ConfigDataLocationResolverContext resolverContext,
 
 	}
 
-	private KubernetesClient registerConfigAndClient(ConfigurableBootstrapContext bootstrapContext,
+	private KubernetesClient registerClient(ConfigurableBootstrapContext bootstrapContext,
 			KubernetesClientProperties kubernetesClientProperties) {
 		Config config = new Fabric8AutoConfiguration().kubernetesClientConfig(kubernetesClientProperties);
-		registerSingle(bootstrapContext, Config.class, config, "fabric8Config");
```
wind57, line 90: we do not need to register the `Config`, as it is only needed in order to create the `KubernetesClient`, that one though, needs to be registered.

spring-cloud-kubernetes-client-config/src/main/java/org/springframework/cloud/kubernetes/client/config/KubernetesClientConfigDataLocationResolver.java
```diff
@@ -83,20 +82,14 @@ protected void registerBeans(ConfigDataLocationResolverContext resolverContext,
 		}
 	}
 
-	private CoreV1Api registerClientAndCoreV1Api(ConfigurableBootstrapContext bootstrapContext,
+	private CoreV1Api registerCoreV1Api(ConfigurableBootstrapContext bootstrapContext,
 			KubernetesClientProperties kubernetesClientProperties) {
 		ApiClient apiClient = kubernetesApiClient();
 		apiClient.setUserAgent(kubernetesClientProperties.userAgent());
-		registerSingle(bootstrapContext, ApiClient.class, apiClient, "configDataApiClient");
```
wind57, line 90: we only need the `ApiClient` in order to create and register `CoreV1Api`. As such : do not register it and rename method to reflect that.

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/KubernetesConfigDataLocationResolver.java
```diff
@@ -140,7 +140,7 @@ private void registerProperties(ConfigDataLocationResolverContext resolverContex
 
 		ConfigurableBootstrapContext bootstrapContext = resolverContext.getBootstrapContext();
 		registerSingle(bootstrapContext, KubernetesClientProperties.class, clientProperties,
-				"configDataKubernetesClientProperties");
+				"kubernetesClientProperties");
```
wind57, line 143: this change is very interesting and I am still not very sure what is going on. If I do not rename it (so that it matches the argument of `Fabric8AutoConfiguration::kubernetesClientConfig`), when running some tests, I would get:

```
No qualifying bean of type 'org.springframework.cloud.kubernetes.commons.KubernetesClientProperties' available: expected single matching bean but found 2: spring.cloud.kubernetes.client-org.springframework.cloud.kubernetes.commons.KubernetesClientProperties,configDataKubernetesClientProperties
```

that is literally the message I get. On one hand, it makes total sense to report that `configDataKubernetesClientProperties` is already present - we register it. 

But what is `spring.cloud.kubernetes.client-org.springframework.cloud.kubernetes.commons.KubernetesClientProperties` is beyond me. 

If I create a constructor in `KubernetesClientProperties` and debug the code (so that I could create a breakpoint inside the constructor to see who is calling it and thus who is creating it), it seems that the test is trying to create another instance of this class, even if one is already registered. 

---

If I rename this field to match the name of the argument, then it will find the existing bean. I did not deal with it too much further (may be in future).

PR #1189, opened 2022-12-30
Simplify config data code a bit

wind57:


spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/ConfigUtils.java
```diff
@@ -271,9 +271,11 @@ private static Map<String, String> decodeData(Map<String, String> data) {
 
 	public static <T> void registerSingle(ConfigurableBootstrapContext bootstrapContext, Class<T> cls, T instance,
 			String name) {
-		bootstrapContext.registerIfAbsent(cls, BootstrapRegistry.InstanceSupplier.of(instance));
-		bootstrapContext.addCloseListener(event -> event.getApplicationContext().getBeanFactory()
-				.registerSingleton(name, event.getBootstrapContext().get(cls)));
+		if (instance != null && !bootstrapContext.isRegistered(cls)) {
```
wind57, line 274: I've done two things here:

- check if `instance` that is passed is null, so that it slightly simplifies callers. 
- check if class is already registered with `!bootstrapContext.isRegistered(cls)`. We already doing that via `regsiterIfAbsent`, but than we were calling `addCloseListener`, no matter if it was registered or not. Not a big deal, but this is cleaner, imo.

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/configdata/Properties.java
```diff
@@ -0,0 +1,72 @@
...
+import org.springframework.cloud.kubernetes.commons.KubernetesClientProperties;
+import org.springframework.cloud.kubernetes.commons.config.ConfigMapConfigProperties;
+import org.springframework.cloud.kubernetes.commons.config.SecretsConfigProperties;
+
+/**
+ * @author wind57
+ *
+ * holds properties needed for ConfigData.
+ */
+record Properties(KubernetesClientProperties clientProperties, ConfigMapConfigProperties configMapProperties,
```
wind57, line 31: I wanted a class that deals with properties only, so that all properties are computed in a single place. The code has not changed at all, it was just moved to this class from other places and moved to a separate package.

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/configdata/KubernetesConfigDataLocationResolver.java
```diff
@@ -0,0 +1,130 @@
...
+		return RETRY_IS_PRESENT && configMapProperties != null && configMapProperties.retry().enabled()
+				&& configMapProperties.failFast();
+	}
+
+	protected final boolean isRetryEnabledForSecrets(SecretsConfigProperties secretsProperties) {
+		return RETRY_IS_PRESENT && secretsProperties != null && secretsProperties.retry().enabled()
+				&& secretsProperties.failFast();
+	}
+
+	private KubernetesNamespaceProvider kubernetesNamespaceProvider(Environment environment) {
```
wind57, line 126: this was protected and overridden in both fabric8 and k8s-native clients, but the overridden version was exactly the same. Most probably an artifact of an initial thought process. I made it private here and dropped the methods in clients.

spring-cloud-kubernetes-fabric8-config/src/main/java/org/springframework/cloud/kubernetes/fabric8/config/Fabric8ConfigDataLocationResolver.java
```diff
@@ -49,11 +49,10 @@ public Fabric8ConfigDataLocationResolver(DeferredLogFactory factory) {
 
 	@Override
 	protected void registerBeans(ConfigDataLocationResolverContext resolverContext, ConfigDataLocation location,
-			Profiles profiles, KubernetesConfigDataLocationResolver.PropertyHolder propertyHolder,
```
wind57, line 52: it used to be `KubernetesConfigDataLocationResolver.PropertyHolder` and now it is `ConfigDataProperties`

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/configdata/KubernetesConfigDataLocationResolver.java
```diff
@@ -0,0 +1,130 @@
...
+			throws ConfigDataLocationNotFoundException, ConfigDataResourceNotFoundException {
+		return Collections.emptyList();
+	}
+
+	@Override
+	public final List<KubernetesConfigDataResource> resolveProfileSpecific(
+			ConfigDataLocationResolverContext resolverContext, ConfigDataLocation location, Profiles profiles)
+			throws ConfigDataLocationNotFoundException {
+
+		ConfigDataProperties properties = ConfigDataProperties.of(resolverContext).register();
```
wind57, line 87: the logic in this class stays exactly the same, but now it's a lot shorter and easier to read. `ConfigDataProperties.of(resolverContext).register()` this both computed all 3 properties needed (client, configmap and secret ones) + registers them.

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/configdata/ConfigDataProperties.java
```diff
@@ -0,0 +1,61 @@
...
+import org.springframework.cloud.kubernetes.commons.KubernetesClientProperties;
+import org.springframework.cloud.kubernetes.commons.config.ConfigMapConfigProperties;
+import org.springframework.cloud.kubernetes.commons.config.SecretsConfigProperties;
+
+import static org.springframework.cloud.kubernetes.commons.config.ConfigUtils.registerSingle;
+
+public record ConfigDataProperties(KubernetesClientProperties clientProperties,
+		ConfigMapConfigProperties configMapProperties, SecretsConfigProperties secretsProperties) {
+
+	static Registrar of(ConfigDataLocationResolverContext context) {
```
wind57, line 30: the idea here is once you call `of`, the only option you have after is `register` (this is partial functions kind of), which will register the properties and give them back in form of the `ConfigDataProperties` instance. My sole purpose of the PR was around this thing, mainly, to decouple and make configdata slightly easier to reason about.

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/configdata/KubernetesConfigDataResource.java
```diff
@@ -88,14 +86,6 @@ List<String> getAcceptedProfiles() {
 		return this.profiles.getAccepted();
 	}
 
-	public void setLog(Log log) {
```
wind57, line 91: we are not using these, so I decided to remove them

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/configdata/KubernetesConfigDataLocationResolver.java
```diff
@@ -0,0 +1,134 @@
...
+	protected final boolean isRetryEnabledForSecrets(SecretsConfigProperties secretsProperties) {
+		return RETRY_IS_PRESENT && secretsProperties != null && secretsProperties.retry().enabled()
+				&& secretsProperties.failFast();
+	}
+
+	private KubernetesNamespaceProvider kubernetesNamespaceProvider(Environment environment) {
+		return new KubernetesNamespaceProvider(environment);
+	}
+
+	private Environment environment(Map<String, Object> kubernetesConfigData, Profiles profiles) {
```
wind57, line 125: extract to a separate method, making `resolveProfileSpecific` even easier to read

PR #1190, opened 2023-01-04
Issue 1169 - left here only for future debugging, not for merge

wind57:


spring-cloud-kubernetes-integration-tests/spring-cloud-kubernetes-fabric8-client-configmap/src/test/resources/fabric8-deployment.yaml
```diff
@@ -26,3 +26,6 @@ spec:
               path: /actuator/health/liveness
           ports:
             - containerPort: 8080
```
wind57, line 28: Need to add some integration tests and assert presence of logs, but can be done in a separate PR too, depending on which one is accepted, if any of course

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/KubernetesConfigDataLoaderLoggerReconfigurer.java
```diff
@@ -0,0 +1,117 @@
...
+
+	private KubernetesConfigDataLoaderLoggerReconfigurer() {
+	}
+
+	static void reconfigureLoggers(DeferredLogFactory logFactory) {
+
+		log = logFactory.getLog(KubernetesConfigDataLoaderLoggerReconfigurer.class);
+
+		try {
+			Resource[] commonResources = RESOLVER.getResources("classpath*:" + COMMONS_PACKAGE + "/*.class");
```
wind57, line 61: the idea here is this:

- let's scan for all classes in the packages we care about, where potentially there are logger instances related to our config-data logic. 
- Once I have one class, create a `MethodHandles.Lookup privateLookup = MethodHandles.privateLookupIn(cls, MethodHandles.lookup());`. Basically in simple words give me a Lookup that can access private members.
- Once I have the above lookup, create a "hidden class" that has `NESTMATE` access to the class where the logger is. Think about it like this: If I found `ConfigUtils` for example, create another class (hidden), that has access to private members of `ConfigUtils`. Another way to think about it is an inner class inside `ConfigUtils`. 
- Once I create that hidden class, get its Nest Host and find a method that is static, and takes a Log.class type as a parameter. Once that method is found, get its name and invoke it with a logger instance that I already created.

---

If even more simplified : If I want to override the logger in `ConfigUtils`, then I need to create a method inside it : 

```
	// called from config-data loader
	private static void logger(Log logger) {
		ConfigUtils.logger = logger;
	}
```

Then I create an inner class inside `ConfigUtils` (not really, it's called a NESTMATE, but to build the mental model, I guess it works). With this inner class, call `getNestHost` (`ConfigUtils` that is) and invoke that `logger` method with the argument that I will provide.

ryanjbaxter:
@wind57 Can we close this one?

PR #1194, opened 2023-01-08
move fabric8 catalog to a separate package (CAN ONLY GO IN THE NEXT MAJOR RELEASE)

wind57:


wind57:
this aligns the implementation in fabric8 with the one in k8s-native. 
- It moves everything related to catalog watch to a separate package
- renames classes to denote that they are fabric8 specific
- tightens up some public modifiers to a more strict ones

PR #1195, opened 2023-01-08
minor visibility changes in k8s catalog watch implementation

wind57:


spring-cloud-kubernetes-client-discovery/src/main/java/org/springframework/cloud/kubernetes/client/discovery/catalog/KubernetesCatalogWatch.java
```diff
@@ -68,7 +68,7 @@ public void setApplicationEventPublisher(ApplicationEventPublisher publisher) {
 	}
 
 	@Scheduled(fixedDelayString = "${spring.cloud.kubernetes.discovery.catalogServicesWatchDelay:30000}")
```
wind57, line 70: this class is package-private, so dropping public here is safe

spring-cloud-kubernetes-client-discovery/src/main/java/org/springframework/cloud/kubernetes/client/discovery/catalog/KubernetesCatalogWatchAutoConfiguration.java
```diff
@@ -41,12 +41,12 @@
 @ConditionalOnDiscoveryEnabled
 @ConditionalOnCloudPlatform(CloudPlatform.KUBERNETES)
 @AutoConfigureAfter({ KubernetesClientAutoConfiguration.class })
-public class KubernetesCatalogWatchAutoConfiguration {
```
wind57, line 44: this is a configuration class and imho, messing with visibility accesses is safe for them.

wind57:
@ryanjbaxter imo, this can go in in the current release. It's not a big deal either, but I like making things better for us, even if they are minor. wdyt?

ryanjbaxter (approved):

PR #1203, opened 2023-01-19
Fix 1199 : fix bug

wind57:


spring-cloud-kubernetes-test-support/src/main/java/org/springframework/cloud/kubernetes/integration/tests/commons/Commons.java
```diff
@@ -105,12 +105,13 @@ public static void assertReloadLogStatements(String left, String right, String a
 					.getStdout();
 			await().pollInterval(Duration.ofSeconds(5)).atMost(Duration.ofSeconds(180)).until(() -> {
 
-				String allLogs = CONTAINER.execInContainer("kubectl", "logs", appPodName.trim()).getStdout();
-				LOG.info("==========================================================================================");
-				LOG.info(allLogs);
-				LOG.info("==========================================================================================");
-				if (allLogs.contains(left)) {
-					Assertions.assertFalse(allLogs.contains(right));
+				String present = CONTAINER
```
wind57, line 108: slightly re-write the way to get the logs

spring-cloud-kubernetes-integration-tests/spring-cloud-kubernetes-configuration-watcher-it/src/test/java/org/springframework/cloud/kubernetes/configuration/watcher/ActuatorRefreshMultipleNamespacesIT.java
```diff
@@ -0,0 +1,157 @@
...
+		util.wiremock(DEFAULT_NAMESPACE, "/", Phase.CREATE);
+	}
+
+	@AfterEach
+	void after() {
+		configWatcher(Phase.DELETE);
+		util.wiremock(DEFAULT_NAMESPACE, "/", Phase.DELETE);
+	}
+
+	/**
```
wind57, line 93: this is what I really wanted to have in this PR, an integration test with reload capability if multiple namespaces.

wind57: somehow this is the universes payback on me, that I did not write such an integration test in the first place, when I wrote this functionality. Lesson learned.

spring-cloud-kubernetes-integration-tests/spring-cloud-kubernetes-configuration-watcher-it/src/test/java/org/springframework/cloud/kubernetes/configuration/watcher/ActuatorRefreshMultipleNamespacesIT.java
```diff
@@ -0,0 +1,221 @@
...
+		util.createAndWait(LEFT_NAMESPACE, leftConfigMap, null);
+
+		// right-config-map
+		V1ConfigMap rightConfigMap = new V1ConfigMapBuilder().editOrNewMetadata()
+			.withLabels(Map.of("spring.cloud.kubernetes.config", "true"))
+			.withName("service-wiremock").withNamespace(RIGHT_NAMESPACE).endMetadata().addToData("color", "green").build();
+		util.createAndWait(RIGHT_NAMESPACE, rightConfigMap, null);
+
+		// comes from handler::onAdd (and as such from "onEvent")
+		Commons.assertReloadLogStatements("ConfigMap service-wiremock was added in namespace left",
```
wind57, line 129: if we assert that these two logs are present, it shows that we get events from both namespaces, meaning we "listen" for both namespaces and this was the initial bug.

spring-cloud-kubernetes-test-support/src/main/java/org/springframework/cloud/kubernetes/integration/tests/commons/Commons.java
wind57: I decided to use `grep` here because this was still sometimes failing for me. tbh, I still don't know why and I tried many things here. My hint is that in the previous version, I was retrieving all the logs and was doing a `contains`. Since there are many of them, and they are extracted once in 3 seconds, it sometimes breaks something, don't know what. Instead of digging into test-containers very deep, I decided to grep here. It's easier and a lot faster

spring-cloud-kubernetes-integration-tests/spring-cloud-kubernetes-configuration-watcher-it/src/test/java/org/springframework/cloud/kubernetes/configuration/watcher/ActuatorRefreshMultipleNamespacesIT.java
```diff
@@ -0,0 +1,221 @@
...
+		util.createAndWait(LEFT_NAMESPACE, leftConfigMap, null);
+
+		// right-config-map
+		V1ConfigMap rightConfigMap = new V1ConfigMapBuilder().editOrNewMetadata()
+				.withLabels(Map.of("spring.cloud.kubernetes.config", "true")).withName("service-wiremock")
+				.withNamespace(RIGHT_NAMESPACE).endMetadata().addToData("color", "green").build();
+		util.createAndWait(RIGHT_NAMESPACE, rightConfigMap, null);
+
+		// comes from handler::onAdd (and as such from "onEvent")
+		Commons.assertReloadLogStatements("ConfigMap service-wiremock was added in namespace left", "",
```
wind57, line 129: these two asserts are the main idea of the test: if they happen, it means we listen for changes in "selective namespaces" and its exactly what we want. This test will fail in `main` branch for example, because has this bug.

spring-cloud-kubernetes-fabric8-config/src/main/java/org/springframework/cloud/kubernetes/fabric8/config/reload/Fabric8EventBasedSecretsChangeDetector.java
```diff
@@ -124,19 +124,22 @@ private SecretInformerAwareEventHandler(SharedIndexInformer<Secret> informer) {
 
 		@Override
 		public void onAdd(Secret secret) {
```
wind57, line 126: decided to add the namespace in logs for easier tracking of the problem

spring-cloud-kubernetes-client-config/src/main/java/org/springframework/cloud/kubernetes/client/config/reload/KubernetesClientEventBasedConfigMapChangeDetector.java
```diff
@@ -111,34 +110,42 @@ void inform() {
...
+			// We need to pass an APIClient to the SharedInformerFactory because if we use
+			// the
+			// default
+			// constructor it will use the configured default APIClient but that may not
+			// contain
+			// an APIClient configured within the cluster and does not contain the
+			// necessary
+			// certificate authorities for the cluster. This results in SSL errors.
+			// See https://github.com/spring-cloud/spring-cloud-kubernetes/issues/885
+			SharedInformerFactory factory = new SharedInformerFactory(apiClient);
```
wind57, line 128: the problem is basically on this line. What I used to do is take a `SharedInformerFactory` and try to create many informers for different namespaces via multiple calls to `sharedIndexInformerFor`. Again: I was doing that on the same instance of `SharedInformerFactory`. Unfortunately, a `SharedInformerFactory` does not work like that (+ I am a total idiot for trusting my understanding and not writing a proper integration test).

A `SharedInformerFactory` has a cache internally on the Type of the informer and its instance. Since in my case I had the same type (`V1ConfigMap` for example), this was never working. I was never basically adding a new informer, all the actions happened on the same one. What was happening is that on the _same_ (I can not stress this enough) instance of an informer, I had two handlers. 

The fix after the above is quite trivial, create a new factory for each namespace and a new informer. Close informers and factories when done.

Now, what could happen if there are many namespaces? Potentially, these are calls to the API Server and I doubt that many of them will make the api-server happy, but I'll leave this one open in my mind and may be ask around on the kubernetes-client project. I hope this is OK for now.

wind57:
@ryanjbaxter ready to be looked at, fixes an existing issue. thank you.

spring-cloud-kubernetes-client-config/src/main/java/org/springframework/cloud/kubernetes/client/config/reload/KubernetesClientEventBasedConfigMapChangeDetector.java
ryanjbaxter: Basically all the informers for each namespace will be making a lot of API calls?

Where as if we could use a single informer they could make one API call for all configmap data for each namespace?

wind57: now that you made your comment, I realize that the same number of calls would be made; even if we could have used a single informer. 

A single informer, multiple namespaces (if that would be possible): 
- since RBAC rules need to be enforced, each of those namespaces would result in a different call to the API Server

Multiple informers, one for a single namespace:

- the same number of calls as above.

spring-cloud-kubernetes-test-support/src/main/java/org/springframework/cloud/kubernetes/integration/tests/commons/native_client/Util.java
```diff
@@ -465,6 +473,22 @@ private void waitForDeploymentToBeDeleted(String deploymentName, String namespac
 		});
 	}
 
+	private void waitForDeploymentPodsToBeDeleted(Map<String, String> labels, String namespace) {
```
wind57, line 476: Wait for the deployment pods to be deleted also. I had issues with some builds if I do not wait for this too, because two subsequents tests can step on each others toes when I search for them. If you need more verbose details, I can provide them, but it seems reasonable to wait for pods deletion also.

spring-cloud-kubernetes-test-support/src/main/java/org/springframework/cloud/kubernetes/integration/tests/commons/Commons.java
```diff
@@ -100,22 +101,36 @@ public static void loadSpringCloudKubernetesImage(String project, K3sContainer c
...
-					Assertions.assertFalse(allLogs.contains(right));
-					return true;
-				}
-				LOG.info("log statement not yet present");
-				return false;
-			});
+			String appPodName = CONTAINER.execInContainer("sh", "-c",
+					"kubectl get pods -l app=" + appLabel + " -o=name --no-headers | tr -d '\n'").getStdout();
+			LOG.info("appPodName : ->" + appPodName + "<-");
+			// we issue a pollDelay to let the logs sync in, otherwise the results are not
```
wind57, line 107: this `pollDelay(20, TimeUnit.SECONDS)` is supposed to fix our current build failure. The problem with the previous approach is that I did not wait for logs to sync in (this 20 seconds is supposed to do that), and I could query the logs at bad timing. Basically I need logs to be fully synced so that I could make proper asserts.

spring-cloud-kubernetes-client-config/src/main/java/org/springframework/cloud/kubernetes/client/config/reload/KubernetesClientEventBasedConfigMapChangeDetector.java
wind57: build is fine now, can be looked at again. thank you

ryanjbaxter: That makes sense

PR #1211, opened 2023-02-04
drop deprecations around informers + remove a few hacks we had where we were reading properties from Environment instead of properties

wind57:


spring-cloud-kubernetes-client-autoconfig/src/main/java/org/springframework/cloud/kubernetes/client/KubernetesClientAutoConfiguration.java
```diff
@@ -42,12 +42,9 @@ public class KubernetesClientAutoConfiguration {
 
 	@Bean
 	@ConditionalOnMissingBean
-	public ApiClient apiClient(Environment environment) {
```
wind57, line 45: the beauty of the previous clean-up that we did... We don't need this hack anymore, properties are correctly set. And we do already have tests that enforce this, and they don't fail once I made this change

spring-cloud-kubernetes-client-config/src/main/java/org/springframework/cloud/kubernetes/client/config/reload/KubernetesClientEventBasedConfigMapChangeDetector.java
```diff
@@ -115,16 +115,6 @@ void inform() {
 			if (enableReloadFiltering) {
 				filter[0] = ConfigReloadProperties.RELOAD_LABEL_FILTER + "=true";
 			}
-
-			// We need to pass an APIClient to the SharedInformerFactory because if we use
```
wind57, line 119: I decided to remove this comment because the default constructor for `SharedInformerFactory` is deprecated anyway

spring-cloud-kubernetes-client-discovery/src/main/resources/META-INF/spring/org.springframework.boot.autoconfigure.AutoConfiguration.imports
```diff
@@ -1,3 +1,4 @@
 org.springframework.cloud.kubernetes.client.discovery.catalog.KubernetesCatalogWatchAutoConfiguration
 org.springframework.cloud.kubernetes.client.discovery.KubernetesInformerDiscoveryClientAutoConfiguration
+org.springframework.cloud.kubernetes.client.discovery.KubernetesInformerAutoConfiguration
```
wind57, line 3: add an auto-configuration

spring-cloud-kubernetes-client-discovery/src/main/java/org/springframework/cloud/kubernetes/client/discovery/reactive/KubernetesInformerReactiveDiscoveryClientAutoConfiguration.java
```diff
@@ -88,24 +85,4 @@ public KubernetesInformerReactiveDiscoveryClient kubernetesReactiveDiscoveryClie
 				serviceLister, endpointsLister, serviceInformer, endpointsInformer, properties);
 	}
 
-	@Bean
-	@ConditionalOnMissingBean
-	public CatalogSharedInformerFactory catalogSharedInformerFactory() {
```
wind57, line 93: both `CatalogSharedInformerFactory` and `SpringCloudKubernetesInformerFactoryProcessor` are removed, as we do not use them anymore. They were using deprecated functionality anyway

spring-cloud-kubernetes-client-discovery/src/main/java/org/springframework/cloud/kubernetes/client/discovery/KubernetesInformerAutoConfiguration.java
```diff
@@ -0,0 +1,126 @@
...
+ * @author wind57
+ */
+@Configuration(proxyBeanMethods = false)
+@ConditionalOnDiscoveryEnabled
+@ConditionalOnKubernetesDiscoveryEnabled
+@ConditionalOnBlockingOrReactiveEnabled
+@ConditionalOnCloudPlatform(CloudPlatform.KUBERNETES)
+@AutoConfigureBefore({ SimpleDiscoveryClientAutoConfiguration.class, CommonsClientAutoConfiguration.class })
+@AutoConfigureAfter({ KubernetesClientAutoConfiguration.class, KubernetesDiscoveryPropertiesAutoConfiguration.class })
+public class KubernetesInformerAutoConfiguration {
```
wind57, line 60: the "meat" of this PR is here. My main trigger towards making these changes was to drop deprecations we had around informers. The suggestion around deprecations was to create all you need manually, and don't rely on `@KubernetesInformer` annotations, for example. And this is exactly what I do here: create all the needed beans by hand. It creates the same beans as we did before, basically.

spring-cloud-kubernetes-client-discovery/src/main/java/org/springframework/cloud/kubernetes/client/discovery/ConditionalOnBlockingOrReactiveEnabled.java
```diff
@@ -0,0 +1,63 @@
...
+ * {@link ConditionalOnReactiveDiscoveryEnabled} matches.
+ *
+ * @author wind57
+ */
+@Target({ ElementType.TYPE, ElementType.METHOD })
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+@Inherited
+@Conditional(ConditionalOnBlockingOrReactiveEnabled.OnBlockingOrReactiveEnabled.class)
+public @interface ConditionalOnBlockingOrReactiveEnabled {
```
wind57, line 43: I need this because informers related beans are only needed when either blocking or reactive clients are active

spring-cloud-kubernetes-client-discovery/src/main/java/org/springframework/cloud/kubernetes/client/discovery/KubernetesInformerAutoConfiguration.java
```diff
@@ -0,0 +1,126 @@
...
+		return new Lister<>(servicesSharedIndexInformer.getIndexer());
+	}
+
+	@Bean
+	@ConditionalOnMissingBean(value = V1Endpoints.class, parameterizedContainer = Lister.class)
+	public Lister<V1Endpoints> endpointsLister(SharedIndexInformer<V1Endpoints> endpointsSharedIndexInformer) {
+		return new Lister<>(endpointsSharedIndexInformer.getIndexer());
+	}
+
+	private String namespace(KubernetesDiscoveryProperties discoveryProperties,
```
wind57, line 109: this logic was verbatim taken as it was before in the previous code. It lacks support for "selective namespaces", but I will add it in a different PR

wind57:
@ryanjbaxter ready to be looked at. thank you.

ryanjbaxter:
There are a number of breaking changes in here so we are going to have to hold off on merging this

wind57:
I thought we are OK with breaking changes as long as these are config related, may be I miss something. Can you direct me to one breaking change that would invalidate this merge, so that I could not make this mistake again?

wind57:
My thought process was like this:

- deleted classes are OK to be removed, they were made public _only_ in the the previous merge that we had, thus they were not released and safe to be removed. 
- adjusting auto-configurations was done before by us in a "breaking manner", and no one should rely on those anyway.

I tend to agree with you a lot, cause you make sense, but I fail to see where the breaking change is in this PR :( please help me a bit. thank you!

spring-cloud-kubernetes-client-discovery/src/main/java/org/springframework/cloud/kubernetes/client/discovery/SpringCloudKubernetesInformerFactoryProcessor.java
```diff
@@ -1,120 +0,0 @@
...
-import org.springframework.beans.factory.support.AbstractBeanDefinition;
-import org.springframework.beans.factory.support.BeanDefinitionRegistry;
-import org.springframework.beans.factory.support.RootBeanDefinition;
-import org.springframework.cloud.kubernetes.commons.KubernetesNamespaceProvider;
-import org.springframework.core.ResolvableType;
-
-/**
- * @author Ryan Baxter
- */
-public class SpringCloudKubernetesInformerFactoryProcessor extends KubernetesInformerFactoryProcessor {
```
ryanjbaxter, line 46: This class existed in previous releases

ryanjbaxter (commented):
Sorry I didn't realize that you removing some stuff that you recently created, but I still see some breaking changes I believe

spring-cloud-kubernetes-client-discovery/src/main/java/org/springframework/cloud/kubernetes/client/discovery/SpringCloudKubernetesInformerFactoryProcessor.java
wind57: yes it did, but it was never public. It became public in the latest merge.

spring-cloud-kubernetes-client-autoconfig/src/main/java/org/springframework/cloud/kubernetes/client/KubernetesClientAutoConfiguration.java
wind57: the method is still there, granted the argument has changed. And this is an auto-configuration class, we had removals like these in configurations classes before. I mean it's a method providing a bean, I kind of thought these are always safe to modify, since they are always public... If you disagree this would be the only change making this PR not possible to go to `main`

wind57: of course, it would be easy to fix:

- `remove @Bean`
- add `@Deprecated(forRemoval=true)`
- add a new bean. 

:)

ryanjbaxter: Yes that is what I was going to suggest, just have the deprecated method call the new one.

We have been bitten by seemingness harmful changes, so I am trying to avoid that.

wind57: I can cause myself a massive headache how someone can potentially abuse this method in their code, or I can deprecate it and have a PR that simply replaces it. I already maintain 4 PRs that can go only in the next major release, so my plan is this:

- leave this method as is, just add `@Deprecated` to it
- update one of the "next major release" PRs with the change where this one is dropped and a new one created.

It's simpler. Thx a lot for your feedback.

PR #1212, opened 2023-02-07
fix

wind57:


spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/discovery/KubernetesDiscoveryClientHealthConfiguration.java
```diff
@@ -0,0 +1,41 @@
...
+import org.springframework.cloud.kubernetes.commons.PodUtils;
+import org.springframework.context.ApplicationEventPublisher;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+/**
+ * @author wind57
+ */
+@Configuration
+public class KubernetesDiscoveryClientHealthConfiguration {
```
wind57, line 31: since both fabric8 and k8s-native clients use the same exact bean, I decided to move them to a separate configuration and `@Import` it in the proper configs. Since we already had proper tests, by not failing, they proved that `@Import` respects AutoConfigureAfter and various @Contional(s)

PR #1219, opened 2023-02-09
drop getter and setter for client (CAN ONLY GO IN THE NEXT MAJOR RELEASE)

wind57:


spring-cloud-kubernetes-fabric8-discovery/src/main/java/org/springframework/cloud/kubernetes/fabric8/discovery/KubernetesDiscoveryClient.java
```diff
@@ -86,14 +86,6 @@ public KubernetesDiscoveryClient(KubernetesClient client,
 				kubernetesDiscoveryProperties, filter);
 	}
 
-	public KubernetesClient getClient() {
```
wind57, line 89: drop setters/getters

PR #1221, opened 2023-02-10
Clean subsetsns

wind57:


spring-cloud-kubernetes-fabric8-discovery/src/main/java/org/springframework/cloud/kubernetes/fabric8/discovery/EndpointSubsetNS.java
```diff
@@ -24,10 +24,4 @@
  * @author Haytham Mohamed
  **/
 record EndpointSubsetNS(String namespace, List<EndpointSubset> endpointSubset) {
-	EndpointSubsetNS {
```
wind57, line 27: I don't know why we were passing `null` and creating an empty List, we could do that at the point of were we called this constructor...

spring-cloud-kubernetes-fabric8-discovery/src/main/java/org/springframework/cloud/kubernetes/fabric8/discovery/KubernetesDiscoveryClient.java
```diff
@@ -257,16 +258,6 @@ private int findEndpointPort(EndpointSubset s, String serviceId, String primaryP
 		}
 	}
 
-	private EndpointSubsetNS getSubsetsFromEndpoints(Endpoints endpoints) {
```
wind57, line 260: move this method to utils. I'd like to "chop" DiscoverClient a lot more and individually test it and make it easier to read and work with

wind57:
@ryanjbaxter same direction...

PR #1230, opened 2023-02-18
get the namespace from the endpoints instead of the client

wind57:


spring-cloud-kubernetes-fabric8-discovery/src/main/java/org/springframework/cloud/kubernetes/fabric8/discovery/KubernetesDiscoveryClientUtils.java
```diff
@@ -49,11 +48,8 @@ private KubernetesDiscoveryClientUtils() {
 
 	}
 
-	static EndpointSubsetNS subsetsFromEndpoints(Endpoints endpoints, Supplier<String> clientNamespace) {
```
wind57, line 52: now that I have a bit cleaned-up and refactored the code, I can understand better what is going on and make some fixes, not just clean-ups. 

This method is called after we find endpoints in different namespace, there is no way for `endpoints` to be `null` here, so the check for `endpoints != null` is obsolete. 

Also, `endpoints.getSubsets()` can not be null either, as the "default" value from fabric8 for `getSubsets` is an empty List. As such, I dropped these checks altogether. 

Then the last point is that getting the namespace from the `client` is not correct, as we query for endpoints is various namespace, so the only correct way is to get it from the endpoints itself.

wind57:
@ryanjbaxter ready to be looked at, thank you.

PR #1233, opened 2023-02-20
Move namespace metadata

wind57:


spring-cloud-kubernetes-fabric8-discovery/src/main/java/org/springframework/cloud/kubernetes/fabric8/discovery/KubernetesDiscoveryClient.java
```diff
@@ -141,14 +140,9 @@ private List<ServiceInstance> getNamespaceServiceInstances(EndpointSubsetNS es,
 		List<ServiceInstance> instances = new ArrayList<>();
 
 		Service service = client.services().inNamespace(namespace).withName(serviceId).get();
-		Map<String, String> serviceMetadata = serviceMetadata(serviceId, service, properties, subsets);
+		Map<String, String> serviceMetadata = serviceMetadata(serviceId, service, properties, subsets, namespace);
 
 		for (EndpointSubset endpointSubset : subsets) {
-
-			if (properties.allNamespaces()) {
```
wind57, line 148: this has moved to the method that computes metadata only. But also fixes kind of an inconsistency. We were returning the namespace metadata (via `k8s_namespace` key) only for when all-namespaces was true. But why? the namespace is accessible for all other options as-well, so I decided to return it regardless of this setting.

spring-cloud-kubernetes-fabric8-discovery/src/main/java/org/springframework/cloud/kubernetes/fabric8/discovery/KubernetesDiscoveryClientUtils.java
```diff
@@ -152,6 +153,7 @@ static Map<String, String> serviceMetadata(String serviceId, Service service,
 			serviceMetadata.putAll(portMetadata);
 		}
 
+		serviceMetadata.put(NAMESPACE_METADATA_KEY, namespace);
```
wind57, line 156: namespace metadata is always present and returned.

wind57:
@ryanjbaxter I guess one more minor one. thank you

PR #1234, opened 2023-02-20
Use SecretsPropertySource for Kubernetes secrets

breun:
This PR replaces the use of Spring Framework's `MapPropertySource` with Spring Cloud's `SecretsPropertySource` (which extends `MapPropertySource`). This is semantically clearer and allows users to sanitize secrets based on their property source class.

### DIFF
--- a/spring-cloud-kubernetes-client-config/src/main/java/org/springframework/cloud/kubernetes/client/config/KubernetesClientSecretsPropertySourceLocator.java
+++ b/spring-cloud-kubernetes-client-config/src/main/java/org/springframework/cloud/kubernetes/client/config/KubernetesClientSecretsPropertySourceLocator.java
@@ -21,9 +21,9 @@ import io.kubernetes.client.openapi.apis.CoreV1Api;
 import org.springframework.cloud.kubernetes.commons.KubernetesNamespaceProvider;
 import org.springframework.cloud.kubernetes.commons.config.NormalizedSource;
 import org.springframework.cloud.kubernetes.commons.config.SecretsConfigProperties;
+import org.springframework.cloud.kubernetes.commons.config.SecretsPropertySource;
 import org.springframework.cloud.kubernetes.commons.config.SecretsPropertySourceLocator;
 import org.springframework.core.env.ConfigurableEnvironment;
-import org.springframework.core.env.MapPropertySource;
 
 import static org.springframework.cloud.kubernetes.client.KubernetesClientUtils.getApplicationNamespace;
 
@@ -45,7 +45,7 @@ public class KubernetesClientSecretsPropertySourceLocator extends SecretsPropert
 	}
 
 	@Override
-	protected MapPropertySource getPropertySource(ConfigurableEnvironment environment, NormalizedSource source) {
+	protected SecretsPropertySource getPropertySource(ConfigurableEnvironment environment, NormalizedSource source) {
 
 		String normalizedNamespace = source.namespace().orElse(null);
 		String namespace = getApplicationNamespace(normalizedNamespace, source.target(), kubernetesNamespaceProvider);
--- a/spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/ConfigDataRetryableSecretsPropertySourceLocator.java
+++ b/spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/ConfigDataRetryableSecretsPropertySourceLocator.java
@@ -20,7 +20,6 @@ import java.util.Collection;
 
 import org.springframework.core.env.ConfigurableEnvironment;
 import org.springframework.core.env.Environment;
-import org.springframework.core.env.MapPropertySource;
 import org.springframework.core.env.PropertySource;
 import org.springframework.retry.support.RetryTemplate;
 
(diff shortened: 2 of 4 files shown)


wind57:
oh nice! thank you for this.

wind57:
@ryanjbaxter can we merge this also? It looks good to me.

ryanjbaxter:
For me my general bias is to keep things as general as possible, but I am open to being convinced otherwise

> This is semantically clearer and allows users to sanitize secrets based on their property source class.

I am a little unclear on what this means, can you elaborate?

wind57:
would an argument be that we already use this `SecretsPropertySource` in other places as-well? 

- https://github.com/spring-cloud/spring-cloud-kubernetes/blob/main/spring-cloud-kubernetes-client-config/src/main/java/org/springframework/cloud/kubernetes/client/config/KubernetesClientSecretsPropertySource.java#L30
- https://github.com/spring-cloud/spring-cloud-kubernetes/blob/main/spring-cloud-kubernetes-fabric8-config/src/main/java/org/springframework/cloud/kubernetes/fabric8/config/Fabric8SecretsPropertySource.java#L33

ryanjbaxter:
Extending `SecretsPropertySource` is different to me then using it as a return type.  I can see the argument that you need to change the return type to then use a method that is specific to `SecretsPropertySource` but that doesn't appear to be the case here.

wind57:
good point!

breun:
[Spring Boot 3.0 masks all property values by default](https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-3.0-Migration-Guide#actuator-endpoints-sanitization) and provides options to mask nothing or only mask when not authenticated, but I'd like to only mask values that come from sensitive property sources, like Kubernetes secretes and Google Cloud Secret Manager.

Spring Boot 3.0 added the concept of a `SanitizingFunction` to let users customize the sanitizing behavior, but creating one for file-based Kubernetes secrets is currently pretty hard, because `SecretsPropertySourceLocator` currently creates a `CompositePropertySource` with one `MapPropertySource` for each secret file. Since these are generic classes, it is not really possible to sanitize these property sources based on their 'origin'.

If this PR would get merged, , then I could sanitize them like this:

```
@Bean
SanitizingFunction secretsPropertySourceSanitizingFunction() {
    return data -> {
        PropertySource<?> propertySource = data.getPropertySource();
        if (propertySource instanceof SecretsPropertySource) {
            return maskValue(data);
        }
        return data;
    };
}
```

Spring Cloud Kubernetes indeed already creates `SecretsPropertySource` instances in other places (like @wind57 showed above), and I think it would be nicely consistent for `SecretsPropertySourceLocator` to also create `SecretsPropertySource` instances. It's even in the name.

ryanjbaxter:
My thought then is to have Spring Cloud Kubernetes mask the values so you (and others) don't have to do it yourself.  I agree that secret values should be masked.

wind57:
That is for sure something I would like to look into, woukd you mind creating a defect? If Ryan agrees, of course.

breun:
@ryanjbaxter Could the issue for letting Spring Cloud Kubernetes mask secret values by default be a follow-up ticket to this one, which basically auto-configures a `SanitizingFunction` like the one I showed above?

ryanjbaxter:
Yes but I am still not convinced by the need to make these changes.

If the only need for them is for you to write a sanitize function, I would rather just make changes to sanitize things in Spring Cloud Kubernetes.

breun:
But wouldn’t Spring Cloud Kubernetes do that using the same `SanitizingFunction`? Do you see another approach?

ryanjbaxter:
> creating one for file-based Kubernetes secrets is currently pretty hard

can you elaborate on what a "file-based Kubernetes secret" is?

ryanjbaxter:
I am testing various cases and when I look at `/actuator/env` all values in the PropertySources from Spring Cloud Kubernetes all appear to be sanitized

breun:
> can you elaborate on what a "file-based Kubernetes secret" is?

A secret that is loaded from a file mounted on the filesystem, as loaded via `org.springframework.cloud.kubernetes.commons.config.SecretsPropertySourceLocator#putPathConfig`.

breun:
> I am testing various cases and when I look at `/actuator/env` all values in the PropertySources from Spring Cloud Kubernetes all appear to be sanitized

All values from all property sources are sanitised by default in Spring Boot 3.0, even non-secrets. What I'm looking for is to only sanitize secrets, but keep other values visible when `management.endpoint.env.show-values` and `management.endpoint.configprops.show-values` are set to `ALWAYS` (which applies the sanitizing functions).

ryanjbaxter:
OK now this makes more sense.  You are trying to still sanitize some values based on the where they are coming from even though the `show-values` property is set to `ALWAYS`.

One quick question, have you looked into looking at the property source name to determine if its a kubernetes secret?

breun:
The `MapPropertySource` instances get the name of the file from which the secret is loaded, so that is hard to use. They do get collected into a `CompositePropertySource` called `composite-secrets`, so currently I’m indeed trying to inspect that composite property source to find the names of the nested `MapPropertySource` instances, but that is quite convoluted and maybe even brittle logic. I think that being able to do an `instanceof SecretsPropertySource` check would be cleaner, clearer and more reliable.

ryanjbaxter:
Sorry it took me so long to understand what you were trying to do!

breun:
No problem. I've created #1242 as a follow-up.

PR #1236, opened 2023-02-22
Check for empty data in secret

ryanjbaxter:
If the secret contains no data the `transform` method will throw a NPE.  This exception gets caught but the name generation of the secret is incorrect.


spring-cloud-kubernetes-client-config/src/main/java/org/springframework/cloud/kubernetes/client/config/KubernetesClientSecretsCache.java
```diff
@@ -81,7 +82,8 @@ private static List<StrippedSourceContainer> strippedSecrets(List<V1Secret> secr
 	}
 
 	private static Map<String, String> transform(Map<String, byte[]> in) {
-		return in.entrySet().stream().collect(Collectors.toMap(Map.Entry::getKey, en -> new String(en.getValue())));
+		return ObjectUtils.isEmpty(in) ? Map.of() :
```
wind57, line 85: did not expect this one :| I guess reproduced somewhere for you internally on a project?

ryanjbaxter: yup ran into it when i was trying to looking into the masking secrets PR

PR #1238, opened 2023-02-23
Add namespace to service instance result

wind57:


spring-cloud-kubernetes-fabric8-discovery/src/main/java/org/springframework/cloud/kubernetes/fabric8/discovery/KubernetesDiscoveryClient.java
```diff
@@ -146,17 +146,9 @@ private List<ServiceInstance> getNamespaceServiceInstances(EndpointSubsetNS es,
 			int endpointPort = endpointsPort(endpointSubset, serviceId, properties, service);
 			List<EndpointAddress> addresses = addresses(endpointSubset, properties);
 			for (EndpointAddress endpointAddress : addresses) {
-
```
wind57, line 149: I removed this code and moved it to the utils + added a test to assert proper fields. 

I've also added the namespace as part of the result.

wind57:
@ryanjbaxter this fixes a tiny little issue (not yet reported) where we do not return the namespace in the service instances

wind57:
btw, I will start porting all these features and changes we do in the fabric8 client to the k8s-native also, but only once all the issues related to it are resolved, so that I have a good baseline what needs to be done. I hope this is OK.

PR #1240, opened 2023-02-23
Introduce namespace provider

wind57:


spring-cloud-kubernetes-fabric8-discovery/src/main/java/org/springframework/cloud/kubernetes/fabric8/discovery/KubernetesDiscoveryClient.java
```diff
@@ -171,4 +178,9 @@ public int getOrder() {
 		return properties.order();
 	}
 
+	@Deprecated(forRemoval = true)
```
wind57, line 181: this PR introduces kubernetes namespace provider support. For it, I need to use `Fabric8Utils::getApplicationNamespace`, which requires a `KubernetesNamespaceProvider`, which requires `Environment`. 

Since adding another variable to the constructor would introduce a breaking change, I am adding this one. I am currently updating an existing PR (that will only go in the next major release) where I drop this method and inject an Environment variable in the constructor.

wind57:
@ryanjbaxter this adds support for kubernetes namespace provider. We already have it in CatalogWatcher, but not here.

In my next PR, I will add more "goodies", where we delegate _all_ calls related to endpoints finding to a single method, cause right now they are scattered across at least 3 places in the fabric8 discovery module.

PR #1241, opened 2023-02-23
Unify endpoints calls and resolution

wind57:


spring-cloud-kubernetes-fabric8-discovery/src/main/java/org/springframework/cloud/kubernetes/fabric8/discovery/Fabric8EndpointsCatalogWatch.java
```diff
@@ -41,30 +38,10 @@
 final class Fabric8EndpointsCatalogWatch
 		implements Function<Fabric8CatalogWatchContext, List<EndpointNameAndNamespace>> {
 
-	private static final LogAccessor LOG = new LogAccessor(LogFactory.getLog(Fabric8EndpointsCatalogWatch.class));
-
 	@Override
 	public List<EndpointNameAndNamespace> apply(Fabric8CatalogWatchContext context) {
-		List<Endpoints> endpoints;
```
wind57, line 48: since we need to find endpoints in two places: CatalogWatch (the implementation that uses `Endpoints`) and Discovery client - why not unify these two to delegate to a single method? This is what this PR does.

wind57:
@ryanjbaxter just some more tightening of the code. thank you.

PR #1243, opened 2023-02-24
Add externalname support

wind57:


spring-cloud-kubernetes-fabric8-discovery/src/main/java/org/springframework/cloud/kubernetes/fabric8/discovery/Fabric8DiscoveryServicesAdapter.java
```diff
@@ -1,92 +0,0 @@
-/*
- * Copyright 2012-2022 the original author or authors.
```
wind57, line 2: this was never public, and I introduced it quite recently. I added it because I was not aware of all the code-paths; but now that I am I realized that it is not needed at all

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/discovery/KubernetesDiscoveryProperties.java
```diff
@@ -56,14 +58,16 @@ public record KubernetesDiscoveryProperties(
 		@DefaultValue Map<String, String> serviceLabels, String primaryPortName,
 		@DefaultValue Metadata metadata,
 		@DefaultValue("" + DEFAULT_ORDER) int order,
-		boolean useEndpointSlices) {
+		boolean useEndpointSlices,
+		boolean includeExternalNameServices) {
```
wind57, line 62: I added `includeExternalNameServices` only in properties as the actual support for it will be added in the upcoming PR. The fact is, I started working on support for `ExternalName` type of services and realized that unless I clean-up things, searching and filtering services is going to be scattered across multiple places (an adapter I recently introduced and an actual place where external-services support will be added). 

So if its OK, I will leave this flag here (will all the test changes also) and once this is merged, will continue work on ExternalName type of services.

spring-cloud-kubernetes-fabric8-discovery/src/main/java/org/springframework/cloud/kubernetes/fabric8/discovery/KubernetesClientServicesFunction.java
```diff
@@ -40,8 +40,11 @@
  *
  * See KubernetesDiscoveryClientAutoConfiguration.servicesFunction
  *
+ * This is not used internally anymore and will be removed in a future release.
+ *
  * @author Georgios Andrianakis
  */
+@Deprecated(forRemoval = true)
```
wind57, line 47: we don't need this now.

So initially, services where look-ed up only via this indirection, that it was OK having it.
Now, when I add support for ExternalName type of services, there will be _two_ places where services are being searched and this interface will only cause problems and inconsistencies. So I dropped it's usage and deprecated it.

spring-cloud-kubernetes-fabric8-discovery/src/main/java/org/springframework/cloud/kubernetes/fabric8/discovery/KubernetesDiscoveryClient.java
```diff
@@ -146,13 +149,14 @@ private List<ServiceInstance> getNamespaceServiceInstances(EndpointSubsetNS es,
 
 	@Override
 	public List<String> getServices() {
-		return adapter.apply(client).stream().map(s -> s.getMetadata().getName()).toList();
+		return services(properties, client, namespaceProvider, filter(), null, "fabric8-discovery").stream()
```
wind57, line 152: don't use "adapter", delegate to `services`. This will also be used by the new functionality of external-name services

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/discovery/KubernetesDiscoveryProperties.java
wind57, line 62: add support to opt-in into this new functionality.

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/discovery/KubernetesExternalNameServiceInstance.java
```diff
@@ -0,0 +1,71 @@
...
+
+import org.springframework.cloud.client.ServiceInstance;
+
+/**
+ * Type of {@link org.springframework.cloud.client.ServiceInstance} when
+ * "spec.type=ServiceInstance".
+ *
+ * @author wind57
+ */
+public record KubernetesExternalNameServiceInstance(String serviceId, String host, String instanceId,
```
wind57, line 30: added a new type of `ServiceInstance`

wind57:
@ryanjbaxter ready to be looked at. Thank you.

wind57:
forgot to mention: all of the functionality in fabric8 discovery will be ported slowly to k8s-native, of course. When that is done, I will also update the documentation. I hope this is OK.

ryanjbaxter:
I am on vacation till next week I will look at it when I return

ryanjbaxter (commented):
This seems obvious, but why wouldn't the fabric8 client return externalname services when you request the services?  It seems like it should include all types...
Or is it because we are actually building the service list from the endpoints which I guess would not include the external name services.

wind57:
you're last point is exactly correct. welcome back!

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/discovery/KubernetesDiscoveryProperties.java
ryanjbaxter: Can we add a new constructor instead of modifying the existing one?

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/discovery/KubernetesExternalNameServiceInstance.java
ryanjbaxter: Why a new one?

wind57: excellent question! as usual. How are users supposed to distinguish between an ExternalName ServiceInstance and one that is based on endpoints? I had the idea to have a `type` here that will present that information...

ryanjbaxter: What about via the type in the metadata?

wind57: I feel like an idiot, you're right. Will change, thx a lot!

ryanjbaxter: Not an idiot at all, this is why we review code!

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/discovery/KubernetesDiscoveryProperties.java
wind57: I don't think we can for records, but we have defined for example `useEndpointSlices` in here in a previous PR

ryanjbaxter: I believe you can create multiple constructors for records.  I would prefer we just create a new one, its less disruptive, and this would be a breaking change.

wind57: right, you can create multiple constructors, but at least one of them (that adds this new parameter `includeExternalNameServices`) will have to set an instance field and those are not allowed in records. I'll think about this more, but so far, have not seen an option to achieve this.

ryanjbaxter: Why wouldn't something like this work?

```
public record KubernetesDiscoveryProperties(
		@DefaultValue("true") boolean enabled, boolean allNamespaces,
		@DefaultValue Set<String> namespaces,
		@DefaultValue("true") boolean waitCacheReady,
		@DefaultValue("60") long cacheLoadingTimeoutSeconds,
		boolean includeNotReadyAddresses, String filter,
		@DefaultValue({"443", "8443"}) Set<Integer> knownSecurePorts,
		@DefaultValue Map<String, String> serviceLabels, String primaryPortName,
		@DefaultValue Metadata metadata,
		@DefaultValue("" + DEFAULT_ORDER) int order,
		boolean useEndpointSlices,
		boolean includeExternalNameServices) {

	public KubernetesDiscoveryProperties(@DefaultValue("true") boolean enabled, boolean allNamespaces,
		@DefaultValue Set<String> namespaces,
		@DefaultValue("true") boolean waitCacheReady,
		@DefaultValue("60") long cacheLoadingTimeoutSeconds,
		boolean includeNotReadyAddresses, String filter,
		@DefaultValue({"443", "8443"}) Set<Integer> knownSecurePorts,
		@DefaultValue Map<String, String> serviceLabels, String primaryPortName,
		@DefaultValue Metadata metadata,
		@DefaultValue("" + DEFAULT_ORDER) int order,
		boolean useEndpointSlices) {
		this(enabled, allNamespaces, namespaces, waitCacheReady, cacheLoadingTimeoutSeconds, includeNotReadyAddresses, filter, knownSecurePorts, serviceLabels, 
			primaryPortName, metadata, order, useEndpointSlices, false);
	}
....
```

wind57: This was simply, sorry, I got confused a little here. I should have thought about this in reverse: add a new constructor to match the public one we already had. Thank you for the good tip. I also had to add:

```
	@ConstructorBinding
	public KubernetesDiscoveryProperties {

	}
```

otherwise the tests are failing because it seems that spring-boot is looking for a single constructor for final classes, so we have to tell it which one to use here.

ryanjbaxter: Great!  Shouldn't we no longer need all the constructor changes in the tests now?

wind57: yeah, sorry, I don't get your point. :( Do you mean we should not longer enforce all those `@DefaultValue` via `KubernetesDiscoveryPropertiesTests`?

ryanjbaxter: No I mean where you added `false` in all the tests

wind57: makes sense, done. seems like you can review again, your comments have been taken care of, imo.

ryanjbaxter (changes requested):
It would make sense to add documentation on the new property

wind57:
agree, I was thinking into first making the same changes into k8s-native client and only then add the documentation, but if you say add now, sure, will add it. thx.

ryanjbaxter:
I would add it now and specify the functionality is only available with fabric8 and then when you get the PR ready for the k8s client you can remove that line.  That way if we happen to release before you get the second PR ready its documented properly

PR #1244, opened 2023-02-25
first

wind57:


wind57:
Now that I have everything in my cache about fabric8 discovery implementation, I can change the code to a simpler version. The changes here are pretty destructive, but they simplify the implementation quite a bit and also hides a lot of things so that it is easier to maintain in the long run. This has the same spirit of the changes that I did at some point in time related to config packages. I will comment and keep this one up to date with changes that will come in main.

PR #1247, opened 2023-03-07
Add caching to load balancer ServiceListSupplier.  

### LINKED ISSUE #1225: LoadBalancer for Kubernetes resulting in high number of requests to K8s API Server prior to every downstream call to lookup
**Describe the bug**
We have both Eureka discovery client (default) and spring-cloud-k8s  discovery client (**Fabric8** impl) in our spring apps that gets enabled based on environment. 

To make the switch from Eureka to K8s discovery we have following properties injected to all:
```
        - name: eureka.client.enabled
          value: "false"
        - name: spring.cloud.kubernetes.enabled
          value: "true"
        - name: spring.cloud.kubernetes.loadbalancer.mode
          value: "SERVICE"
```
When communicating from one service to another via LoadBalanced RestTemplate K8s API server is hit every time. This leads high load on K8s API resulting in rate limiting/blocking for the overall cluster.

 `spring.cloud.kubernetes.loadbalancer.mode` default to `POD` in which case its fine (K8s API server is not hit on each request ) but we wanted to "SERVICE" for our specific use case. 

A workaround is to disable loadbalancer in general when relying on K8s-Service using `spring.cloud.kubernetes.loadbalancer.enabled=false`, but this raises the question of usage of `loadbalancer.mode` other values.

Any insights on expected behaviour?

Versions - Happens both in spring-boot-3.0.2/spring-cloud-2022.0.1 and spring-boot-2.7.7/spring-cloud-2021.0.5

**Sample**
Sample code and process to reproduce locally has been pushed [here](https://github.com/dhananjay12/spring-microservice-demo/tree/master/k8s/spring-k8s-loadbalancer-test). The services involved in the sample K8s file source code is in the same repository.

ryanjbaxter:
Fixes #1225


OlgaMaciaszek (approved):
LGTM.

PR #1253, opened 2023-03-12
unique service names (main)

wind57:


spring-cloud-kubernetes-fabric8-discovery/src/main/java/org/springframework/cloud/kubernetes/fabric8/discovery/KubernetesDiscoveryClient.java
```diff
@@ -162,13 +162,13 @@ private List<ServiceInstance> getNamespaceServiceInstances(EndpointSubsetNS es,
 
 	@Override
 	public List<String> getServices() {
```
wind57, line 164: if we have for example enabled `all-namespaces=true` and there is such instances:

- `my-service` in namespace `a`
- `my-service` in namespace `b`

a call to `getServices` will currently return `[my-service, my-service]`, which imo, does not make sense to the callers. wdyt?

wind57:
@ryanjbaxter I guess a minor thing to look at. thank you.

ryanjbaxter:
Wait I thought I said they should be unique?

wind57:
> Yes we should be returning duplicates of the same name, because they would be considered the same service

did I miss-understood your comment?

ryanjbaxter:
Oops yes I have a typo there..."should NOT be returning duplicates"

ryanjbaxter:
We should fix this in 2.1.x

wind57:
no worries, I'll create a PR for that also.

ryanjbaxter:
We should just need one and merge it forward

wind57:
but the branches have diverged a lot in this area. wouldn't it be easier to have two PRs?

ryanjbaxter:
2 PRs is fine, i still have to merge 2.1.x into main though, but its no big deal to me

wind57:
in such a case, can we merge this one? :)

ryanjbaxter:
I want to merge the 2.1.x version first because like I said I need to merge that forward (unnecessarily).  Are there tests that need to be changed in that branch?

wind57:
no, iirc, there are not test changes.

But then, I'd like to make sure that the tests indeed pass on `2.1.x`. So how about I create a PR to have a pipeline in `2.1.x`, first? I would hate to push something without the entire pipeline to run. I'll make this my priority.

PR #1254, opened 2023-03-12
add pod metadata and annotations as part of the service discovery response

wind57:


spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/discovery/KubernetesDiscoveryProperties.java
```diff
@@ -92,15 +92,30 @@ public KubernetesDiscoveryProperties(@DefaultValue("true") boolean enabled, bool
...
-			@DefaultValue("true") boolean addPorts, @DefaultValue("port.") String portsPrefix) {
+			@DefaultValue("true") boolean addPorts, @DefaultValue("port.") String portsPrefix, boolean addPodLabels,
+			boolean addPodAnnotations) {
+
+		@ConstructorBinding
+		public Metadata {
+
+		}
+
+		public Metadata(@DefaultValue("true") boolean addLabels, String labelsPrefix,
```
wind57, line 108: add a new constructor not to break the public existing one

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/discovery/DefaultKubernetesServiceInstance.java
```diff
@@ -48,7 +48,12 @@ public record DefaultKubernetesServiceInstance(String instanceId, String service
 	 */
 	public DefaultKubernetesServiceInstance(String instanceId, String serviceId, String host, int port,
 			Map<String, String> metadata, boolean secure) {
-		this(instanceId, serviceId, host, port, metadata, secure, null, null);
+		this(instanceId, serviceId, host, port, metadata, secure, null, null, Map.of(), Map.of());
+	}
+
+	public DefaultKubernetesServiceInstance(String instanceId, String serviceId, String host, int port,
```
wind57, line 54: keep the previous constructor intact and introduce a new one

wind57:
@ryanjbaxter ready to look at. thank you.

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/discovery/DefaultKubernetesServiceInstance.java
```diff
@@ -101,6 +106,16 @@ public String getCluster() {
 		return this.cluster;
 	}
 
+	@Override
+	public Map<String, String> podLabels() {
```
ryanjbaxter, line 110: Is there a reason to not just put this in the metadata?

wind57: I was afraid that service labels and pod labels could create a mess if put together, like what if labels colide? i thought that separating them would be cleaner.

ryanjbaxter: Yeah that is kind of what I was saying here....I suppose there is no other way, we either run the risk of colliding or we separate things.

Since we have some freedom here with this new property could we make it Map<String, Map<String, String> call it `podMetadata` and put both labels and annotations in that map?

wind57: my turn to miss-read!

```
Map<String, Map<String, String>
```

so something like : `labels = {a = b}` and `annotations = {c = d}`, right?

ryanjbaxter: correct

PR #1255, opened 2023-03-13
unique service names (2.1.x)

wind57:


wind57:
I can't understand why the PR checks are not triggered here :| will try again later in the day

ryanjbaxter:
I dont think we ever added github actions to the 2.1.x branch

wind57:
Ha! Indeed! :) You OK me creating a PR to add it?

ryanjbaxter:
Looks like `ActuatorRefreshRabbitMQIT.testRefresh` is failing.  Wouldnt think it would be related, but maybe.  Let me know if you have a chance to look at it.

wind57:
oh wow, I totally forgot that `2.1.x` is still on kind, is it still worth to create a github actions pipeline? it might take a while from my side

Looking into the test locally to see what is going on...

ryanjbaxter:
Actually it appears to be a fluke because the build passed this morning

wind57:
1) so still OK for me to create a pipeline where tests run without integration tests? 
2) indeed the tests run just fine for me locally too - just tested it.

ryanjbaxter:
>so still OK for me to create a pipeline where tests run without integration tests for 2.1.x?
yup!

PR #1256, opened 2023-03-13
add common pipeline for main and 2.1.x branches

wind57:


wind57:
to be honest, this PR is easier to review in reverse, look at the one to the `main` branch. Cause what I did is this:

- take the existing github actions set-up that we had for the `main` branch
- adapt it in such a manner that it can be used for both `2.1.x` and `main`
- create a minimal foundation for any _new_ potential branches 

There are two changes here that are not related to the pipeline itself, only in the examples module.

spring-cloud-kubernetes-examples/kubernetes-hello-world-example/pom.xml
```diff
@@ -9,7 +9,7 @@
 	</parent>
 	<modelVersion>4.0.0</modelVersion>
 
-	<artifactId>kubernetes-hello-world</artifactId>
+	<artifactId>kubernetes-hello-world-example</artifactId>
```
wind57, line 12: rename the module, we use one artifactId here, but in modules declaration of the parent pom we use a different name, as such when I try to do `-pl '-:kubernetes-hello-world'` or `-pl -:kubernetes-hello-world-example` to exclude this one from the build, maven is really confused and says that neither is present. A simple rename fixes this.

wind57:
@ryanjbaxter ready to be looked at, but as said the much easier way to review this one is to look [here](https://github.com/spring-cloud/spring-cloud-kubernetes/pull/1259). thank you.

test.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
```
ryanjbaxter, line 1: What is this file?

wind57: ah snap! sorry about that, by accident in here... removed

wind57: deleted and rebuild now. sorry about that!

ryanjbaxter:
I merged this into main, you might want to double check that I didn't mess up the merge though ;)

wind57:
yup, I'm keeping an eye right now on both builds...

wind57:
2.1.x failed, but because it could not get a dependency, I think a re-trigger would be needed

wind57:
main passed! all good there

ryanjbaxter:
Rerunning 2.1. now

PR #1259, opened 2023-03-14
Test new pipeline

wind57:


wind57:
this is verbatim a copy-paste of the entire `.github/workflows` folder from the branch that is supposed to go into `2.1.x`. The only reason this PR exists here, is that it will simplify reviewing it greatly. 

also notice that some files here are "Empty", the only thing that has changed was the rights on it. For whatever reason, my build was not triggered a few times and this seemed to have solved it? Not sure if it's really the case, but it's a ver minor thing anyway, imho.

.github/workflows/composites/env-variables/action.yaml
```diff
@@ -6,5 +6,19 @@ runs:
     - name: set env variables
       shell: bash
       run: |
+        
+        baseRef=$GITHUB_BASE_REF
```
wind57, line 10: this is by far not intuitive, but I hope that my comment here will make it clear why this is needed. We run the pipeline on two events : a PR and a push to the branch. In order to know if this needs to run on jdk-8 or jdk-17, I need to know what "base branch" is this going to. If it's `2.1.x` it must use `jdk-8`, if it's main, it must use jdk-17.

When this is a PR, things are easy, github provides: `GITHUB_BASE_REF`, so I know what branch is this PR supposed to go into. The problem is, that when this is a simple push (when you actually merge any PR), that variable is not available. So I am getting it from `${{ github.ref }}`. It needs a bit of parsing, cause it comes in the form of `xxx/yyy/<BRANCH>`.

I hope this makes sense.

.github/workflows/composites/env-variables/action.yaml
```diff
@@ -6,5 +6,19 @@ runs:
...
+        if [[ ! -z $baseRef ]]; then
+          echo "This is a PR to branch : $baseRef"
+          echo "BASE_BRANCH=$(echo $baseRef)" >> $GITHUB_ENV
+        else 
+          raw=${{ github.ref }}
+          echo "${raw}"
+          branch=${raw##*/}
+          echo "This is a merge to branch : ${branch}"
+          echo "BASE_BRANCH=$(echo ${branch})" >> $GITHUB_ENV
```
wind57, line 19: I then use `BASE_BRANCH` to run or not a certain step, for example.

.github/workflows/maven.yaml
```diff
@@ -2,26 +2,34 @@ name: github-workflow
 
 on:
   push:
-    branches: [ main ]
+    branches: [ main, 2.1.x ]
```
wind57, line 5: both branches that we still support are now taken into account

.github/workflows/maven.yaml
```diff
@@ -2,26 +2,34 @@ name: github-workflow
...
   pull_request:
-    branches: [ main ]
+    branches: [ main, 2.1.x ]
 
 jobs:
   build:
     runs-on: ubuntu-latest
     env:
       TESTCONTAINERS_REUSE_ENABLE: false
+    outputs:
```
wind57, line 14: in plain english this says: "if I ran on the step related to jdk-17 (which means base branch is `main`), set `jdk_17_branch=true`". I need this because this let's me skip fabric8-istio integration test, cause when we run `2.1.x` - we do not run integration tests. And fabric8-istio, unlike all other integration tests, is a separate step (we need this when we run on `main` as the base branch, basically to start a testcontainer and reuse it later).

.github/workflows/maven.yaml
```diff
@@ -33,27 +41,35 @@ jobs:
 
       - name: build controllers project
         uses: ./.github/workflows/composites/build-controllers-project
+        if: env.BASE_BRANCH == 'main'
```
wind57, line 44: skip a few steps that are strictly related to integration tests, that we do not run on `2.1.x`

.github/workflows/maven.yaml
```diff
@@ -78,7 +94,7 @@ jobs:
         uses: ./.github/workflows/composites/fabric8-istio-integration-test
 
   test:
-    needs: [ build, fabric8_istio_integration_test ]
```
wind57, line 81: we do not need this `fabric8_istio_integration_test`, because it runs one at a time anyway, we do not run the steps in parallel.

.github/workflows/maven.yaml
```diff
@@ -127,16 +150,34 @@ jobs:
...
-            PLAIN_TEST_CLASSNAMES=($(find . -name '*.java' \
-                        | grep 'src/test/java' \
-                        | grep -v 'Fabric8IstioIT' \
-                        | xargs grep -l '@Test' \
-                        | xargs grep -L 'abstract class' \
-                        | sed 's/.*src.test.java.//g' \
-                        | sed 's@/@.@g' \
-                        | sed 's/.\{5\}$//' \
-                        | tr '\n' ' '))
+            # exclude all integration tests for branch '2.1.x' via 'grep -v '.*IT.java''
```
wind57, line 153: this comments should make it clear why this `if/else` exists

.github/workflows/maven.yaml
```diff
@@ -227,17 +268,35 @@ jobs:
...
-                -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn \
-                -Dhttp.keepAlive=false \
-                -Dmaven.wagon.http.pool=false \
-                -Dmaven.wagon.http.retryHandler.class=standard \
-                -Dmaven.wagon.http.retryHandler.count=3 \
-                -Dskip.build.image=true
+          
+            if [[ $baseBranch == "2.1.x" ]]; then
+          
+              ./mvnw -s .settings.xml -pl '-:kubernetes-leader-election-example' -pl '-:kubernetes-hello-world-example' \
```
wind57, line 274: skip examples and controllers, otherwise the pipeline runs > 1h, and it gets killed by github

.github/workflows/maven.yaml
```diff
@@ -96,148 +114,191 @@ jobs:
...
-            # - only take classes that have @Test inside them
-            # - ignore the ones that have 'abstract class'. we do this because otherwise we would pass
-            #   to -DtestsToRun an abstract class, and it will not run anything.
-            # - drop the "begining" xxx/src/test/java
-            # - replace / with .
-            # - drop last ".java"
-            # - replace newline with space
-            # - replace '\n' with ' '
-
+
```
wind57, line 144: there are two changes only in this script:

- exclude all integration tests for `2.1.x` branch (this comment : `# exclude all integration tests for branch '2.1.x' via 'grep -v '.*IT.java''`)
- exclude some projects when building `2.1.x` and running tests, via:

```
if [[ $baseBranch == "2.1.x" ]]; then
          
            ./mvnw -s .settings.xml -pl '-:kubernetes-leader-election-example' -pl '-:kubernetes-hello-world-example' \
            -pl '-:kubernetes-reload-example' -pl '-:kubernetes-loadbalancer-example' \
            -pl '-:spring-cloud-kubernetes-configserver' -pl '-:spring-cloud-kubernetes-configuration-watcher' \
            -pl '-:spring-cloud-kubernetes-discoveryserver' \
```

wind57:
I am closing this one, because it will come in here when `2.1.x` is merged into `main`

PR #1262, opened 2023-03-16
Fix 513

wind57:


wind57:
an attempt to fix [this](https://github.com/spring-cloud/spring-cloud-kubernetes/issues/513) issue. Does not look that complicated at first sight...

ryanjbaxter:
This should also be fixed in 2.1.x

PR #1265, opened 2023-03-17
Fix 976

wind57:


spring-cloud-kubernetes-fabric8-discovery/src/main/java/org/springframework/cloud/kubernetes/fabric8/discovery/Fabric8ServicePortData.java
```diff
@@ -0,0 +1,23 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.kubernetes.fabric8.discovery;
+
+/**
+ * @author wind57
+ */
+record Fabric8ServicePortData(int portNumber, String portName) {
```
wind57, line 22: keep state of both number and name of the port that we find, as such, if it's `https` we can tell that this is a secured one

spring-cloud-kubernetes-fabric8-discovery/src/main/java/org/springframework/cloud/kubernetes/fabric8/discovery/ServicePortSecureResolver.java
```diff
@@ -49,23 +50,29 @@ class ServicePortSecureResolver {
...
 		}
 
-		if (port != null && properties.knownSecurePorts().contains(port)) {
-			logEntry(serviceName, port, "port is known to be a https port");
+		if (properties.knownSecurePorts().contains(portData.portNumber())) {
+			logEntry(serviceName, portData.portNumber(), "port is known to be a https port");
+			return true;
+		}
+
+		if ("https".equalsIgnoreCase(input.portData().portName())) {
```
wind57, line 74: if we know it's `https`, we can return `true` here, meaning it's secured

wind57:
let me know what you think of this one @ryanjbaxter

PR #1266, opened 2023-03-20
Fix 1182

wind57:


spring-cloud-kubernetes-fabric8-discovery/src/main/java/org/springframework/cloud/kubernetes/fabric8/discovery/Fabric8KubernetesDiscoveryClientUtils.java
```diff
@@ -208,7 +211,30 @@ else if (!properties.namespaces().isEmpty()) {
 					serviceName);
 		}
 
-		return endpoints;
+		return withFilter(endpoints, client, filter);
+	}
+
+	// see https://github.com/spring-cloud/spring-cloud-kubernetes/issues/1182 on why this
+	// is needed
+	static List<Endpoints> withFilter(List<Endpoints> initial, KubernetesClient client, Predicate<Service> filter) {
```
wind57, line 219: the core change is here. Here is the flow:

- you called `getServices` and it has returned to you a certain list : `[a, b]`. This searched happened potentially in many namespaces and filter was applied.
- you call now `getInstances` with those `[a, b, c]`, one at a time. 
- let's say you call `getInstances` with `a`. what we know is the namespace where `a` is hosted, so we can query all services in the same namespace, apply the filter on the services and see if an endpoints name matches service name. If it does, we need to return it. 
- This emulates the initial call to `getServices`, where the filter was applied. 
- please notice that the initial call to `getServices` also filters by labels, but I do not apply such a filter here in the call : 

```
client.services().inNamespace(entry.getKey()).list().getItems().stream()
					.filter(filter).map(service -> service.getMetadata().getName()).collect(Collectors.toSet());
```

This is because endpoints object has the same labels as the service, so when we search for endpoints, this filtering has already happened. 

I honestly hope this makes sense...

ryanjbaxter: I am not sure I understand why we need to fetch the services in order to filter the endpoints.  Wouldnt that mean we fetched the services and filtered them twice at this point?

wind57: Done

ryanjbaxter:
I guess technically we would also need a fix for the 2.1.x branch for this, the bug is also present there.

wind57:
sure thing boss! I'll create one against `2.1.x` branch

wind57:
So, ok to merge this one then, too?

PR #1272, opened 2023-03-24
Fix 1271

wind57:


ryanjbaxter:
We cannot bump to a new major of a dependency unless we do a major release so we cannot move to version 18

wind57:
right, I get that, that is why I would have put that (MAJOR RELEASE) thing in the header and leave it as a Draft

ryanjbaxter:
I would rather just wait to see whatever major version is available at the time we have a major rather than have this PR hanging around until whenever that may be

PR #1208, opened 2023-01-28
fix 1207 for kubernetes native client

wind57:

PR #1280, opened 2023-03-30
Fix 1278

wind57:


spring-cloud-kubernetes-fabric8-discovery/src/main/java/org/springframework/cloud/kubernetes/fabric8/discovery/KubernetesDiscoveryClientConfigClientBootstrapConfiguration.java
```diff
@@ -29,6 +32,7 @@
 @Configuration(proxyBeanMethods = false)
 @ConditionalOnProperty("spring.cloud.config.discovery.enabled")
 @Import({ Fabric8AutoConfiguration.class, KubernetesDiscoveryClientAutoConfiguration.class })
+@EnableConfigurationProperties({ KubernetesDiscoveryProperties.class, KubernetesClientProperties.class })
```
wind57, line 35: this is the core problem right here. this bootstrap imports two configurations:

- `Fabric8AutoConfiguration`, which internally requires `KubernetesClientProperties`
- `KubernetesDiscoveryClientAutoConfiguration` which internally requires `KubernetesDiscoveryProperties`

These work just fine on their own if used by auto-configuration _only_, but once we use bootstrap annotations like `@AutoConfigureBefore` break.

The one change that I did not account for was the fact that before [this change](https://github.com/spring-cloud/spring-cloud-kubernetes/pull/1208/files), we used to provide an explicit `KubernetesDiscoveryProperties` `@Bean` and I dropped it once I fixed that issue. Totally my mistake here.

---

The other one with `KubernetesClientProperties` was still present, but as said in the issue itself, we got very lucky that no one actually stepped on it. 

It's kind of sad that I spent so much time thinking into not breaking anything once I make my changes, and it still happens...

spring-cloud-kubernetes-integration-tests/spring-cloud-kubernetes-fabric8-discovery-with-bootstrap/src/main/resources/bootstrap.yaml
```diff
@@ -0,0 +1,16 @@
+spring:
```
wind57, line 1: I've added an integration test with bootstrap enabled, that should catch these kind of things in the future.

wind57:
@ryanjbaxter ready to be looked at. thank you.

Issue #1257, opened 2023-03-14
kubernetes-leader-election-example: "No Node Port" error


shoner-sul:
**Describe the bug**
Following https://github.com/spring-cloud/spring-cloud-kubernetes/tree/main/spring-cloud-kubernetes-examples/kubernetes-leader-election-example , getting service URL failed with "No Node Port" error as below:

[spring-cloud-kubernetes-examples]$ kubectl apply -f deploy.yml 
service/kubernetes-leader-election-example created
serviceaccount/kubernetes-leader-election-example created
rolebinding.rbac.authorization.k8s.io/kubernetes-leader-election-example:view created
role.rbac.authorization.k8s.io/namespace-reader created
deployment.apps/kubernetes-leader-election-example created

[spring-cloud-kubernetes-examples]$ minikube service kubernetes-leader-election-example --url
  service default/kubernetes-leader-election-example has no node port

[spring-cloud-kubernetes-examples]$ minikube service list
|-------------|------------------------------------|--------------|-----|
|  NAMESPACE  |                NAME                | TARGET PORT  | URL |
|-------------|------------------------------------|--------------|-----|
| default     | kubernetes                         | No node port |
| default     | kubernetes-leader-election-example | No node port |
| kube-system | kube-dns                           | No node port |
|-------------|------------------------------------|--------------|-----|

**Sample**
Used https://github.com/spring-cloud/spring-cloud-kubernetes/tree/main/spring-cloud-kubernetes-examples/kubernetes-leader-election-example in main branch as is. "deploy.yml" file in the description above is a file created as instructed under "You can then deploy the application using the following yaml."

shoner-sul:
github authentication problem. closing.

ryanjbaxter:
Might want to read through this
https://minikube.sigs.k8s.io/docs/handbook/accessing/

shoner-sul:
Before asking here, I've read the same document and tried step 1~4 of "Example of NodePort" part. It resulted same error "Connection refused". (I guessed that it was because of missing role settings in deployment process, which was present in YML file.)

It says in step 4 that there is a ssh tunnel, but ps command does not show any ssh process.

[shoner@shoner spring-cloud-kubernetes-examples]$ ps -ef | grep docker@127.0.0.1
shoner   1010161 1008796  0 06:10 pts/2    00:00:00 grep --color=auto docker@127.0.0.1
[...]

PR #1281, opened 2023-03-31
Update leader election sample

ryanjbaxter:


shoner-sul (commented):
spec.type: ClusterIP -> NodePort required
https://github.com/spring-cloud/spring-cloud-kubernetes/issues/1257

PR #1283, opened 2023-04-04
Increase timeout for integration tests

wind57:


spring-cloud-kubernetes-test-support/src/main/java/org/springframework/cloud/kubernetes/integration/tests/commons/native_client/Util.java
```diff
@@ -455,7 +455,7 @@ private String secretName(V1Secret secret) {
 
 	private void waitForDeployment(String namespace, V1Deployment deployment) {
 		String deploymentName = deploymentName(deployment);
-		await().pollInterval(Duration.ofSeconds(1)).atMost(600, TimeUnit.SECONDS)
```
wind57, line 458: This is needed mainly because of the `ActuatorRefreshRabbitMQIT` - which is a very slow test. Locally, in my Mac M1, it takes around 4 minutes to run. But on github, not even 10 is enough. As such, increase it to 900 seconds.

wind57:
@ryanjbaxter this small PR will fix the existing build. thank you for looking into it.

PR #1284, opened 2023-04-05
Test via non deprecated constructor

wind57:


spring-cloud-kubernetes-client-discovery/src/test/java/org/springframework/cloud/kubernetes/client/discovery/KubernetesInformerDiscoveryClientTests.java
```diff
@@ -83,7 +83,7 @@ void testServiceWithUnsetPortNames() {
 		Lister<V1Service> serviceLister = setupServiceLister(SERVICE_1);
 		Lister<V1Endpoints> endpointsLister = setupEndpointsLister(ENDPOINTS_NO_UNSET_PORT_NAME);
 
-		KubernetesInformerDiscoveryClient discoveryClient = new KubernetesInformerDiscoveryClient("",
```
wind57, line 86: this test was using the now deprecated constructor, so just was the actual one

PR #1285, opened 2023-04-05
Deprecate constructor that uses namespace

wind57:


wind57:
@ryanjbaxter one more minor PR before two more far more interesting ones, imho. thank you!

spring-cloud-kubernetes-client-discovery/src/main/java/org/springframework/cloud/kubernetes/client/discovery/reactive/KubernetesInformerReactiveDiscoveryClientAutoConfiguration.java
```diff
@@ -85,4 +86,15 @@ public KubernetesInformerReactiveDiscoveryClient kubernetesReactiveDiscoveryClie
 				serviceLister, endpointsLister, serviceInformer, endpointsInformer, properties);
 	}
 
+	@Bean
+	@ConditionalOnMissingBean
+	KubernetesInformerReactiveDiscoveryClient kubernetesClientReactiveDiscoveryClient(
```
ryanjbaxter, line 91: We shouldnt be creating 2 bean instances, we can remove the `@Bean` annotation above

ryanjbaxter (changes requested):

spring-cloud-kubernetes-client-discovery/src/main/java/org/springframework/cloud/kubernetes/client/discovery/reactive/KubernetesInformerReactiveDiscoveryClientAutoConfiguration.java
wind57: there's `@ConditionalOnMissingBean` on both of them, so only one will be created. We don't care which one will be created at this point in time, because they do the same thing in their constructors. 

Later, in the next major release, I will have a PR that drops all of these deprecations. 

Btw, the same situation exists in `KubernetesInformerDiscoveryClientAutoConfiguration` (the non-reactive auto-configuration).

wind57: Or even better now that I think about your suggestion is to remove the `@Bean` in both reactive and non-reactive auto-configurations. This is one nice piece of advice, thank you.

ryanjbaxter: Exactly

PR #1288, opened 2023-04-07
Remove k8s client spring integration dependency

wind57:


spring-cloud-kubernetes-client-autoconfig/pom.xml
```diff
@@ -24,10 +24,6 @@
 			<groupId>io.kubernetes</groupId>
 			<artifactId>client-java-extended</artifactId>
 		</dependency>
-		<dependency>
```
wind57, line 27: this is an easy fix that took almost a day to figure out. :) While working on a different thing (add support for selective namespaces in discovery k8s-client), I wrote some integration tests that were constantly failing. After investigation, I realized that some beans that we create, for example:

```
	@Bean
	@ConditionalOnMissingBean
	public SharedInformerFactory sharedInformerFactory(ApiClient client) {
		return new SharedInformerFactory(client);
	}
```

were never created by us, but were coming from this maven dependency. We do not require it anymore, since we don't use anything from it (the vast things that we were using were deprecated in it, so it refactored in a different PR anyway).

wind57:
@garyrussell one more easy one, thank you.

PR #1290, opened 2023-04-07
Fix properties retrieval order from configmaps

disimo74:
The properties defined by the configmap "application" should be the last in the environment list, so you can always override the properties values by profile and by application


wind57:
this shows a lot of file changes, you need to checkout branch `2.1.x` and create a PR from there, you probably created it from `main` :)

PR #1291, opened 2023-04-07
Fix properties retrieval order from configmaps

disimo74:
The properties defined by the configmap "application" should be the last in the environment list, so you can always override the properties values by profile and by application

### DIFF
--- a/spring-cloud-kubernetes-controllers/spring-cloud-kubernetes-configserver/src/test/java/org/springframework/cloud/kubernetes/configserver/KubernetesEnvironmentRepositoryTests.java
+++ b/spring-cloud-kubernetes-controllers/spring-cloud-kubernetes-configserver/src/test/java/org/springframework/cloud/kubernetes/configserver/KubernetesEnvironmentRepositoryTests.java
@@ -244,4 +244,29 @@ class KubernetesEnvironmentRepositoryTests {
 		});
 	}
 
+	@Test
+	public void testApplicationPropertiesAnSecretsOverride() throws ApiException {
+		CoreV1Api coreApi = mock(CoreV1Api.class);
+		when(coreApi.listNamespacedConfigMap(eq("default"), eq(null), eq(null), eq(null), eq(null), eq(null), eq(null),
+			eq(null), eq(null), eq(null), eq(null))).thenReturn(CONFIGMAP_DEFAULT_LIST);
+		when(coreApi.listNamespacedSecret(eq("default"), eq(null), eq(null), eq(null), eq(null), eq(null), eq(null),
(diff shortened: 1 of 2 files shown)


wind57:
I guess some tests would nice to have here...

spring-cloud-kubernetes-controllers/spring-cloud-kubernetes-configserver/src/test/java/org/springframework/cloud/kubernetes/configserver/KubernetesEnvironmentRepositoryTests.java
```diff
@@ -244,4 +244,26 @@ public void testStoresProfileCase() throws ApiException {
...
+		CoreV1Api coreApi = mock(CoreV1Api.class);
+		when(coreApi.listNamespacedConfigMap(eq("default"), eq(null), eq(null), eq(null), eq(null), eq(null), eq(null),
+			eq(null), eq(null), eq(null), eq(null))).thenReturn(CONFIGMAP_DEFAULT_LIST);
+		when(coreApi.listNamespacedSecret(eq("default"), eq(null), eq(null), eq(null), eq(null), eq(null), eq(null),
+			eq(null), eq(null), eq(null), eq(null))).thenReturn(SECRET_LIST);
+		when(coreApi.listNamespacedConfigMap(eq("dev"), eq(null), eq(null), eq(null), eq(null), eq(null), eq(null),
+			eq(null), eq(null), eq(null), eq(null))).thenReturn(CONFIGMAP_DEV_LIST);
+		KubernetesEnvironmentRepository environmentRepository = new KubernetesEnvironmentRepository(coreApi,
+			kubernetesPropertySourceSuppliers, "default");
+		Environment environment = environmentRepository.findOne("stores-dev", "", "");
```
wind57, line 258: I'm not sure I agree with this test. Don't you need to assert _order_ here? Cause right now you are asserting only for presence.

disimo74: I can change the test checking that the last configmap on the list is "application" what I want with this test is to allow to override the properties in "application" configmap with properties in more specific configmaps.
So without the fix this test fails  because the fist configmap retrieved is the "application" configmap.

wind57: ah! indeed, that makes sense.

PR #1293, opened 2023-04-07
Fix incorrect pod data for reactive k8s discovery

wind57:


spring-cloud-kubernetes-client-discovery/src/main/java/org/springframework/cloud/kubernetes/client/discovery/reactive/KubernetesInformerReactiveDiscoveryClientAutoConfiguration.java
```diff
@@ -69,7 +69,7 @@ public ReactiveDiscoveryClientHealthIndicator kubernetesReactiveDiscoveryClientH
 			KubernetesClientPodUtils podUtils) {
 		ReactiveDiscoveryClientHealthIndicator healthIndicator = new ReactiveDiscoveryClientHealthIndicator(client,
 				properties);
-		InstanceRegisteredEvent<?> event = new InstanceRegisteredEvent<>(podUtils.currentPod(), null);
+		InstanceRegisteredEvent<?> event = new InstanceRegisteredEvent<>(podUtils.currentPod().get(), null);
```
wind57, line 72: this is the existing bug we have in our code base. Instead of publishing the actual Pod, we publish the `Supplier<V1Pod>`. Everything else are just test changes because of that

wind57:
this was already fixed elsewhere

Issue #1301, opened 2023-04-12
reactive k8s discovery is not working if blocking is disabled


wind57:
The scenario is this:

- disable blocking implementation of k8s discovery
- leave only reactive present

The outcome of this is that `getServices` will always return an empty list, it will not find any services.

The explanation for this:

For services to be found, we need `KubernetesInformerDiscoveryClient`, that needs to call its : 

```
@PostConstruct
	public void afterPropertiesSet() {
            sharedInformerFactory.startAllRegisteredInformers();
            ....
        }
```

This can be called only when such a `@Bean` will be created. If we disabled blocking, such a bean is never created, instead we call constructor directly:

```
	KubernetesInformerReactiveDiscoveryClient(SharedInformerFactory sharedInformerFactory,
			Lister<V1Service> serviceLister, Lister<V1Endpoints> endpointsLister,
			SharedInformer<V1Service> serviceInformer, SharedInformer<V1Endpoints> endpointsInformer,
			KubernetesDiscoveryProperties properties) {
		this.kubernetesDiscoveryClient = new KubernetesInformerDiscoveryClient(sharedInformerFactory, serviceLister,
				endpointsLister, serviceInformer, endpointsInformer, properties);
	}
```

@ryanjbaxter this is a bug. I'll fix it, of course.

Issue #1298, opened 2023-04-09
k8s discovery client inconsistent between reactive and blocking health indicators


wind57:
If you look in blocking vs reactive implementations of health of k8s native client, we currently do things differently. What I mean by that is, in `KubernetesInformerReactiveDiscoveryClientAutoConfiguration`:

```
ReactiveDiscoveryClientHealthIndicator healthIndicator = new ReactiveDiscoveryClientHealthIndicator(client,
				properties);
		InstanceRegisteredEvent<?> event = new InstanceRegisteredEvent<>(podUtils.currentPod(), null);
		healthIndicator.onApplicationEvent(event);
```

But in `KubernetesInformerDiscoveryClientAutoConfiguration` we use `KubernetesDiscoveryClientHealthIndicatorInitializer`, that has a post-construct in the form:

```
	@PostConstruct
	private void postConstruct() {
		LOG.debug(() -> "publishing InstanceRegisteredEvent");
		this.applicationEventPublisher.publishEvent(new InstanceRegisteredEvent<>(
				new RegisteredEventSource("kubernetes", podUtils.isInsideKubernetes(), podUtils.currentPod().get()),
				null));
	}
```

How come one publishes an event and one does not? And how come the blocking health does not have a bean of `DiscoveryClientHealthIndicator` to be on par with the reactive one? 

I am inclined to make them behave exactly the same, as I see this as a tiny bug. 

@ryanjbaxter I have several issues opened around health indicators that I would like to fix, but this is the first one I would like to get your input on. Thank you.

ryanjbaxter:
I know I wrote this code but I cannot for the life of me figure out why I published an event in one case and just called the method in the other.  Maybe it has to do with the underlying implementation in the reactive and blocking cases...not sure.  To me it seems better to publish the event in both cases.

wind57:
That's OK, that's why I'm here (I guess), to figure out these things and ask for your wisdom. 

I see things slightly differently. imo, we need to both to publish the event (so that a potential custom users `ApplicationListener` can listen to that) and provide a bean of type `ReactiveDiscoveryClientHealthIndicator/DiscoveryClientHealthIndicator`, so that user's `/health` endpoint would be correct. 

There are some other things that imho need to be done here, but I'll present a PR and we can debate them via the code if you do not mind.

wind57:
[...]

Issue #1297, opened 2023-04-08
Inconsistent response from reactive and blocking health indicators in k8s discovery client


wind57:
If you take a look in k8s-native discovery client `KubernetesInformerDiscoveryClientAutoConfiguration`:

```
	@Bean
	@ConditionalOnClass({ HealthIndicator.class })
	@ConditionalOnDiscoveryHealthIndicatorEnabled
	public KubernetesDiscoveryClientHealthIndicatorInitializer indicatorInitializer(
			ApplicationEventPublisher applicationEventPublisher, PodUtils<?> podUtils) {
		return new KubernetesDiscoveryClientHealthIndicatorInitializer(podUtils, applicationEventPublisher);
	}
```

we create a bean `KubernetesDiscoveryClientHealthIndicatorInitializer` that has a `@PostConstruct`:

```
	@PostConstruct
	private void postConstruct() {
		LOG.debug(() -> "publishing InstanceRegisteredEvent");
		this.applicationEventPublisher.publishEvent(new InstanceRegisteredEvent<>(
				new RegisteredEventSource("kubernetes", podUtils.isInsideKubernetes(), podUtils.currentPod().get()),
				null));
	}

	public record RegisteredEventSource(String cloudPlatform, boolean inside, Object pod) {

	}
```

that is: it publishes an instance of `RegisteredEventSource` as the source from `InstanceRegisteredEvent`.

On the other hand if you take a look at the reactive implementation in `KubernetesInformerReactiveDiscoveryClientAutoConfiguration`:

```
	@Bean
	@ConditionalOnClass(name = "org.springframework.boot.actuate.health.ReactiveHealthIndicator")
	@ConditionalOnDiscoveryHealthIndicatorEnabled
	public ReactiveDiscoveryClientHealthIndicator kubernetesReactiveDiscoveryClientHealthIndicator(
			KubernetesInformerReactiveDiscoveryClient client, DiscoveryClientHealthIndicatorProperties properties,
			KubernetesClientPodUtils podUtils) {
		ReactiveDiscoveryClientHealthIndicator healthIndicator = new ReactiveDiscoveryClientHealthIndicator(client,
				properties);
		InstanceRegisteredEvent<?> event = new InstanceRegisteredEvent<>(podUtils.currentPod(), null);
		healthIndicator.onApplicationEvent(event);
		return healthIndicator;
	}
```

The source of `InstanceRegisteredEvent` is very different. 

---

[...]

PR #1304, opened 2023-04-12
Fix some existing issues

wind57:


spring-cloud-kubernetes-client-discovery/src/main/java/org/springframework/cloud/kubernetes/client/discovery/KubernetesInformerDiscoveryClient.java
```diff
@@ -209,8 +209,12 @@ private int findEndpointPort(List<CoreV1EndpointPort> endpointPorts, String prim
 
 	@Override
 	public List<String> getServices() {
-		return serviceLister.list().stream().filter(service -> matchesServiceLabels(service, properties))
-				.map(s -> s.getMetadata().getName()).distinct().toList();
+		List<String> services = serviceLister.list().stream()
```
wind57, line 212: just add a debug statement to easier track what is returned

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/discovery/KubernetesDiscoveryClientHealthIndicatorInitializer.java
```diff
@@ -45,9 +45,10 @@ public KubernetesDiscoveryClientHealthIndicatorInitializer(PodUtils<?> podUtils,
 	@PostConstruct
 	private void postConstruct() {
 		LOG.debug(() -> "publishing InstanceRegisteredEvent");
-		this.applicationEventPublisher.publishEvent(new InstanceRegisteredEvent<>(
```
wind57, line 48: nothing changes here, just minor code adjustments, easier to read code

spring-cloud-kubernetes-client-discovery/src/main/java/org/springframework/cloud/kubernetes/client/discovery/reactive/KubernetesInformerReactiveDiscoveryClientAutoConfiguration.java
```diff
@@ -84,13 +95,43 @@ public KubernetesInformerReactiveDiscoveryClient kubernetesReactiveDiscoveryClie
 				serviceLister, endpointsLister, serviceInformer, endpointsInformer, properties);
 	}
 
+	/**
+	 * Post an event so that health indicator is initialized.
+	 */
+	@Bean
```
wind57, line 101: publish an event to be on par with the blocking implementation, also fixes this one: https://github.com/spring-cloud/spring-cloud-kubernetes/issues/1298

spring-cloud-kubernetes-client-discovery/src/main/java/org/springframework/cloud/kubernetes/client/discovery/reactive/KubernetesInformerReactiveDiscoveryClientAutoConfiguration.java
```diff
@@ -84,13 +95,43 @@ public KubernetesInformerReactiveDiscoveryClient kubernetesReactiveDiscoveryClie
...
+	@ConditionalOnClass(name = "org.springframework.boot.actuate.health.ReactiveHealthIndicator")
+	@ConditionalOnDiscoveryHealthIndicatorEnabled
+	KubernetesDiscoveryClientHealthIndicatorInitializer reactiveIndicatorInitializer(
+			ApplicationEventPublisher applicationEventPublisher, PodUtils<?> podUtils) {
+		LOG.debug(() -> "Will publish InstanceRegisteredEvent from reactive implementation");
+		return new KubernetesDiscoveryClientHealthIndicatorInitializer(podUtils, applicationEventPublisher);
```
wind57, line 107: Since both blocking and reactive code delegates to the same code for the publishing, it also fixes this one: https://github.com/spring-cloud/spring-cloud-kubernetes/issues/1297

spring-cloud-kubernetes-client-discovery/src/main/java/org/springframework/cloud/kubernetes/client/discovery/reactive/KubernetesInformerReactiveDiscoveryClientAutoConfiguration.java
```diff
@@ -84,13 +95,43 @@ public KubernetesInformerReactiveDiscoveryClient kubernetesReactiveDiscoveryClie
...
+		return new KubernetesInformerReactiveDiscoveryClient(kubernetesClientInformerDiscoveryClient);
+	}
+
+	@Bean
+	@ConditionalOnMissingBean
+	KubernetesInformerDiscoveryClient kubernetesClientInformerDiscoveryClient(
 			SharedInformerFactory sharedInformerFactory, Lister<V1Service> serviceLister,
 			Lister<V1Endpoints> endpointsLister, SharedInformer<V1Service> serviceInformer,
 			SharedInformer<V1Endpoints> endpointsInformer, KubernetesDiscoveryProperties properties) {
-		return new KubernetesInformerReactiveDiscoveryClient(sharedInformerFactory, serviceLister, endpointsLister,
```
wind57, line 93: create a bean (thus fixing : https://github.com/spring-cloud/spring-cloud-kubernetes/issues/1301 )

spring-cloud-kubernetes-client-discovery/src/main/java/org/springframework/cloud/kubernetes/client/discovery/reactive/KubernetesInformerReactiveDiscoveryClientAutoConfiguration.java
```diff
@@ -84,13 +95,43 @@ public KubernetesInformerReactiveDiscoveryClient kubernetesReactiveDiscoveryClie
...
+		return new KubernetesDiscoveryClientHealthIndicatorInitializer(podUtils, applicationEventPublisher);
+	}
+
+	/**
+	 * unlike the blocking implementation, we need to register the health indicator.
+	 */
+	@Bean
+	@ConditionalOnClass(name = "org.springframework.boot.actuate.health.ReactiveHealthIndicator")
+	@ConditionalOnDiscoveryHealthIndicatorEnabled
+	ReactiveDiscoveryClientHealthIndicator kubernetesReactiveDiscoveryClientHealthIndicator(
```
wind57, line 116: it took forever to understand what is going on. It has to do with the fact that `org.springframework.cloud.client.CommonsClientAutoConfiguration` and `org.springframework.cloud.client.ReactiveCommonsClientAutoConfiguration` have a different way of working. One expects `ReactiveDiscoveryClientHealthIndicator` to be provided, and the blocking one: creates them based on the `DiscoveryClient` bean already in the context. 

This different approach means different approaches in our case too. Not obvious (not very pleasant either), but not a big deal. I can go into a lot more details if needed, just let me know. But in general, your approach that was already here makes total sense now.

spring-cloud-kubernetes-integration-tests/spring-cloud-kubernetes-client-discovery-it/src/test/java/org/springframework/cloud/kubernetes/client/discovery/KubernetesClientDiscoveryClientIT.java
```diff
@@ -187,6 +188,8 @@ void testAllNamespaces() {
 	 * </pre>
 	 */
 	@Test
+	@Disabled
```
wind57, line 191: disable a test for now - it will be enabled when the underlying issue is going to be fixed.

I either disable it or fix that issue in this PR, but then it will grow quite a lot and be harder to review. 

It's on my next TODO anyway. If you are OK with such an approach...

wind57:
@ryanjbaxter  ready to be looked at. this fixes 4 minor existing issues. let me know if there are any questions! thank you

spring-cloud-kubernetes-test-support/src/main/java/org/springframework/cloud/kubernetes/integration/tests/commons/native_client/Util.java
```diff
@@ -132,6 +132,9 @@ public void createAndWait(String namespace, String name, V1Deployment deployment
 			}
 		}
 		catch (Exception e) {
+			if (e instanceof ApiException apiException) {
+				System.out.println(apiException.getResponseBody());
```
ryanjbaxter, line 136: Do we need this?

ryanjbaxter (approved):

spring-cloud-kubernetes-test-support/src/main/java/org/springframework/cloud/kubernetes/integration/tests/commons/native_client/Util.java
wind57: well, I'd say yes. This can be printed via a Logger, but since `kubectl logs` takes stdout, it does not really matter. Especially since we use this for tests only. 

This `ApiException` is a bit weird, imho. If you do `apiException.printStackTrace()` - you will get a stack trace, but its going to be almost meaningless, the "real" issue is under `apiException.getResponseBody` - that is why I will slowly change this everywhere in our tests.

wind57:
this is some really unexpected failure, taking a look...

wind57:
@ryanjbaxter ready! in the end it was fairly easy (if you want I can go into the details), but TL;DR: it was a miss-configuration in the libraries that I cleaned-up in the previous PR and we got really lucky it did not fail once we merged to main...

spring-cloud-kubernetes-integration-tests/spring-cloud-kubernetes-configuration-watcher-it/pom.xml
```diff
@@ -49,33 +49,48 @@
...
 		<dependency>
 			<groupId>io.kubernetes</groupId>
 			<artifactId>client-java-extended</artifactId>
-			<scope>test</scope>
 		</dependency>
-
 		<dependency>
-			<groupId>org.testcontainers</groupId>
-			<artifactId>k3s</artifactId>
+			<groupId>com.github.docker-java</groupId>
```
ryanjbaxter, line 74: Can you elaborate on whats happening here?

wind57: the failure reported initially was that a container could not be started, because port 9092 was already in use. 

We should not see that, because we start a container and re-use it in all integration tests, via two things:

- `withReuse(true)` when creating it
- `echo "testcontainers.reuse.enable=true" > ~/.testcontainers.properties`

So the question I had, why was an integration test trying to create a new container, when one was already present?

The answer is that testcontainers creates a hash from a running container and tries to re-use that on in every subsequent usage. The issue we faced, is because I removed `com.github.docker-java` (that is used when computing that hash), the hash computed was different, so no re-use. To make it simpler:

- `testA` _has_ `com.github.docker-java` dependency, creates hash = 123
- `testB` _does not_ have the `com.github.docker-java` dependency, creates hash = 456
- `testC` _has_ the `com.github.docker-java` dependency, creates hash = 123

My fix here was to simply revert the dependencies, but the more general fix and far more stable is to clean-up every integration test, which I will try to do.

ryanjbaxter: Meaning you want to make changes so every integration test is not required to have the dependency?

wind57: yes, that is exactly what I was thinking of.

ryanjbaxter: Wouldn't just placing the dependency in the integration tests parent Pom do that?

wind57: we don't need it, at all. we can remove it from all integration tests and parent pom. As a matter of fact, we can get rid from other dependencies too. I already have a PR [here](https://github.com/spring-cloud/spring-cloud-kubernetes/pull/1312) where I started this work.

ryanjbaxter: OK so by removing it from all tests then the hash would be the same and we would reuse the container?

ryanjbaxter: Great thanks for the explanation

PR #1322, opened 2023-04-24
Fix 990

wind57:


spring-cloud-kubernetes-client-discovery/src/main/java/org/springframework/cloud/kubernetes/client/discovery/KubernetesInformerDiscoveryClient.java
```diff
@@ -72,6 +74,8 @@ public class KubernetesInformerDiscoveryClient implements DiscoveryClient {
 
 	private final KubernetesDiscoveryProperties properties;
 
+	private final Predicate<V1Service> filter;
```
wind57, line 77: introduce a filter based on the SPELExpression, just like we have in fabric8.

wind57:
@ryanjbaxter fixes a bug this one. thank you.

PR #1327, opened 2023-04-27
Fix 1323 part 1

wind57:


spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/reload/ConfigurationUpdateStrategy.java
```diff
@@ -26,6 +26,13 @@
  */
 public record ConfigurationUpdateStrategy(String name, Runnable reloadProcedure) {
 
+	/**
+	 * Strategy that does nothing.
+	 */
+	public static final ConfigurationUpdateStrategy NOOP = new ConfigurationUpdateStrategy("no-op", () -> {
```
wind57, line 32: this is needed for configuration watcher, when reload is disabled. In that case, we do not care about a `ConfigurationUpdateStrategy`. The only reason we need it, is because we extend various classes that need it via the calls to `super`, but in the configuration watcher itself, we make no use of it. 

I can go into the details here, if needed, just let me know.

spring-cloud-kubernetes-integration-tests/spring-cloud-kubernetes-configuration-watcher-it/src/test/java/org/springframework/cloud/kubernetes/configuration/watcher/ActuatorRefreshIT.java
```diff
@@ -109,13 +115,68 @@ void testActuatorRefresh() {
...
+		V1EnvVar commonsDebug = new V1EnvVar()
+				.name("LOGGING_LEVEL_ORG_SPRINGFRAMEWORK_CLOUD_KUBERNETES_COMMONS_CONFIG_RELOAD").value("DEBUG");
+		V1EnvVar watcherDebug = new V1EnvVar()
+				.name("LOGGING_LEVEL_ORG_SPRINGFRAMEWORK_CLOUD_KUBERNETES_CONFIGURATION_WATCHER").value("DEBUG");
+
+		envVars.add(commonsDebug);
+		envVars.add(watcherDebug);
+
+		if (disableReload) {
+			V1EnvVar disableReloadEnvVar = new V1EnvVar().name("SPRING_CLOUD_KUBERNETES_RELOAD_ENABLED").value("FALSE");
```
wind57, line 175: disable reload: `SPRING_CLOUD_KUBERNETES_RELOAD_ENABLED` - app still works. This test fails, without the fix that we did here. 

I've also "tested" a build with disabling reload everywhere - that is, in the configuration watcher `application.yaml` and it passed all the integration tests.

wind57:
@ryanjbaxter part-1 of the recent issue we discussed recently. I've separated these in two PRs (one coming later), because it is easier to review and debate if needed. Thank you for looking into it.

ryanjbaxter:
Is the second pr dependent on this one?

wind57:
no. 

the other issue is the fact that we react and schedule a reload event even if nothing changed in the config maps/secrets. They are totally independent, to me.

ryanjbaxter:
OK let me know when the other PR is ready and we can merge these both together

PR #1329, opened 2023-04-27
Fix 1323 part 2

wind57:


spring-cloud-kubernetes-integration-tests/spring-cloud-kubernetes-fabric8-client-secrets-event-reload/src/test/resources/secret.yaml
```diff
@@ -6,4 +6,4 @@ metadata:
 data:
   # from.properties.key=initial
   application.properties: |
-    ZnJvbS5wcm9wZXJ0aWVzLmtleT1pbml0aWFsCg==
```
wind57, line 9: `Cg==` is a space... yeah, took a bit of har pulling to understand where my test was failing.

spring-cloud-kubernetes-fabric8-config/src/main/java/org/springframework/cloud/kubernetes/fabric8/config/reload/Fabric8EventBasedSecretsChangeDetector.java
```diff
@@ -133,7 +134,12 @@ public void onAdd(Secret secret) {
 		public void onUpdate(Secret oldSecret, Secret newSecret) {
 			LOG.debug("Secret " + newSecret.getMetadata().getName() + " was updated in namespace "
 					+ newSecret.getMetadata().getNamespace());
-			onEvent(newSecret);
+			if (Objects.equals(oldSecret.getData(), newSecret.getData())) {
```
wind57, line 137: we don't really care if anything but `data` has changed, so trigger a reload when data changes.

spring-cloud-kubernetes-client-config/src/main/java/org/springframework/cloud/kubernetes/client/config/reload/KubernetesClientEventBasedSecretsChangeDetector.java
```diff
@@ -90,6 +98,21 @@ public void onDelete(V1Secret secret, boolean deletedFinalStateUnknown) {
 					+ secret.getMetadata().getNamespace());
 			onEvent(secret);
 		}
+
+		private static boolean equals(Map<String, byte[]> left, Map<String, byte[]> right) {
```
wind57, line 102: needed because the value coming from `getData` is `Map<String, byte[]>`

spring-cloud-kubernetes-client-config/src/test/java/org/springframework/cloud/kubernetes/client/config/reload/KubernetesClientEventBasedSecretsChangeDetectorTests.java
```diff
@@ -71,6 +65,12 @@
  */
 class KubernetesClientEventBasedSecretsChangeDetectorTests {
```
wind57, line 67: I slightly adjusted this test, to make it more readable

spring-cloud-kubernetes-client-config/src/main/java/org/springframework/cloud/kubernetes/client/config/reload/KubernetesClientEventBasedConfigMapChangeDetector.java
```diff
@@ -81,7 +82,12 @@ public void onAdd(V1ConfigMap configMap) {
 		public void onUpdate(V1ConfigMap oldConfigMap, V1ConfigMap newConfigMap) {
 			LOG.debug(() -> "ConfigMap " + newConfigMap.getMetadata().getName() + " was updated in namespace "
 					+ newConfigMap.getMetadata().getNamespace());
-			onEvent(newConfigMap);
+			if (Objects.equals(oldConfigMap.getData(), newConfigMap.getData())) {
```
wind57, line 85: this is where your input is really needed @ryanjbaxter  . Until now, even without a change in the `data`, we would schedule an event, kind of. I'll try to explain.

Let's consider plain reload, a reload without configuration watcher.
Something changes in the configmap, a label is added (`data` does not change). We catch the change here, go to `onEvent` and there check if `data` has changed. We do that by comparing property sources, but they only care about the thing that resides inside the `data`, it does not care about `labels`. As such, since there is no change, nothing happens.
So this check here, is simply a short-cut for the check that we will be doing anyway later in the code. I see it as a good thing. 

Same example, but with configuration watcher image. There, we override the `onEvent`. So before this change, we did not care about what actually has changed, we would still schedule an event (even if only `labels` change like in the example above). With this change, the event will not be scheduled anymore. 
Now, it depends how you look at this. If this is something we should "fix", the fix in the `if (Objects.equals(oldConfigMap.getData(), newConfigMap.getData()))` is fine.

On the other hand, if you see this as a breaking change - we need to feature flag it, for example via a property:

```
spring.cloud.kubernetes.realod.watch-data-only
```

for example and only enable the check if this flag is `true`.

ryanjbaxter: No i dont view it as a breaking change because in reality at the end of the day the application receiving the reload notification just cares about the data and not anything else.

spring-cloud-kubernetes-fabric8-config/src/main/java/org/springframework/cloud/kubernetes/fabric8/config/reload/Fabric8EventBasedSecretsChangeDetector.java
ryanjbaxter: Does the Fabric8 client not return secrets data as a byte array like K8S client does?

PR #1330, opened 2023-04-30
Fix paths documentation

wind57:


docs/src/main/asciidoc/property-source-config.adoc
```diff
@@ -583,9 +583,17 @@ Another option for using `ConfigMap` instances is to mount them into the Pod by
 and having Spring Cloud Kubernetes read them from the file system.
 This behavior is controlled by the `spring.cloud.kubernetes.config.paths` property. You can use it in
 addition to or instead of the mechanism described earlier.
-You can specify multiple (exact) file paths in `spring.cloud.kubernetes.config.paths` by using the `,` delimiter.
```
wind57, line 586: The problem is in this sentence:

> You can specify multiple (exact) file paths in `spring.cloud.kubernetes.config.paths` by using the `,` delimiter.

It was probably left in the documentation from an older implementation, because right now, this is expected to be a `List<String>`:

```
@ConfigurationProperties(ConfigMapConfigProperties.PREFIX)
public record ConfigMapConfigProperties(....@DefaultValue List<String> paths, ...
```

You can look in `ConfigMapPropertySourceLocator::addPropertySourcesFromPaths` also, if needed.

wind57:
@ryanjbaxter minor thing that results of a bug investigation

PR #1337, opened 2023-05-02
Drop paths support (CAN ONLY GO IN THE NEXT MAJOR RELEASE)

wind57:


ryanjbaxter:
Unfortunately not.  We just bumped minor versions from 3.0.x to 3.1.x

PR #1342, opened 2023-05-11
Drop publish reports pipeline stage

wind57:


.github/workflows/maven.yaml
```diff
@@ -300,12 +300,6 @@ jobs:
               -Dskip.build.image=true          
             fi
 
-      - name: Publish Test Report
```
wind57, line 303: this is never executed anyway, meaning it constantly fails. If it failed for so much time and no cared, may be we don't need it at all? :) 

I was working on a different thing on the pipeline and this is how I noticed this.

wind57:
@ryanjbaxter  I started working on some changes on the pipeline and noticed two stages that we do not really need...

.github/workflows/maven.yaml
ryanjbaxter: Agree nice to have but if it saves time thats a bonus

PR #1344, opened 2023-05-12
Fix 1338 main

wind57:


spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/KubernetesConfigDataLoader.java
```diff
@@ -43,13 +43,14 @@ public ConfigData load(ConfigDataLoaderContext context, KubernetesConfigDataReso
 		ConfigurableBootstrapContext bootstrapContext = context.getBootstrapContext();
 		Environment env = resource.getEnvironment();
 
-		if (bootstrapContext.isRegistered(ConfigMapPropertySourceLocator.class)) {
```
wind57, line 46: this is an interesting change right here. For config data, we need to provide the sources in the order we want, looks like `@Order` is not respected. The documentation of `ConfigData` constructor is a bit interesting:

> propertySources - the config data property sources in ascending priority order.

I assume it means we need to provide the proper order to it, but it accepts a `Collection` as input; which can be a `Set`, for example... which has no order. I would have expected a List her; but then again a `LinkedHashSet` is ordered too, tricky! 

Sequenced collections would be a great addition to these kind of APIs.

wind57: This is the only important change, everything else are just tests. I had to adjust them a little, cause in the `main` branch - we have two types of tests : bootstrap and config data. Functionally, the tests have not changed at all.

wind57:
@ryanjbaxter ready. fixes the current build + add some config data tests for the issues we did yesterday on the `2.1.x` branch

PR #1366, opened 2023-06-14
Merging pr 1354 Into 3.0.x

ryanjbaxter:


ryanjbaxter:
@wind57 thanks, that got me past that error

But now I am running into this error
`Caused by: java.io.FileNotFoundException: class path resource [org/springframework/boot/BootstrapRegistryInitializer=/.class] cannot be opened because it does not exist`.  

It is coming from a new test in the PR, but I do not see this locally and I have no idea how it is possible this class file does not exist.  IDK if its something to do with splitting the tests but it appears to be something specific in github actions.

ryanjbaxter:
I am starting to think this has to do with us branching main...investigation continues tomorrow

PR #1370, opened 2023-06-17
Fix current build failure

wind57:


.github/workflows/composites/clean-space/action.yaml
```diff
@@ -0,0 +1,20 @@
+name: clean space
```
ryanjbaxter, line 1: Can we just leverage this?
https://github.com/marketplace/actions/free-disk-space-ubuntu

ryanjbaxter (commented):
Also could we save a bunch of space by removing the docker images we saved after we load them?

ryanjbaxter:
Its possible, although I dont know for sure.

We branched last week, so `main` is now using Spring Boot 3.2 and Spring 6.1.  I would submit this against k8s 3.0.x, we will have to merge it forward to main anyways.

ryanjbaxter:
In other words the failure has nothing to do with these changes

ryanjbaxter:
Correct, we (me) will need to deal with the snakeyaml stuff.  Looks like there are some issues with the latest Boot and Spring Framework builds right now so I am going to wait for them to sort that stuff out

PR #1379, opened 2023-07-03
refactor fabric8 test

wind57:


ryanjbaxter:
So essentially we were never actually reusing the container for any integration test before?

Issue #990, opened 2022-03-15
SpEL expression filter support in Kubernetes Java Client


timtebeek:
**Describe the bug**
The [documentation lists a property](https://github.com/spring-cloud/spring-cloud-kubernetes/blob/797ead29785c2a50c9ebb188577e3a7cdea3723c/docs/src/main/asciidoc/_configprops.adoc) `spring.cloud.kubernetes.discovery.filter`:
> SpEL expression to filter services AFTER they have been retrieved from the Kubernetes API server.

This property is [declared in spring-cloud-kubernetes-commons](https://github.com/spring-cloud/spring-cloud-kubernetes/blob/b73b9528b2b894a3aa24224a45e87c983bdef3bc/spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/discovery/KubernetesDiscoveryProperties.java#L61), yet seems to be [only implemented in spring-cloud-kubernetes-fabric8-discovery](https://github.com/spring-cloud/spring-cloud-kubernetes/blob/9364fb9f4c19eef5ea74656c6d3d8db309a74ad9/spring-cloud-kubernetes-fabric8-discovery/src/main/java/org/springframework/cloud/kubernetes/fabric8/discovery/KubernetesDiscoveryClient.java#L288). The [corresponding getServices in spring-cloud-kubernetes-client-discovery](https://github.com/spring-cloud/spring-cloud-kubernetes/blob/9364fb9f4c19eef5ea74656c6d3d8db309a74ad9/spring-cloud-kubernetes-client-discovery/src/main/java/org/springframework/cloud/kubernetes/client/discovery/KubernetesInformerDiscoveryClient.java#L205) does not (yet?) filter using this filter expression.

I'm not sure what the plans are for parity between discovery implementations. If this support is not planned an addition to the documentation might be in order, to note that some properties are not supported in all implementations.

wind57:
I agree with this bug, both clients should work the same. In general, I could fix this, but when I get my hands to it, unless there will be someone else before.

dounix:
Stumbled across the same when migrating from fabric8, thanks for validating what I was seeing.

wind57:
I started work on this one [here](https://github.com/spring-cloud/spring-cloud-kubernetes/pull/1322). If accepted, it will be part of the next release...

PR #1385, opened 2023-07-18
Correct behavior of discovery after SpEL feature

EvertonSA:
I was using fabric8 but I migrated to native client,

after migrating, the app i'm working on is able to discover but it behaves differently (it's matching what it should not).

this test case is the behavior using fabric8 (I assume), I think it should have the same behavior. it's breaking now because native client seems not to have the same behavior.

can someone confirm this behavior?


EvertonSA:
> what version are you using? this has been fixed, see [here](https://github.com/spring-cloud/spring-cloud-kubernetes/issues/990). And it's not clear why you are opening a PR, instead of a bug?

i'm using <spring-cloud.version>2022.0.3</spring-cloud.version> 

I have opened a PR to contribute with the unit test and fix the possible bug. but fine I will close it.

---

PR #1395, opened 2023-08-02
Move endpoints ports to commons

wind57:


ryanjbaxter:
Just noticed this was against main, I assume it should also be in 3.0.x?  I can cherry pick it back if so

Issue #1405, opened 2023-08-08
Service Metadata differs in fabric8 and k8s client in some corner case


wind57:
As part of the `ServiceInstance` that a discovery client returns, there is a field called `metadata`, which is some meta information about the service. As part of it, we return a so called port name and port number, it really does not even matter what they mean or where they come from. 

The relevant part for this issue is that sometimes port name might not be present. In that case:

- fabric8 ignores that field
- k8s takes that field into account

For example, here is how fabric8 would represent a response:

```
Map.of("k8s_namespace", "namespace1", "type", "ClusterIP")
```

while k8s:

```
Map.of("<unset>", "8080", "k8s_namespace", "namespace1", "type", "ClusterIP")
```

Notice the `"<unset>", "8080"`. 

---

imo, k8s is not doing the correct thing, for a few reasons:

- in general, we do not care about this metadata too much. We care about the "primary port" of the service, that is returned elsewhere and is computed correctly for both clients.
- if there are multiple ports without a name, k8s will override each other, because it does:

```
Map<String, String> ports = ...

for (CoreV1EndpointPorts onePort: allOfServicesPorts) { 
     if (onePort.getName == null) {
           ports.put("<unset>", onePort.getNumber())
     }
}
```

you can see that if there are more then one port without a name, one will override the other. 

I am open to any suggestions, cause this is rather confusing for me too.

ryanjbaxter:
The k8s client behavior was changed as a result of this issue

https://github.com/spring-cloud/spring-cloud-kubernetes/issues/938

I think we should recreate that with fabric8 as well

PR #1406, opened 2023-08-08
Fix 1405

wind57:


ryanjbaxter:
I assume you meant that this fixes #1405 not #1045, right?

ryanjbaxter:
We are in the middle of a release so I will merge this once that finishes, thanks

PR #1411, opened 2023-08-16
Simplify fabric8 discovery

wind57:


ryanjbaxter (commented):
This looks good.

The one thing we need to mindful of is adding more integration tests.  The build is already taking way too long and one of my top priorities once I get back from VMware Explore is to figure out a way to start to make the build more reasonable.  I know you have done a lot of work to cut down the time on GitHub actions but our Jenkins build can't take advantage of that.  

I think we need to avoid adding anymore integration tests unless there are situations that can't be covered in unit tests.  

That said is the integration test covering something that is not already covered in a unit test?

ryanjbaxter:
Its not that I dont want the test, the test is needed, but can we run the test without starting building a container, starting k8s, deploying the container for the test and wiremock?  Can we use a regular JUnit test + wiremock to mock the k8s API calls and achieve the same thing?

PR #1414, opened 2023-08-20
simply-k8s-client-catalog-watcher-it (5)

wind57:


ryanjbaxter (commented):
Just so I understand this you are patching the existing deployments to setup the second test, and the `Order` is used to make sure the tests are run in the exact order?

PR #1420, opened 2023-08-25
Improve finding tests (1)

wind57:


spring-cloud-kubernetes-test-support/pom.xml
```diff
@@ -43,4 +49,24 @@
...
+				<artifactId>exec-maven-plugin</artifactId>
+				<version>3.1.0</version>
+				<executions>
+					<execution>
+						<goals>
+							<goal>java</goal>
+						</goals>
+					</execution>
+				</executions>
+				<configuration>
```
ryanjbaxter, line 65: @wind57 I think we need to skip this unless a specific profile is activated.  In other words we should only due this is we are running on GitHub actions.  If we don't generate the tmp file before running the build, the build will fail.

PR #1445, opened 2023-09-16
conditional ingress for wiremock pods

wind57:


spring-cloud-kubernetes-test-support/src/main/java/org/springframework/cloud/kubernetes/integration/tests/commons/fabric8_client/Util.java
```diff
@@ -260,6 +260,7 @@ private void waitForConfigMap(String namespace, ConfigMap configMap, Phase phase
 		});
 	}
 
+	@Deprecated(forRemoval = true)
```
ryanjbaxter, line 263: why not have this method call the new method?

PR #1463, opened 2023-09-29
separate surefire/failsafe directory for report on each matrix step

wind57:


pom.xml
```diff
@@ -62,15 +62,14 @@
 	</scm>
 
 	<properties>
+		<CURRENT_INSTANCE>one</CURRENT_INSTANCE>
```
ryanjbaxter, line 65: Can you add a comment here about what this is used for?

PR #1478, opened 2023-10-12
Change directory of failsafe and surefire reports when not using GitHub Actions

ryanjbaxter:


ryanjbaxter:
@wind57 can you please take a look at this?  The directory change of surefire and failsafe reports was causing issues for us on Jenkins

ryanjbaxter:
We have an assumption across all spring cloud projects that the surefire xml files are in `/surefire` and changing that assumption for just one project is more work than it is worth

PR #1530, opened 2023-12-05
Move classes to commons

wind57:


spring-cloud-kubernetes-controllers/spring-cloud-kubernetes-discoveryserver/src/main/java/org/springframework/cloud/kubernetes/discoveryserver/DiscoveryServerController.java
```diff
@@ -56,8 +58,4 @@ public Mono<ServiceInstance> appInstance(@PathVariable String name, @PathVariabl
 				.filter(serviceInstance -> serviceInstance.getInstanceId().equals(instanceId)).singleOrEmpty();
 	}
 
-	record Service(String name, List<ServiceInstance> serviceInstances) {
```
wind57, line 59: 1. the idea of this PR really starts from here. So in discovery server we have a "Service", since we talk over http with the client, at the time of the creation it meant to create a "Service" in the client also. But since then, we have refactored the code a lot, and now its possible to move this logical piece of Service into the commons package. This is what I am doing here: remove from here, move to commons

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/discovery/Service.java
```diff
@@ -0,0 +1,26 @@
...
+
+package org.springframework.cloud.kubernetes.commons.discovery;
+
+import java.util.List;
+
+/**
+ * Use for discovery service implementation.
+ * @author wind57
+ */
+public record Service(String name, List<DefaultKubernetesServiceInstance> serviceInstances) {
```
wind57, line 25: this is now in commons, that both server and client can re-use.

spring-cloud-kubernetes-controllers/spring-cloud-kubernetes-discoveryserver/src/main/java/org/springframework/cloud/kubernetes/discoveryserver/DiscoveryServerController.java
```diff
@@ -41,8 +41,10 @@ public DiscoveryServerController(KubernetesInformerReactiveDiscoveryClient react
 
 	@GetMapping("/apps")
 	public Flux<Service> apps() {
-		return reactiveDiscoveryClient.getServices().flatMap(service -> reactiveDiscoveryClient.getInstances(service)
-				.collectList().flatMap(serviceInstances -> Mono.just(new Service(service, serviceInstances))));
+		return reactiveDiscoveryClient.getServices()
```
wind57, line 44: This is a bit involved and needs a bit of wording to explain, IMO.

We used to return a `Service(String name, List<ServiceInstance> serviceInstances)`, where a `ServiceInstance` is an interface and `Service` is a package-private record. As a caller of this endpoint, you could have not used this as-is, for example by going to "/apps" because : 

- `Service` was package private, so it could have not been used in any caller's code.

- when calling this endpoint you have to provide a real `ServiceInstance` otherwise the deserialization would have failed with something along the lines of : "you need to provide an implementation of ServiceInstance, I, fasterxml, don't know in what you want me to deserialize this".

And that is exactly the reason we had (the now deprecated) `KubernetesServiceInstance` and `Service` in discovery client and where using it like this:

```
	@Override
	public List<ServiceInstance> getInstances(String serviceId) {
		KubernetesServiceInstance[] responseBody = rest
				.getForEntity(discoveryServerUrl + "/apps/" + serviceId, KubernetesServiceInstance[].class).getBody();

        ..... 
   }
```

and : 

```
	@Override
		Service[] services = rest.getForEntity(discoveryServerUrl + "/apps", Service[].class).getBody();

        ....
   }
```

---

Now if we want to (and I really want to) re-use the _same_ new `Service` record (that is now in commons package) in both the server and the client, we need to provide a real type in the `Service` record. That is why our `Service` from commons is declared as : 

```
record Service(String name, List<DefaultKubernetesServiceInstance> serviceInstances)
```

and _not_ as:

```
record Service(String name, List<ServiceInstance> serviceInstances)
```

so that the discovery client can use that as (without fasterxml complaining): 

```
	@Override
	public List<ServiceInstance> getInstances(String serviceId) {
		DefaultKubernetesServiceInstance[] responseBody = rest
				.getForEntity(discoveryServerUrl + "/apps/" + serviceId, DefaultKubernetesServiceInstance[].class)
				.getBody();

       ......
```

and 

```
	@Override
		Service[] services = rest.getForEntity(discoveryServerUrl + "/apps", Service[].class).getBody();
        .....
    }
```

---

Because of the reasons above, the code now looks like this:

```
.... serviceInstances.stream().map(x -> (DefaultKubernetesServiceInstance) x).toList()))));
```

notice the needed cast. We were already returning the `DefaultKubernetesServiceInstance` just not explicitly, under the hood it was still this type. 

So, we actually _change_ the return type now from `public Flux<Service> apps()` because that `Service` is different: it is the one from commons, not the package private record anymore. Is that a breaking change? To me no, since as a caller if you wanted to use this directly (not with discovery client), you would still have to create separate classes that would be able to parse this response. If there are classes that users created for such a use-case, their parsing would not be broken, cause the field types/names are still the same, just packaged differently.

wind57: Im sorry for the long explanation above, but I really think its important

spring-cloud-kubernetes-discovery/src/main/java/org/springframework/cloud/kubernetes/discovery/KubernetesDiscoveryClient.java
```diff
@@ -68,8 +70,9 @@ public String description() {
 
 	@Override
 	public List<ServiceInstance> getInstances(String serviceId) {
-		KubernetesServiceInstance[] responseBody = rest
-				.getForEntity(discoveryServerUrl + "/apps/" + serviceId, KubernetesServiceInstance[].class).getBody();
+		DefaultKubernetesServiceInstance[] responseBody = rest
```
wind57, line 73: we don't pase into the custom `KubernetesServiceInstance` anymore, but as all other discovery clients do, into the  `DefaultKubernetesServiceInstance`. This also makes sure that we align all 3 discovery implementations.

wind57:
@ryanjbaxter one more refactor/alignment to look at. thank you

spring-cloud-kubernetes-controllers/spring-cloud-kubernetes-discoveryserver/src/main/java/org/springframework/cloud/kubernetes/discoveryserver/DiscoveryServerController.java
ryanjbaxter: Even thought the type may be different the JSON produced by the endpoint is unchanged correct?

PR #1565, opened 2024-01-23
Use propertyresolver revert function

ryanjbaxter:
This relates to https://github.com/spring-cloud/spring-cloud-config/pull/2375

### DIFF
--- a/spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/KubernetesConfigServerBootstrapper.java
+++ b/spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/KubernetesConfigServerBootstrapper.java
@@ -16,11 +16,14 @@
 
 package org.springframework.cloud.kubernetes.commons.config;
 
+import org.springframework.boot.BootstrapContext;
 import org.springframework.boot.BootstrapRegistryInitializer;
 import org.springframework.boot.context.properties.bind.BindHandler;
 import org.springframework.boot.context.properties.bind.Bindable;
 import org.springframework.boot.context.properties.bind.Binder;
 import org.springframework.cloud.config.client.ConfigClientProperties;
+import org.springframework.cloud.config.client.ConfigServerConfigDataLocationResolver;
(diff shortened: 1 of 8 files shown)


spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/KubernetesConfigServerBootstrapper.java
```diff
@@ -41,14 +44,40 @@ public static KubernetesDiscoveryProperties createKubernetesDiscoveryProperties(
...
+				.withNamespace(
+						propertyResolver.get(KubernetesNamespaceProvider.NAMESPACE_PROPERTY, String.class, null));
+	}
+
 	public static Boolean getDiscoveryEnabled(Binder binder, BindHandler bindHandler) {
 		return binder.bind(ConfigClientProperties.CONFIG_DISCOVERY_ENABLED, Bindable.of(Boolean.class), bindHandler)
 				.orElse(false);
 	}
 
+	public static Boolean getDiscoveryEnabled(BootstrapContext bootstrapContext) {
```
pantherdd, line 72: Ah, I see there was already an `isEnabled`-type method that was `public` named `getDiscoveryEnabled` here, so it's not like you can simply choose any name ([as I assumed in the other PR](https://github.com/spring-cloud/spring-cloud-netflix/pull/4240#discussion_r1462850445)) and use that in every bootstrapper to make things consistent. It's either `getDiscoveryEnabled` everywhere, or it needs to be deprecated here and a new `is[Discovery]Enabled` created as a replacement. I'd personally prefer one of the `is*` names, but the one used here is okay too if need be. It's really up to you, I just wanted to note that now I see what you meant by breaking change.

pantherdd: I noticed one thing though: should this method not return a `boolean` primitive, as opposed to a boxed `Boolean`? Of course that would be slightly inconsistent with the other method this one overloads.

...Aah, I just realized, that's why this implementation has a `get*` method instead of `is*`, because of the boxed return type. 💡

PR #1580, opened 2024-02-19
discoveryserver: fixes deployment yaml and adds notes about use of Eureka's port

codefromthecrypt:
tried the docs, which drifted a bit.

Note: My brain broke about why this is using the same port as eureka. It seems it was an arbitrary choice, not a transition story. So, I didn't document anything about that.

### DIFF
--- a/docs/modules/ROOT/pages/spring-cloud-kubernetes-discoveryserver.adoc
+++ b/docs/modules/ROOT/pages/spring-cloud-kubernetes-discoveryserver.adoc
@@ -8,9 +8,9 @@ the `DiscoveryClient` implementation provided by that starter.
 
 ## Permissions
 The Spring Cloud Discovery server uses
-the Kubernetes API server to get data about Service and Endpoint resources so it needs list, watch, and
-get permissions to use those endpoints.  See the below sample Kubernetes deployment YAML for an
-example of how to configure the Service Account on Kubernetes.
+the Kubernetes API server to get data about Pod, Service and Endpoint resources, so it needs list,
+watch, and get permissions to use those endpoints.  See the below sample Kubernetes deployment YAML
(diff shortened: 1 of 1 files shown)


docs/modules/ROOT/pages/spring-cloud-kubernetes-discoveryserver.adoc
```diff
@@ -204,3 +204,7 @@ items:
 
 
 ----
+
+*Note* While the port is the same as Eureka's default (8761), there is no conflict despite the JSON
```
codefromthecrypt, line 208: ps wild guess this was chosen intentionally, like so custom servers can mount both endpoints and you only need to manage one discovery pod. If anyone knows the trivia why, do tell! cc @ryanjbaxter @spencergibb

ryanjbaxter: IDK why you would be using two different discovery servers

codefromthecrypt: on my side, it was a guess wondering why this is using the eureka port. I don't know the reason, so the important part was the question not my guess of the answer ;) If no one remembers, no worries.

codefromthecrypt: I know of at least one site trying to migrate from eureka to not eureka. That said, maybe the note is more trivia and calling attention to choosing the same well-known port as something else isn't helpful. I'll remove it!

codefromthecrypt: removed it!

PR #1603, opened 2024-03-17
Replaces manual workaround for Paketo with Maven profile

codefromthecrypt:
The former process is too easy to forget and doesn't error either.


ryanjbaxter:
Epic!  Thanks @codefromthecrypt

PR #1635, opened 2024-04-12
Fix 1634

wind57:


spring-cloud-kubernetes-test-support/src/main/resources/istio/istioctl-deployment.yaml
```diff
@@ -0,0 +1,21 @@
...
+    matchLabels:
+      app: istio-ctl
+  template:
+    metadata:
+      labels:
+        app: istio-ctl
+    spec:
+      containers:
+        - name: istio-ctl
+          image: istio/istioctl:1.20.5
```
codefromthecrypt, line 18: added https://github.com/spring-cloud/spring-cloud-kubernetes/pull/1637 for follow-up as unlike the other images, this is using a different istio version and also isn't pulled locally first (so it will be pulled from within k3s)

codefromthecrypt (commented):
thanks! I still can't test this with colima per the other matter, but looks like good progress!

spring-cloud-kubernetes-test-support/src/main/resources/istio/istioctl-deployment.yaml
codefromthecrypt: be careful caching, as in zipkin we had a security glitch that caused a lot of headache. We now have this in all our workflows.
```
      # Don't attempt to cache Docker. Sensitive information can be stolen
      # via forks, and login session ends up in ~/.docker. This is ok because
      # we publish DOCKER_PARENT_IMAGE to ghcr.io, hence local to the runner.
```

codefromthecrypt: Between first time we met and now, I understand why we are loading to k3s.

Basically, this:
* prevents double-downloading of well-known images (like istio), as they can use the copy from local docker
* allows you to use the image you just built which isn't yet published

So, theses are still helpful regardless of github caching

codefromthecrypt: I have been using k3s in go, and it also took me a very long time to understand all the nuance. I believe it must have been a lot more work before, as few if any must have been using it in testcontainers when you did this.

PR #1658, opened 2024-05-28
Fabric8 leader election (CAN ONLY GO IN THE NEXT MAJOR RELEASE)

wind57:


spring-cloud-kubernetes-fabric8-leader/src/main/java/org/springframework/cloud/kubernetes/fabric8/leader/election/Fabric8LeaderElectionCallbacksAutoConfiguration.java
```diff
@@ -0,0 +1,107 @@
...
+
+	@Bean
+	String holderIdentity() throws UnknownHostException {
+		String podHostName = LeaderUtils.hostName();
+		LOG.debug(() -> "using pod hostname : " + podHostName);
+		return podHostName;
+	}
+
+	@Bean
+	String podNamespace() {
```
stefvic, line 61: How about to add one more namespace inferring rule from service account path before fallback it to ENV? 
In most use cases, the pod should have it injected by default configuration. 
The Fabric8 KubernetesClient#getNamespace() already do it, see https://github.com/fabric8io/kubernetes-client/blob/main/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/Config.java#L915.
```
public static final String KUBERNETES_NAMESPACE_PATH = "/var/run/secrets/kubernetes.io/serviceaccount/namespace";
```

docs/modules/ROOT/pages/leader-election.adoc
```diff
@@ -25,3 +25,76 @@ To specify the name of the configmap used for leader election use the following
...
+
+'''
+
+There is another way you can configure leader election and it comes with native support in the fabric8 library. In the long run, this will be the default way to configure leader election, while the previous one will be dropped. You can treat this one much like the JDKs "preview" features.
+
+To be able to use it, you first need to disable the old implementation:
+
+[source]
+----
+spring.cloud.kubernetes.leader.enabled=false
```
ryanjbaxter, line 37: Could we disable this if the new one is enabled so folks dont need to set 2 properties?

docs/modules/ROOT/pages/leader-election.adoc
```diff
@@ -25,3 +25,76 @@ To specify the name of the configmap used for leader election use the following
...
+----
+
+Unlike the old implementation, this one will use either the `Lease` _or_ `ConfigMap` as the lock, depending on your cluster version. The name of that `Lease` or `ConfigMap` can be defined using the property (default value is `spring-k8s-leader-election-lock`):
+
+[source]
+----
+spring.cloud.kubernetes.leader.election.lockName=other-name
+----
+
+The namespace (`default` being set if no explicit one exists) can be set also:
```
ryanjbaxter, line 54: Shouldnt the namespace be the namespace the pod is running in?

docs/modules/ROOT/pages/leader-election.adoc
```diff
@@ -25,3 +25,76 @@ To specify the name of the configmap used for leader election use the following
...
+
+Once a certain pod establishes itself as the leader (by acquiring the lock), it will continuously (every `spring.cloud.kubernetes.leader.election.retryPeriod`) try to renew its lease, or in other words: it will try to extend its leadership. When a renewal happens, the "record" that is stored inside the lock, is updated. For example, `renewTime` is updated inside the record, to denote when the last renewal happened. (You can always peek inside these fields by using `kubectl describe lease...` for example).
+
+Renewal must happen within a certain interval, specified by `spring.cloud.kubernetes.leader.election.renewDeadline`. By default, it is equal to 10 seconds and it means that the leader pod has a maximum of 10 seconds to renew its leadership. If that does not happen, this pod is taken out from leader election process and never participates again (unless you refresh the Spring context or restart the pod).
```
ryanjbaxter, line 81: It can never participate again, meaning the pod can no longer become the leader?  Why?

docs/modules/ROOT/pages/leader-election.adoc
ryanjbaxter: But wouldn't it make sense for the default location for the lock to be created be the namespace the pods are running not 'default'?

docs/modules/ROOT/pages/leader-election.adoc
ryanjbaxter: Are you still working on this part?

ryanjbaxter: No worries at all just wanted to make sure I wasn’t missing something

ryanjbaxter:
Is this still a work in progress?

ryanjbaxter:
No worries as all I have been heads down prepping for a conference anyways, I just wanted to make sure I wasn't blocking you :)

ryanjbaxter:
3.1.x is not actually a major for us so there cannot be breaking changes in main.  However this is a big enough change it seems that we should keep it in main and guarded by a feature flag.

docs/modules/ROOT/pages/leader-election.adoc
```diff
@@ -25,3 +25,76 @@ To specify the name of the configmap used for leader election use the following
...
+----
+
+The namespace where the lock is created (`default` being set if no explicit one exists) can be set also:
+
+[source]
+----
+spring.cloud.kubernetes.leader.election.lockNamespace=other-namespace
+----
+
+Before the leader election process kicks in, you can wait until the pod is ready (via the readiness check). This is enabled by default, but you can disable it if needed:
```
ryanjbaxter, line 54: What is the point of disabling this if it won't work without it?

spring-cloud-kubernetes-fabric8-leader/src/main/java/org/springframework/cloud/kubernetes/fabric8/leader/election/Fabric8LeaderElectionInitiator.java
```diff
@@ -0,0 +1,156 @@
...
+	}
+
+	@PostConstruct
+	void postConstruct() {
+		LOG.info(() -> "starting leader initiator");
+		executorService.set(Executors.newSingleThreadExecutor());
+		CompletableFuture<Void> podReadyFuture = new CompletableFuture<>();
+
+		// wait until pod is ready
+		if (leaderElectionProperties.waitForPodReady()) {
```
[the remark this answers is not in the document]
ryanjbaxter: This seems like a useful feature.  I wonder if it makes sense for it to live in the commons package?

Honestly it would be useful to have in Fabric8 IMO

ryanjbaxter (commented):
I am a bit confused by the PR title.  If this is disabled by default I think it can go in main and then make the old implementation as deprecated.  Then we enable it by default in the major then remove the old implementation.  

Also do you plan on an equivalent implementation for the K8S Java client?

ryanjbaxter:
It should go into main IMO

ryanjbaxter:
I thought everything was wrapped in a feature flag and it wasn't introducing any breaking changes.  If that is not the case then it will need to wait and cannot go into main right now.

docs/modules/ROOT/pages/leader-election.adoc
```diff
@@ -25,3 +25,76 @@ To specify the name of the configmap used for leader election use the following
...
+
+After `podA` has become the leader, at every 2 seconds, it will try to "extend" or _renew_ its leadership. So at `12:00:02`, then at `12:00:04` and so on, `podA` goes to the lock and updates its record to reflect that it is still the leader. Between the last successful renewal and the next one, it has exactly 10 seconds (`renewalDeadline`). If it fails to renew its leadership (there is a connection problem or a big GC pause, etc.) within those 10 seconds, it stops leading and `podB` can acquire the leadership now. When `podA` stops being a leader in a graceful way, the lock record is "cleared", basically meaning that `podB` can acquire leadership immediately.
+
+A different story happens when `podA` dies with an OOM for example, without being able to gracefully update lock record and this is when `leaseDuration` argument matters. The easiest way to explain is via an example:
+
+`podA` has renewed its leadership at `12:00:04`, but at `12:00:05` it has been killed by the OOMKiller. At `12:00:06`, `podB` will try to become the leader. It will check if "now" (`12:00:06`) is _after_ last renewal + lease duration, essentially it will check:
```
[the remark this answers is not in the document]
ryanjbaxter: Thats fine, it was more of an efficiency

ryanjbaxter:
LGTM once you are satisfied with it we can merge it into main.

wind57:
@ryanjbaxter from the previous conversation that we had some time ago (where you approved the changes), all I did now is add an integration test (that is why I left this PR aside and moved to refactor integration test, so that we can reduce the time, so that I can add this test, for example). I hope this makes sense. it is now ready. thank you

ryanjbaxter:
But it still needs a major?  Also the DCO check needs to pass

wind57:
- from the previous conversation, we agreed that since this is feature flag protected (and we do not drop the previous implementation) it does not need a major
- DCO is is going to be a pain, since I've started this work a lot before we had DCO in place. If you say its a must-have, I will close this one and create a new PR, but it will take a while

spring-cloud-kubernetes-fabric8-leader/src/main/java/org/springframework/cloud/kubernetes/fabric8/leader/election/Fabric8LeaderElectionInitiator.java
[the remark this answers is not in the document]
ryanjbaxter: @wind57 👆

spring-cloud-kubernetes-fabric8-leader/src/main/java/org/springframework/cloud/kubernetes/fabric8/leader/election/Fabric8LeaderElectionAutoConfiguration.java
```diff
@@ -0,0 +1,112 @@
...
+		return new LeaderElectionConfigBuilder().withReleaseOnCancel().withName("Spring k8s leader election")
+				.withLeaseDuration(Duration.ofSeconds(properties.leaseDuration())).withLock(lock)
+				.withRenewDeadline(Duration.ofSeconds(properties.renewDeadline()))
+				.withRetryPeriod(Duration.ofSeconds(properties.retryPeriod()))
+				.withLeaderCallbacks(fabric8LeaderElectionCallbacks).build();
+	}
+
+	@Bean
+	@ConditionalOnMissingBean
+	Lock lock(KubernetesClient fabric8KubernetesClient, LeaderElectionProperties properties, String holderIdentity) {
```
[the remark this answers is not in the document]
ryanjbaxter: @wind57 👆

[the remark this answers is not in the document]
wind57: this is already implemented in `Fabric8LeaderElectionAutoConfiguration::lock`:

```
		if (leaseSupported) {
			if (properties.useConfigMapAsLock()) {
				LOG.info(() -> "leases are supported on the cluster, but config map will be used "
						+ "(because 'spring.cloud.kubernetes.leader.election.use-config-map-as-lock=true')");
				return new ConfigMapLock(properties.lockNamespace(), properties.lockName(), holderIdentity);
			}
			else {
				LOG.info(() -> "will use lease as the lock for leader election");
				return new LeaseLock(properties.lockNamespace(), properties.lockName(), holderIdentity);
			}
		}
```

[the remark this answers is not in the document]
wind57: RBAC documentation is already provided too:

```
You might have to give proper RBAC to be able to use this functionality, for example:

[source]
----
 - apiGroups: [ "coordination.k8s.io" ]
   resources: [ "leases", "configmaps" ]
   verbs: [ "get", "update", "create", "patch"]
----
```

spring-cloud-kubernetes-fabric8-leader/src/main/java/org/springframework/cloud/kubernetes/fabric8/leader/election/Fabric8LeaderElectionInitiator.java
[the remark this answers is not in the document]
wind57: I'm not sure I follow you here, you mean the functionality of the pod being ready? Just this part, correct?

spring-cloud-kubernetes-fabric8-leader/src/main/java/org/springframework/cloud/kubernetes/fabric8/leader/election/Fabric8LeaderElectionAutoConfiguration.java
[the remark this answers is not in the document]
wind57: its also documented : 

```
Unlike the old implementation, this one will use either the `Lease` _or_ `ConfigMap` as the lock, depending on your cluster version. You can force using configMap still, even if leases are supported, via :

[source]
----
spring.cloud.kubernetes.leader.election.use-config-map-as-lock=true
----
```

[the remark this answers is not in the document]
wind57: imho, this conversation can be resolved, but I'll leave it up to you. thank you

spring-cloud-kubernetes-fabric8-leader/src/main/java/org/springframework/cloud/kubernetes/fabric8/leader/election/Fabric8LeaderElectionInitiator.java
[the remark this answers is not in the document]
ryanjbaxter: Correct

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/PodReady.java
```diff
@@ -0,0 +1,78 @@
...
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.BooleanSupplier;
+
+import org.springframework.core.log.LogAccessor;
+
+/**
+ * @author wind57
+ */
+public final class PodReady {
```
wind57, line 30: @ryanjbaxter this was the comment you made to extract to the commons package and now I agree with this a lot, because I started to look into the k8s-native implementation for leader election and we will most probably need the same thing there...

spring-cloud-kubernetes-fabric8-leader/src/main/java/org/springframework/cloud/kubernetes/fabric8/leader/Fabric8LeaderAutoConfiguration.java
```diff
@@ -44,7 +44,7 @@
 @Configuration(proxyBeanMethods = false)
 @EnableConfigurationProperties(LeaderProperties.class)
 @ConditionalOnBean(KubernetesClient.class)
-@ConditionalOnProperty(value = "spring.cloud.kubernetes.leader.enabled", matchIfMissing = true)
+@ConditionalOnLeaderElectionDisabled
```
wind57, line 48: tagging you in here also @ryanjbaxter because I need an opinion. 

So before we made this change, you could disable leader election via : 

```
spring.cloud.kubernetes.leader.enabled=false
```

So you would get this dependency and for whatever reason, if you wanted to disable leader election, you could. Now, we are taking away this option from users and they are essentially left with two options only:

- use the previous implementation (via `spring.cloud.kubernetes.leader.election.enabled=false`)
- use the new implementation (via `spring.cloud.kubernetes.leader.election.enabled=true`)

There are two things out of this: 1) there is no way to opt-out of leader election now; though it is weird to have the dependency on your classpath and still disable it 2) this can easily be seen as a breaking change...

Let me know if you need more details or this is clear for you. We can't this one merged until we are both on par. Thank you

spring-cloud-kubernetes-fabric8-leader/src/main/java/org/springframework/cloud/kubernetes/fabric8/leader/election/Fabric8LeaderElectionInitiator.java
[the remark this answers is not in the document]
wind57: tagged you where we can further debate this. resolving here, so we can take a look there

spring-cloud-kubernetes-fabric8-leader/src/test/java/org/springframework/cloud/kubernetes/fabric8/leader/election/Fabric8LeaderElectionConcurrentITTest.java
```diff
@@ -0,0 +1,161 @@
...
+
+import org.springframework.boot.test.system.CapturedOutput;
+import org.springframework.boot.test.system.OutputCaptureExtension;
+import org.springframework.cloud.kubernetes.commons.leader.election.LeaderElectionProperties;
+import org.springframework.cloud.kubernetes.integration.tests.commons.Commons;
+
+/**
+ * @author wind57
+ */
+@ExtendWith(OutputCaptureExtension.class)
```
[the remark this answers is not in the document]
wind57: I wrote an integration test, too

spring-cloud-kubernetes-fabric8-leader/src/main/java/org/springframework/cloud/kubernetes/fabric8/leader/Fabric8LeaderAutoConfiguration.java
ryanjbaxter: If the leader election functionality is still present but the underlying implementation changes, I think we can just remove the old implementation.  Of course that means that this would only go in a major then.

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/PodReady.java
ryanjbaxter: Awesome!

spring-cloud-kubernetes-fabric8-leader/src/main/java/org/springframework/cloud/kubernetes/fabric8/leader/Fabric8LeaderAutoConfiguration.java
wind57: I actually wanted to keep both implementations :

- the old stays intact and works the same as it used to 
- you can opt-in into the new one

The reason for this is that this I would like to users to be able to "test" the new functionality and raise any issues/concerns they have, but at the same time have a safe fallback. 

If you agree, I will change the flags a bit to support this.

ryanjbaxter: OK and it needs to be documented

wind57: its already documented:

```
There is another way you can configure leader election, and it comes with native support in the fabric8 library (k8s native client support is not yet implemented). In the long run, this will be the default way to configure leader election, while the previous one will be dropped. You can treat this one much like the JDKs "preview" features.

To be able to use it, you need to set the property:

[source]
----
spring.cloud.kubernetes.leader.election.enabled=true
----
```

I've added tests and left the old implementation intact. 

This is finally ready to go imo.

PR #1677, opened 2024-07-23
Fix build failure source plugin

wind57:


ryanjbaxter:
I think actually this dependency upgrade broke things
https://github.com/spring-cloud/spring-cloud-build/pull/277

I am not sure why, but I reverted the train and so far the build looks pretty good.

ryanjbaxter:
Thanks for the investigation here.  I agree that eventually we want to upgrade, but there isn't much harm in staying where we are for now and IMO is preferable to making the changes to work around the issue in this PR.

I think it is worth opening an issue with Spring Boot with the details above and see if they have some thoughts on this.  My guess is that we are not the only ones running into this and maybe they have some suggestions.

ryanjbaxter:
I merged the change in spring cloud build, once the build finishes and the snapshots are available you can test things out.  It should only take a few minutes for the build to finish

PR #1716, opened 2024-09-10
Inline application.yaml/yml/properties into constants 

wind57:


ryanjbaxter:
Should this also be done in 3.1.x?

PR #1768, opened 2024-10-16
Cleanup in it utilities 4

wind57:


ryanjbaxter:
Why do we no longer need to call `cleanup`?

ryanjbaxter:
Is it saving a lot of time?

PR #1773, opened 2024-10-21
Close fabric8 client

wind57:

PR #1780, opened 2024-10-22
Close Fabric8 Client When Context Is Stopped

ryanjbaxter:


ryanjbaxter:
@wind57 thanks.  I am still not really happy with this.  Since `KubernetesClient` implements `AutoClosable` it seems like the `close` method should be called automatically.  Maybe the timing in which that happens is too late 🤷‍♂️.

wind57:
@ryanjbaxter do you remember why we did this? I was trying to find a relevant issue, but no luck : https://github.com/spring-cloud/spring-cloud-kubernetes/issues?q=is%3Aissue%20state%3Aclosed%20created%3A2024-07-01..2024-10-22 

The thing is, this fix causes a problem in a different place, in the leader election. a) we have an issue for it b) I can re-produce it. So I'm trying to see what was the underlying problem that triggered this

ryanjbaxter:
@wind57 I think it stemmed from this https://github.com/spring-cloud/spring-cloud-kubernetes/pull/1773/

PR #1805, opened 2024-11-28
Update Kubernetes client deps

desiderantes:
Avoids inconsistency with Spring Boot 3.4.0 which requires gson 2.11.x due to the `com.google.gson.Strictness`

### DIFF
--- a/spring-cloud-kubernetes-dependencies/pom.xml
+++ b/spring-cloud-kubernetes-dependencies/pom.xml
@@ -33,8 +33,8 @@
 	<description>Spring Cloud Kubernetes Dependencies</description>
 	<properties>
 		<kubernetes-fabric8-client.version>6.13.4</kubernetes-fabric8-client.version>
-		<kubernetes-native-client.version>19.0.1</kubernetes-native-client.version>
-		<wiremock.version>3.4.2</wiremock.version>
+		<kubernetes-native-client.version>22.0.0</kubernetes-native-client.version>
+		<wiremock.version>3.9.2</wiremock.version>
 	</properties>
 	<dependencyManagement>
 		<dependencies>


ryanjbaxter:
Are you getting some kind of runtime exception using Boot 3.4.p?

desiderantes:
> Are you getting some kind of runtime exception using Boot 3.4.0?

Yes, kubernetes client java is pulling Gson 2.10.1, and it's downgrading it from 2.11, which is required on spring boot 3.4.0 because now the gson configuration class, GsonProperties, uses the Strictness class from latest gson. This, for example, breaks spring-boot-admin when used with kubernetes discovery. Of course, I can just force the gson version via the build system, but it's doing the wrong thing by default.

xcq1:
I stumbled on this while encountering https://github.com/kubernetes-client/java/issues/1659 and trying to use a fixed version in the build pipeline: It appears that due to API incompatibility there currently isn't a version of kubernetes-client with that bugfix that can be used with spring-cloud-kubernetes.

xcq1:
@ryanjbaxter Upgrading the kubernetes-client version like suggested in this PR would resolve that problem as well. If you prefer, I can open a detailed issue.

xcq1:
Sorry, I am unfamiliar with Spring cloud's exact release cycles, I just wanted to point out the general need. I assume the next major version is a longer time away, in that case a backport of the fix would be well appreciated.

ryanjbaxter:
@xcq1 Can you open a separate issue for this and include the fix in the Kubernetes Java Client that needs to be backported?  I can submit a PR with that fix to the correct branch and hopefully get a release from them that we can then use in our next release of Spring Cloud Kubernetes.

xcq1:
So while trying to make a minimal sample for my problem, I discovered I can completely remove any manual registration call to ModelMapper which makes the error go away and everything still works. With that I don't see a need for an additional backport. Sorry for the noise, it was just one of those days. Thanks anyway.

ryanjbaxter:
@desiderantes @xcq1 the issues you have identified with the Kubernetes Java client should he addressed in the 19.0.2 release.  I have upgraded the dependency version in Spring Cloud Kuberentes and our next release of 2023.0.x and 2024.0.x should include this latest release.

PR #1815, opened 2024-12-05
Fix failing tests change logic

wind57:


ryanjbaxter:
Looks like this worked, thanks so much!

PR #1833, opened 2025-01-14
drop ingress from fabric8 integration tests

wind57:


ryanjbaxter:
@wind57 unfortunately this broke our Jenkins build here is the log from one of the failing tests
[test-failure-logs.txt](https://github.com/user-attachments/files/18440410/test-failure-logs.txt)

PR #1889, opened 2025-03-09
drop repackage from integration tests root pom

wind57:


ryanjbaxter:
Did you get an understanding of why this was here in the first place?

PR #1892, opened 2025-03-11
Use application-profile as source for configuration

josef-pones:


ryanjbaxter:
Can you elaborate on why you are making this change?  You also need to sign your commit to pass the DCO check.

wind57:
@ryanjbaxter do you think we should close this one also? It has been quite a while

PR #1893, opened 2025-03-20
Fix the doc with correct links

arjavdongaonkar:


arjavdongaonkar:
@ryanjbaxter Whats the stable branch now?
is it 3.1.x or 3.2.x ? (as 3.2.1 is latest release)

wind57:
I make PRs against `3.1.x`, by default, so I think you are good here

PR #1901, opened 2025-03-27
Replace junit assertions

wind57:


wind57:
@ryanjbaxter I dont know how you are looking at this, but I replaced junit assertions with the ones from asserj, so that we have a single way to do the tests...

ryanjbaxter:
Huh...I thought we had a checkstyle rule for this 🧐

Either way thanks!

PR #1902, opened 2025-03-28
Refactor k8s client reload

wind57:


spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/SecretsPropertySourceLocator.java
```diff
@@ -123,12 +124,16 @@ protected abstract SecretsPropertySource getPropertySource(ConfigurableEnvironme
 
 	protected void putPathConfig(CompositePropertySource composite) {
 
-		if (!properties.paths().isEmpty()) {
+		Set<String> uniquePaths = new LinkedHashSet<>(properties.paths());
```
wind57, line 127: we do this for configmaps, but not for secrets, so I decided to add it here

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/reload/ConfigReloadUtil.java
```diff
@@ -120,6 +121,10 @@ else if (propertySource instanceof MountConfigMapPropertySource mountConfigMapPr
 					// we know that the type is correct here
 					managedSources.add((S) mountConfigMapPropertySource);
 				}
+				else if (propertySource instanceof SecretsPropertySource secretsPropertySource) {
```
wind57, line 124: I am fixing a bug here with this one. 

In case of configmaps, when we create a property source from paths, we create an instance of `MountConfigMapPropertySource`. Then, in `ConfigReloadUtil`, there is such a check:

```
if (source instanceof MountConfigMapPropertySource mountConfigMapPropertySource) { ... }
```

which means that reloads will take it into consideration.

---

While refactoring one integration test, I saw that it is failing, and after some debugging, realized that we have no code like the above for secrets. As such, I added that simple condition, because when we create a property source from paths in case of secrets, we create an instance of `SecretsPropertySource `

spring-cloud-kubernetes-test-support/src/main/java/org/springframework/cloud/kubernetes/integration/tests/commons/native_client/Util.java
```diff
@@ -436,10 +438,38 @@ public void deleteClusterWide(String serviceAccountNamespace, Set<String> namesp
 	}
 
 	public void deleteNamespace(String name) {
+
+		// sometimes we get errors like :
+
+		// "message": "Discovery failed for some groups,
```
wind57, line 444: we have already this logic in fabric8 client, so I added it in here too

wind57:
@ryanjbaxter this was supposed to be just one more IT refactor, but it turned out a bit more complicated. I added comments to the relevant parts that I need your eyes on, the rest is refactoring on an existing integration test. thank you

ryanjbaxter:
Can we separate the bug fix out into its own PR?

wind57:
I thought about it too, but it get kind of ugly. If you say its a must do, I will, of course.

ryanjbaxter:
Why is it ugly?  Would it help to merge the fix before the refactor (or the other way around)?

wind57:
hmm, you're probably right. Let me try that

PR #1906, opened 2025-03-29
fix SecretsPropertySource reload bug

wind57:


wind57:
@ryanjbaxter the other 1/2 of a previous ticket, that fixes a bug...

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/reload/ConfigReloadUtil.java
```diff
@@ -110,6 +111,10 @@ else if (source instanceof MountConfigMapPropertySource mountConfigMapPropertySo
 				// we know that the type is correct here
 				managedSources.add((S) mountConfigMapPropertySource);
 			}
+			else if (source instanceof SecretsPropertySource secretsPropertySource) {
```
wind57, line 114: when we were mounting secrets, reload was not working, since there was no this check in place. Even it I'm fixing it here, it will eventually be dropped, since we will move from mounting to config import.

Issue #1757, opened 2024-10-07
`use-name-as-prefix` does not work as expected


jnodorp-jaconi:
**Describe the bug**

When importing Kubernetes secrets with `use-name-as-prefix`, I would expect the prefix to be the name of the secret. This does not work when using a label selector with duplicate keys:

`application.yaml`
```yaml
spring:
  config:
    import: "kubernetes:"

  cloud:
    kubernetes:
      secrets:
        enable-api: true
        sources:
          - labels:
              load: "true"
        use-name-as-prefix: true
```

`k8s-secrets.yaml`
```yaml
---
apiVersion: v1
kind: Secret
metadata:
  name: client-1
  labels:
    load: "true"
stringData:
  client-id: client-1
  client-secret: S3cr371!
---
apiVersion: v1
kind: Secret
metadata:
  name: client-2
  labels:
    load: "true"
stringData:
  client-id: client-2
  client-secret: S3cr372!
```

will result in these properties:

```
client-1.client-2.client-secret: S3cr372!
client-1.client-2.client-id: client-2
```

Expected result:

```
client-1.client-secret: S3cr371!
client-1.client-id: client-1
client-2.client-secret: S3cr372!
client-2.client-id: client-2
```

Spring Cloud: 2023.0.3

**Sample**

I built an example application here: https://github.com/jnodorp-jaconi/spring-cloud-kubernetes-reproducer

jekkel:
Mmh, seems this is [expected behavior](https://docs.spring.io/spring-cloud-kubernetes/reference/property-source-config/configmap-propertysource.html):


> One more thing to bear in mind is that we support prefix per source, not per secret. The easiest way to explain this is via an example:
> 
> [...]
> 
> Suppose that a query matching such a label will provide two secrets as a result: secret-a and secret-b. Both of these secrets have the same property name: color=sea-blue and color=ocean-blue. It is undefined which color will end-up as part of property sources, but the prefix for it will be secret-a.secret-b (concatenated sorted naturally, names of the secrets).

But IMO the behavior OP describes is makes more sense to me, too.

ryanjbaxter:
@wind57 why did we originally decide on the existing behavior?

ryanjbaxter:
I also agree with the OP

ryanjbaxter:
Well it doesn't seem like we intended this to work this way right?  So in my mind it is a bug.

And if by breaking compatibility you mean that we would change the property name then yes.

[...]

PR #1911, opened 2025-03-31
Fix `use-name-as-prefix`

jnodorp-jaconi:
Fixes gh-1757

### DIFF
--- a/spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/PrefixContext.java
+++ b/spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/PrefixContext.java
@@ -16,8 +16,7 @@
 
 package org.springframework.cloud.kubernetes.commons.config;
 
-import java.util.Map;
-import java.util.Set;
+import java.util.function.Function;
 
 /**
  * A holder for data needed to compute prefix based properties, in case of a secret or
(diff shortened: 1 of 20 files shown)


spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/PrefixContext.java
```diff
@@ -16,15 +16,16 @@
...
 
 /**
  * A holder for data needed to compute prefix based properties, in case of a secret or
  * config map.
  *
  * @author wind57
  */
-public final record PrefixContext(Map<String, Object> data, String prefix, String namespace,
-		Set<String> propertySourceNames) {
+public final record PrefixContext(MultipleSourcesContainer data, Function<String, String> prefix, String namespace) {
```
wind57, line 27: this is a breaking change and unfortunately we can't merge breaking changes in `main`. This only happens on "next major" releases. afaik: next one is at the end of 2025. Ryan can confirm (I'm not a vmware employee), but it seems this would be a NOGO, for the time being at least. What I do is maintain some PRs that are targeted towards next major, but this is not easy, to say the least...

wind57: I have not looked at anything else in the code tbh, will wait for Ryan's input

ryanjbaxter: We had discussed in the original issue that there would be breaking changes and it would need to [wait for a major](https://github.com/spring-cloud/spring-cloud-kubernetes/issues/1757#issuecomment-2416830110).

ryanjbaxter:
We would also need to update this documentation:
```
Suppose that a query matching such a label will provide two secrets as a result: secret-a and secret-b. Both of these secrets have the same property name: color=sea-blue and color=ocean-blue. It is undefined which color will end-up as part of property sources, but the prefix for it will be secret-a.secret-b (concatenated sorted naturally, names of the secrets).
```

wind57:
can this be closed btw?

PR #1914, opened 2025-04-08
Fix 1757

wind57:


spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/ConfigUtils.java
```diff
@@ -390,7 +399,7 @@ public static final class Prefix {
 		/**
 		 * prefix is known at the callsite.
 		 */
-		public static Prefix KNOWN;
+		public static Prefix KNOWN = new Prefix(() -> "", "KNOWN");
```
wind57, line 402: this is a minor bug fix, that someone was not found until now, not really related to the fix itself

wind57:
btw, there is no documentation for this PR, which I wanted to make a PR immediately after this one

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/LabeledSourceData.java
```diff
@@ -50,43 +58,49 @@ public final SourceData compute(Map<String, String> labels, ConfigUtils.Prefix p
...
 
-		String names = data.names().stream().sorted().collect(Collectors.joining(PROPERTY_SOURCE_NAME_SEPARATOR));
-		return new SourceData(ConfigUtils.sourceName(target, names, namespace), data.data());
+	}
 
+	/*
+	 * When there is no data, the name of the property source is made from provided
+	 * labels, unlike when the data is present: when we use secret names.
+	 */
+	private SourceData emptySourceData(Map<String, String> labels, String target, String namespace) {
```
ryanjbaxter, line 96: Wouldn't the secret still have a name in this case?

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/NamedSourceData.java
```diff
@@ -37,47 +42,68 @@ public abstract class NamedSourceData {
...
+				return new SourceData(sourceDataName,
+						defaultFlattenedSourceData(sourceNamesFound, sourceDataForSourceName));
 			}
 
+			if (prefix.getName().equals(Prefix.KNOWN.getName())) {
+				return new SourceData(sourceDataName, prefixFlattenedSourceData(sourceNamesFound,
+						sourceDataForSourceName, prefix.prefixProvider().get()));
+			}
+
+			if (prefix.getName().equals(Prefix.DELAYED.getName())) {
```
ryanjbaxter, line 85: What is a delayed prefix?

wind57: we had this for a while, I am not introducing it. 

It's for the cases when you enabled "use-name-as-prefix", but the search happens based on labels, so the actual prefix is "delayed". Meaning we do not know it upfront, we will later, when we actually read by labels and find those names

ryanjbaxter: ok i just needed a refresher because its been a while

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/LabeledSourceData.java
ryanjbaxter: But did the name of the empty property source also have the same problem as non-empty property sources?

jnodorp-jaconi (approved):
I reviewed this PR with respect to wether it fixes the issue we face (#1757), which it does.

wind57:
thank you for looking at it @jnodorp-jaconi

wind57:
I am aware that PR failed, I am looking into why and update

wind57:
@ryanjbaxter all PRs failed (`3.1.x` , `3.2.x` and `main`) because of some github internal issue. You might want to re-trigger builds

PR #1922, opened 2025-05-22
Pin version of Jackson since Fabric8 6.x is incompatible with Jackson 2.19.x

ryanjbaxter:


spencergibb (approved):
Glad it is only in tests

PR #1942, opened 2025-07-01
clean-up a bit

wind57:


.github/workflows/composites/run-and-save-test-times-when-cache-missing/action.yaml
```diff
@@ -46,7 +46,9 @@ runs:
             -Dhttp.keepAlive=false \
             -Dmaven.wagon.http.pool=false \
             -Dmaven.wagon.http.retryHandler.class=standard \
+            -Dmaven.wagon.httpconnectionManager.ttlSeconds=120 \
```
wind57, line 49: I have observed that we get some timeouts from time to time when we need something from maven central, seems that adding these options fixes the problem.

.settings.xml
```diff
@@ -46,6 +46,10 @@
 				</repository>
 			</repositories>
 			<pluginRepositories>
+				<pluginRepository>
```
wind57, line 49: I am not sure why, but if I put this one here explicitly, it will not go to spring repositories for download, but use maven central

spring-cloud-kubernetes-examples/kubernetes-leader-election-example/pom.xml
```diff
@@ -73,14 +72,6 @@
 					<skip>true</skip>
 				</configuration>
 			</plugin>
-			<plugin>
-				<artifactId>maven-surefire-plugin</artifactId>
```
wind57, line 77: remove explicit versions and use the inherited ones

spring-cloud-kubernetes-examples/pom.xml
```diff
@@ -26,6 +26,12 @@
 		<version>3.1.7-SNAPSHOT</version>
 	</parent>
 
+	<!-- override the ones coming from spring-cloud-build, until it is fixed other there -->
```
wind57: until (and if you agree) with this one : https://github.com/spring-cloud/spring-cloud-build/pull/487, I am proposing to override the versions here

.github/workflows/composites/clean-space/action.yaml
```diff
@@ -4,6 +4,12 @@ runs:
   using: "composite"
   steps:
 
+    - name: Disable man-db to make package install and removal faster
```
wind57, line 7: about 1m improvement here because of this

spring-cloud-kubernetes-commons/pom.xml
```diff
@@ -11,18 +11,6 @@
 
 	<artifactId>spring-cloud-kubernetes-commons</artifactId>
 	<name>${project.artifactId}</name>
-    <build>
-        <plugins>
-            <plugin>
```
wind57, line 16: drop explicit java version

wind57:
@ryanjbaxter some clean-up here and there, thank you.  Ready to be looked at

spring-cloud-kubernetes-examples/pom.xml
ryanjbaxter: This is merged now, can you remove this?

wind57: you're right, removed.

PR #1962, opened 2025-07-29
fix issue 1953

wind57:


spring-cloud-kubernetes-fabric8-leader/src/main/java/org/springframework/cloud/kubernetes/fabric8/leader/Fabric8LeaderAutoConfiguration.java
```diff
@@ -101,7 +101,7 @@ public Fabric8LeaderRecordWatcher leaderRecordWatcher(LeaderProperties leaderPro
 		return new Fabric8LeaderRecordWatcher(leaderProperties, fabric8LeadershipController, kubernetesClient);
 	}
 
-	@Bean(destroyMethod = "stop")
+	@Bean
```
wind57, line 104: `LeaderInitiator` is declared like this:

```
LeaderInitiator implements SmartLifecycle
```

and already has : 


```
	@Override
	public void stop() {
		if (isRunning()) {
			LOGGER.debug(() -> "Leader initiator stopping");
			scheduledExecutorService.shutdown();
			scheduledExecutorService = null;
			hostPodWatcher.stop();
			leaderRecordWatcher.stop();
			leadershipController.revoke();
			isRunning = false;
		}
	}

	@Override
	public void stop(Runnable callback) {
		stop();
		callback.run();
	}
```

The fact that we had its declaration as : `@Bean(destroyMethod = "stop")` would mean that the `stop` method is called twice, cause the issue

PR #1964, opened 2025-07-31
fix 1963: order for discovery clients

wind57:


spring-cloud-kubernetes-client-discovery/src/main/java/org/springframework/cloud/kubernetes/client/discovery/reactive/KubernetesInformerReactiveDiscoveryClient.java
```diff
@@ -73,4 +73,9 @@ public Flux<String> getServices() {
 			.subscribeOn(Schedulers.boundedElastic());
 	}
 
+	@Override
```
wind57, line 76: since we use the same properties for the reactive and blocking reactive clients + our reactive clients are based on the blocking ones, we can re-use the order here

spring-cloud-kubernetes-client-discovery/src/test/java/org/springframework/cloud/kubernetes/client/discovery/reactive/KubernetesInformerReactiveDiscoveryClientTests.java
```diff
@@ -338,7 +352,7 @@ private Lister<V1Service> setupServiceLister(String namespace, V1Service... serv
 		return serviceLister;
 	}
 
-	private Lister<V1Endpoints> setupEndpointsLister(String namespace, V1Endpoints... endpoints) {
+	private Lister<V1Endpoints> setupEndpointsLister(V1Endpoints... endpoints) {
```
wind57, line 355: minor simplification that is un-related to the issue itself

wind57:
@ryanjbaxter one more fix ready here. thank you

PR #1975, opened 2025-08-12
fix issue 1968

wind57:


spring-cloud-kubernetes-fabric8-autoconfig/src/main/java/org/springframework/cloud/kubernetes/fabric8/Fabric8AutoConfiguration.java
```diff
@@ -121,13 +117,4 @@ public Fabric8PodUtils kubernetesPodUtils(KubernetesClient client) {
 		return new Fabric8PodUtils(client);
 	}
 
-	@EventListener
```
wind57, line 124: so after reviewing the history a bit, we only did this because of a failure in an integration test.

Since then, we refactored integration tests a lot and cleaned up the code. I see no errors now in this build, it passes.

Also, to be honest, I would drop this code (because it impacts leader election) and review again, if we ever have the same problem.

wind57:
@ryanjbaxter this is ready to be looked at, thank you

PR #1977, opened 2025-08-18
do not use Flux::defer when there is no need

wind57:


spring-cloud-kubernetes-client-loadbalancer/src/main/java/org/springframework/cloud/kubernetes/client/loadbalancer/KubernetesClientServicesListSupplier.java
```diff
@@ -82,7 +82,7 @@ else if (!discoveryProperties.namespaces().isEmpty()) {
 		}
 
 		LOG.debug(() -> "found services : " + result);
-		return Flux.defer(() -> Flux.just(result));
+		return Flux.just(result);
```
wind57, line 85: since `result` is already a computed `List<ServiceInstance>`, there is no need to defer

spring-cloud-kubernetes-fabric8-loadbalancer/src/main/java/org/springframework/cloud/kubernetes/fabric8/loadbalancer/Fabric8ServicesListSupplier.java
```diff
@@ -101,7 +101,7 @@ else if (!discoveryProperties.namespaces().isEmpty()) {
 		}
 
 		LOG.debug(() -> "found services : " + result);
-		return Flux.defer(() -> Flux.just(result));
+		return Flux.just(result);
```
wind57, line 104: same here, no need to defer

wind57:
@ryanjbaxter minor thing over here. thank you

PR #1992, opened 2025-08-26
fix issue 1970

wind57:


spring-cloud-kubernetes-client-loadbalancer/src/main/java/org/springframework/cloud/kubernetes/client/loadbalancer/KubernetesClientServicesListSupplier.java
```diff
@@ -57,32 +57,34 @@ public KubernetesClientServicesListSupplier(Environment environment,
 
 	@Override
 	public Flux<List<ServiceInstance>> get() {
-		List<ServiceInstance> result = new ArrayList<>();
```
wind57, line 60: we had our code as : 

```
public Flux<List<ServiceInstance>> get() { 

      List<ServiceInstance> result = ....
      return Flux.just(result); 

}
```

and it changes to : 

```
public Flux<List<ServiceInstance>> get() { 

      return Flux.defer(() -> {
                 List<ServiceInstance> result = ....
                 return Flux.just(result); 
      })
}
```

That is the actual fix.

spring-cloud-kubernetes-fabric8-loadbalancer/src/main/java/org/springframework/cloud/kubernetes/fabric8/loadbalancer/Fabric8ServicesListSupplier.java
```diff
@@ -57,51 +57,53 @@ public class Fabric8ServicesListSupplier extends KubernetesServicesListSupplier<
 
 	@Override
 	public Flux<List<ServiceInstance>> get() {
-		List<ServiceInstance> result = new ArrayList<>();
-		String serviceName = getServiceId();
-		LOG.debug(() -> "serviceID : " + serviceName);
+		return Flux.defer(() -> {
```
wind57, line 60: same fix as above, nothing changes besides that

spring-cloud-kubernetes-fabric8-loadbalancer/src/test/java/org/springframework/cloud/kubernetes/fabric8/loadbalancer/it/Util.java
```diff
@@ -63,24 +56,4 @@ public static Endpoints endpoints(int port, String host, String namespace) {
 			.build();
 	}
 
-	@TestConfiguration
-	public static class LoadBalancerConfiguration {
```
wind57, line 67: not removed, but moved to individual classes

wind57:
@ryanjbaxter since the `main` is fine now, can we start looking into some fixes may be? This one is pretty trivial imho, the actual fix is in two files only (where I left comments). everything else is just tests.

PR #1997, opened 2025-08-29
drop methods

wind57:


wind57:
@ryanjbaxter I want to start very simple. This one removes some `public` methods that we do not use anymore

PR #2010, opened 2025-09-22
Simplify config reload (ONLY FOR THE NEXT MAJOR)

wind57:


spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/reload/ConfigurationChangeDetector.java
```diff
@@ -33,16 +32,9 @@ public abstract class ConfigurationChangeDetector {
 
 	private static final LogAccessor LOG = new LogAccessor(LogFactory.getLog(ConfigurationChangeDetector.class));
 
-	protected ConfigurableEnvironment environment;
```
wind57, line 36: the idea of the PR starts here : we do not need these in this class...

PR #2015, opened 2025-09-25
fix issue 2008 (3.2.x)

wind57:


spring-cloud-kubernetes-fabric8-config/src/test/java/org/springframework/cloud/kubernetes/fabric8/config/reload_it/PollingReloadConfigMapAndSecretTest.java
```diff
@@ -0,0 +1,198 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Proves that
+ * <a href="https://github.com/spring-cloud/spring-cloud-kubernetes/issues/2008">this</a>
+ * issue is fixed.
+ *
+ * @author wind57
+ */
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT,
```
wind57, line 65: this is a test that uses the exact scenario the OP was presenting in the issue and shows that this is now fixed

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/MountSecretPropertySource.java
```diff
@@ -0,0 +1,28 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.kubernetes.commons.config;
+
+/**
+ * @author wind57
+ */
+public final class MountSecretPropertySource extends SecretsPropertySource {
```
wind57, line 22: it was long due for us to have this. We have `MountConfigMapPropertySource`, but not one for secret. This has caused various issues over time and confusion. Introducing it should make things far easier imo

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/SecretsPropertySourceLocator.java
```diff
@@ -152,33 +152,33 @@ protected void putPathConfig(CompositePropertySource composite) {
 	 * @author wind57
 	 */
 	private static class SecretsPropertySourceCollector
-			implements Collector<Path, List<SecretsPropertySource>, List<SecretsPropertySource>> {
+			implements Collector<Path, List<MountSecretPropertySource>, List<MountSecretPropertySource>> {
```
wind57, line 155: the only change here is : `SecretsPropertySource  -> MountSecretPropertySource `

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/reload/ConfigReloadUtil.java
```diff
@@ -111,9 +112,9 @@ else if (source instanceof MountConfigMapPropertySource mountConfigMapPropertySo
 				// we know that the type is correct here
 				managedSources.add((S) mountConfigMapPropertySource);
 			}
-			else if (source instanceof SecretsPropertySource secretsPropertySource) {
+			else if (source instanceof MountSecretPropertySource mountSecretPropertySource) {
```
wind57, line 115: we now check for `MountSecretPropertySource`, exactly like we do in case of configmap:

```
				else if (propertySource instanceof MountConfigMapPropertySource mountConfigMapPropertySource) {
				}
				else if (propertySource instanceof MountSecretPropertySource mountSecretPropertySource) {
					managedSources.add((S) mountSecretPropertySource);
				}
```

wind57:
@ryanjbaxter this is ready also. I know you said those breaking changes are a priority, but this one needs to be in main also, because later it will have many changes, both removal and insertions and this way it will be easier to handle.

PR #2016, opened 2025-09-25
lower visibility of method ( main )

wind57:


spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/reload/ConfigReloadUtil.java
```diff
@@ -88,8 +88,7 @@ public static boolean reload(PropertySourceLocator locator, ConfigurableEnvironm
 	 * @return finds all registered property sources of the given type
 	 * @deprecated this method will not be public in the next major release.
 	 */
-	@Deprecated(forRemoval = false)
```
wind57, line 91: lowers the visibility of this method: it is needed because reasoning about the code becomes far easier here.

spring-cloud-kubernetes-fabric8-config/src/test/java/org/springframework/cloud/kubernetes/fabric8/config/EventBasedConfigurationChangeDetectorTests.java
```diff
@@ -1,80 +0,0 @@
-/*
- * Copyright 2013-present the original author or authors.
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
```
wind57, line 5: drop this test that relies on the now non-public above method. tbh, we already cover all these cases in other places anyway

wind57:
@ryanjbaxter ready

PR #2019, opened 2025-09-25
clean-up in main after 1757 ( main )

wind57:


wind57:
The whole idea of this PR, is that a while ago, while fixing an issue, we abused a `Map` where the value was `Object`:

```
LinkedHashSet<String> names, Map<String, Object> rawData
```

where `rawData` value is an `Object`, but inside it we had another `Map`. 

We had to do this, otherwise there would be breaking changes.

---

The current PR changes that to : 

```
LinkedHashMap<String, Map<String, Object>> sourceData
```

without weird casts and abuses of that `Object` now.


It may be vague as an explanation, but its the idea here. If needed, I will go into a lot more details, just let me know

wind57:
@ryanjbaxter this one is ready also

PR #2023, opened 2025-09-26
replace @MockBean and @SpyBean

wind57:


spring-cloud-kubernetes-fabric8-config/src/test/java/org/springframework/cloud/kubernetes/fabric8/config/locator_retry/fail_fast_enabled_retry_disabled/ConfigFailFastEnabledButRetryDisabled.java
```diff
@@ -89,4 +96,15 @@ void locateShouldFailWithoutRetrying() {
 		verify(propertySourceLocator, times(1)).locate(any());
 	}
 
+	@TestConfiguration
```
wind57, line 97: this is the only place where a drop-in replacement for the deprecated annotation is not possible. It is rather interesting why. 

`@SpyBean` creates a bean if such is not present already, while `@MockitoSyBean` expects one to be present already. 

In this case, we don't have a bean of type `Fabric8ConfigMapPropertySourceLocator`, so we need to provide one. But its not visible, so we provide an extension of it

wind57:
@ryanjbaxter this one is ready also.

PR #2024, opened 2025-09-26
revert workaround

wind57:


wind57:
@ryanjbaxter this is also ready. we used to have a work-around for a fabric8 issue, but now that you upgraded, we can get rid of it.

PR #2025, opened 2025-09-26
simplify arguments

wind57:


spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/reload/ConfigurationChangeDetector.java
```diff
@@ -33,16 +32,9 @@ public abstract class ConfigurationChangeDetector {
 
 	private static final LogAccessor LOG = new LogAccessor(LogFactory.getLog(ConfigurationChangeDetector.class));
 
-	protected ConfigurableEnvironment environment;
```
wind57, line 36: the idea of this PR is here. we do not really need these two arguments in this class, so I removed them. Everything else is just an artifact of this change

wind57:
@ryanjbaxter ready. Thank you

PR #2027, opened 2025-09-28
Fix 1715 drop profiles support for labeled sources

wind57:


wind57:
- what are we doing here?

We are removing the support for labeled sources coupled with profile based sources

- What is the current way of doing things?

Currently, when we read some sources by labels, we would get their names. For example:

```
apiVersion: v1
kind: ConfigMap
metadata:
  labels:
    color: red
  name: paint
```

when we search by labels : `color: red`, we will find and read this configmap.

But, if `includeProfileSpecificSources` exists, we would read all _other_ configmaps matching this profile. For example, if active profiles : `[dev, k8s]`, then we would also try to read:

```
paint-dev
paint-k8s
```

- Why is this a problem and we have to drop such support?

This has been talked [here](https://github.com/spring-cloud/spring-cloud-kubernetes/pull/1720) 


@ryanjbaxter this is ready also, and unlike the previous ones, is more substantial.

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/ConfigUtils.java
```diff
@@ -279,32 +276,9 @@ public static MultipleSourcesContainer processLabeledData(List<StrippedSourceCon
 			return labelsToSearchAgainst.entrySet().containsAll((labels.entrySet()));
 		}).toList();
 
-		// Compute profile-based source names (based on the ones we found by labels)
```
wind57, line 282: this is the actual fix: we are no longer taking into consideration `includeProfileSpecificSources` for labeled searches. 

Everything else in this PR is changes around this sole piece of code

PR #2039, opened 2025-10-05
Fix 1681 (1)

wind57:


spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/ReadType.java
```diff
@@ -19,23 +19,15 @@
 /**
  * @author wind57
  */
-public interface SecretsCache {
+public enum ReadType {
```
wind57, line 22: the idea of this PR is here and this is what the OP has described in the issue. 

At the moment, we read configmaps and secrets in a batch fashion, meaning all in a namespace. For some users this means reading a lot of sources and its not OK. So we introduce the possibility to read them individually.

wind57: its a lot of changes, but they all stem from the above separation. I've written quite a lot of tests to be sure that we are OK with the old and new implementation

wind57:
@ryanjbaxter fix one is ready also: lots of breaking changes were needed to support this. thank you

docs/modules/ROOT/pages/property-source-config/configmap-propertysource.adoc
```diff
@@ -588,6 +588,9 @@ NOTE: If you already have `spring-retry` and `spring-boot-starter-aspectj` on th
 and want to enable fail-fast, but do not want retry to be enabled; you can disable retry for `ConfigMap` `PropertySources`
 by setting `spring.cloud.kubernetes.config.retry.enabled=false`.
 
+
+NOTE: Since version `5.0.0-M3`, we have introduced the possibility to read sources individually. Until now, we would go to the namespace and read all the configmaps / secrets available and then filter out the ones requested. Since `5.0.0-M3` you can specify that you want to read them individually, by setting the property: `spring.cloud.kubernetes.config.read-type=SINGLE`. The previous option to read them all in a namespace is controlled by `spring.cloud.kubernetes.config.read-type=BATCH` and it is the default option.
```
ryanjbaxter: I would drop the `-M3` here, no one will use a milestone once the GA versions is out

wind57: ok, changed

PR #2040, opened 2025-10-05
Drop paths support (2)

wind57:


wind57:
@ryanjbaxter and final one for today : dropping paths support. 

I re-wrote the integration tests to prove that replacing config import with now removed paths still works. I understand that these are big PRs, and I tried to break them apart as much as I could

PR #2041, opened 2025-10-05
Drop enable-api property (3)

wind57:


spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/ConfigMapConfigProperties.java
```diff
@@ -35,9 +35,9 @@
  * @author Isik Erhan
  */
 @ConfigurationProperties(ConfigMapConfigProperties.PREFIX)
```
wind57, line 37: I am proposing to drop : `@DefaultValue("true") boolean enableApi`. 

At the moment, we have two flags : 

```
spring.cloud.kubernetes.config.enabled
spring.cloud.kubernetes.config.enable-api
```

With `spring.cloud.kubernetes.config.enabled` we control path support and also the creation of specific beans. With `spring.cloud.kubernetes.config.enable-api` we control if we are allowed to read configmaps via the api. 

Since we have dropped paths support, we can have a single flag now for controlling everything, so we can drop `spring.cloud.kubernetes.config.enable-api`

---

Since users are supposed to control everything in this regard with a single flag : `spring.cloud.kubernetes.config.enabled`, they should be OK simply because `@DefaultValue("true") boolean enabled`.

For secrets though, `@DefaultValue("false") boolean enabled`, so they will need to switch this flag to `true`. Thus this a breaking change and had to wait for its turn.

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/ConfigMapPropertySourceLocator.java
```diff
@@ -55,21 +55,18 @@ public PropertySource<?> locate(Environment environment) {
...
-						LOG.warn("Failed to load source: " + configMapSource);
-					}
-					else {
-						LOG.debug("Adding config map property source " + propertySource.getName());
-						composite.addFirstPropertySource(propertySource);
-					}
-				});
-			}
+			Set<NormalizedSource> sources = new LinkedHashSet<>(this.properties.determineSources(environment));
+			LOG.debug("Config Map normalized sources : " + sources);
```
wind57, line 59: the only thing that changed here is dropping : 

```
if (this.properties.enableApi()) { 
```

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/SecretsPropertySourceLocator.java
```diff
@@ -61,19 +61,17 @@ public PropertySource<?> locate(Environment environment) {
...
-						LOG.warn("Failed to load source: " + secretSource);
-					}
-					else {
-						LOG.debug("Adding secret property source " + propertySource.getName());
-						composite.addFirstPropertySource(propertySource);
-					}
-				});
-			}
+			uniqueSources.forEach(secretSource -> {
+				MapPropertySource propertySource = getPropertySource(env, secretSource, properties.readType());
```
wind57, line 65: same here, all I am doing is dropping the if in `if (this.properties.enableApi()) {`

wind57:
@ryanjbaxter this one is ready and touches a lot of files, but they all stem from some simple ones, that I commented in the PR

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/ConfigMapConfigProperties.java
ryanjbaxter: In other words the only option now is to use the API to load secrets and config maps since path support is no longer an option so the `enabled` flag covers all options, right?

ryanjbaxter: How did the secrets related tests work before with spring.cloud.kubernetes.secrets.enabled=false by default?

wind57: > In other words the only option now is to use the API to load secrets and config maps since path support is no longer an option so the enabled flag covers all options, right?

correct.

> How did the secrets related tests work before with spring.cloud.kubernetes.secrets.enabled=false by default?

Until now, we were not setting that property at all in tests, and : 

```
@Target({ ElementType.METHOD, ElementType.TYPE })
@Retention(RetentionPolicy.RUNTIME)
@Documented
@Inherited
@ConditionalOnProperty(value = "spring.cloud.kubernetes.secrets.enabled", matchIfMissing = true)
public @interface ConditionalOnKubernetesSecretsEnabled {

}
```

Notice that `matchIfMissing = true`. So there was no need to set anything.

---

I have changed that to `matchIfMissing = false` in this PR. My thought process was that `enableApi = false` until now, so if you wanted to use the API to read properties, you had to _enable_ that. And since we control that now via `enabled`, you still have to enable it specifically. 

The other idea of why I decided to have it as `matchIfMissing = false` is because we are not supposed to read secrets by default, that is why they are secrets after all; it was natural for me when I was thinking about it.

ryanjbaxter: Ah makes sense, path support for secrets was on by default but api support was not, right?

wind57: spot on!

PR #2047, opened 2025-10-06
Unify config properties

wind57:


spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/config/SecretsConfigProperties.java
```diff
@@ -36,81 +30,19 @@
  * @author Isik Erhan
  */
 @ConfigurationProperties(SecretsConfigProperties.PREFIX)
-public record SecretsConfigProperties(@DefaultValue Map<String, String> labels, @DefaultValue List<Source> sources,
-		@DefaultValue("false") boolean enabled, String name, String namespace, boolean useNameAsPrefix,
-		@DefaultValue("true") boolean includeProfileSpecificSources, boolean failFast,
-		@DefaultValue RetryProperties retry, @DefaultValue("BATCH") ReadType readType) {
+public final class SecretsConfigProperties extends SourceConfigProperties {
```
wind57, line 33: this entire PR is based on this (and same configmap) change. 

In the secrets and configmaps normalization and determining the sources, we do the same exact code. Meaning methods `SecretsConfigProperties::determineSources` and `ConfigMapConfigProperties::determineSources`, do the same exact thing. The only difference is where the result is stored.

As such, let's create a SourceConfigProperties that does all that work, and these two classes simply extend it. 

Nothing changes besides this, same code, same logic. 

The many changes in this PR are triggered by the fact that I changed a couple or arguments with places, so that `ConfigMapConfigProperties` and `SecretsConfigProperties` match exactly.

wind57:
@ryanjbaxter one more in main to take a look. Ive left a comment explaining it. Thank you

PR #2053, opened 2025-10-09
major refactor of fabric8 discovery implementation

wind57:


wind57:
this refactors (but honestly just removes deprecations, moves things to some separate auto-configurations as well as makes things less accessible, so it not really refactor per-se) fabric8 discovery client.

wind57:
@ryanjbaxter ready to be looked at, thank you.

PR #2062, opened 2025-10-11
fix-1641: serialization concerns (1)

wind57:


spring-cloud-kubernetes-discovery/src/main/java/org/springframework/cloud/kubernetes/discovery/KubernetesReactiveDiscoveryClient.java
```diff
@@ -47,15 +46,13 @@ public String description() {
 	}
 
 	@Override
-	@Cacheable("serviceinstances")
```
wind57, line 50: I think we should start with all services being non-cacheable and I will work in reverse, to add cacheable separate clients.

Removing this option is a breaking change, but adding separate clients on top, would be not.

wind57:
@ryanjbaxter this is part-1 of some other PRs that will eventually close issue 1641. I will let you know when that issue is completely covered. 

This is now ready to be looked at, thank you.

spring-cloud-kubernetes-discovery/src/main/java/org/springframework/cloud/kubernetes/discovery/KubernetesReactiveDiscoveryClient.java
ryanjbaxter: Lets separate removing the cachable annotations into its own PR to keep things a bit cleaner

wind57: ok, done.

PR #2065, opened 2025-10-11
Move discovery conditionals to separate package (4)

wind57:


wind57:
this is a "boring" PR, but we have some many conditionals in the discovery implementation, that I decided to move them to a dedicated package. Nothing else changes, besides this.

wind57:
@ryanjbaxter this is ready to be looked at. thank you

PR #2069, opened 2025-10-12
Simplify k8s client discovery to drop selective namespaces (6)

wind57:


wind57:
the only reason we have this dance around selective namespaces, is because we did not want to do breaking changes at the time. Since we are allowed to do it now, this code (even if I liked it), has to go. 

Things are much simpler now and very closely look like fabric8 implementation

wind57:
@ryanjbaxter uno mas! gracias!

PR #2070, opened 2025-10-13
Fix 1641 make k8s discovery cacheable

wind57:


docs/modules/ROOT/pages/discovery-client.adoc
```diff
@@ -279,6 +279,22 @@ spring.cloud.kubernetes.http.discovery.catalog.watcher.enabled=true
 - Since http discovery has _two_ components : server and client, we strongly recommend to align versions between them, otherwise things might not work.
 - If you decide to disable catalog watcher, you need to disable it in both server and client.
 
+Since version `5.0.0`, there is the possibility to cache the responses from a discovery client (we do it via the `@Cacheable` annotation). There are two properties to keep in mind here:
```
wind57, line 282: @ryanjbaxter is this OK documentation wise to you?

spring-cloud-kubernetes-client-discovery/src/main/java/org/springframework/cloud/kubernetes/client/discovery/KubernetesClientInformerReactiveHealthAutoConfiguration.java
```diff
@@ -0,0 +1,107 @@
...
+ */
+@Configuration(proxyBeanMethods = false)
+@ConditionalOnSpringCloudKubernetesReactiveDiscovery
+@AutoConfigureBefore({ SimpleReactiveDiscoveryClientAutoConfiguration.class,
+		ReactiveCommonsClientAutoConfiguration.class })
+@AutoConfigureAfter({ ReactiveCompositeDiscoveryClientAutoConfiguration.class,
+		KubernetesDiscoveryPropertiesAutoConfiguration.class,
+		KubernetesClientDiscoveryClientSpelAutoConfiguration.class,
+		KubernetesClientInformerReactiveDiscoveryClientAutoConfiguration.class })
+class KubernetesClientInformerReactiveHealthAutoConfiguration {
```
wind57, line 61: it turns out that `@ConditionalOnBean` does not work the way I thought it does, so we need proper auto-configurations for the reactive health, that would have `@@AutoConfigureAfter(KubernetesClientDiscoveryClientSpelAutoConfiguration)`


essentially let the clients be created first, then the beans for the health

wind57: I've done this for all 3 clients

docs/modules/ROOT/pages/discovery-client.adoc
ryanjbaxter: I would state that by default the non-cacheable versions are enabled and to enable the cachable versions you need to enable one or the other of the below properties.

wind57:
@ryanjbaxter ready. I don't have anything else that is breaking. thank you.

PR #2071, opened 2025-10-14
Fix 1641 issue with wiring + k8-native discovery cacheable implementation

wind57:


spring-cloud-kubernetes-fabric8-discovery/src/main/java/org/springframework/cloud/kubernetes/fabric8/discovery/Fabric8ReactiveDiscoveryClientAutoConfiguration.java
```diff
@@ -77,18 +78,55 @@ Fabric8ReactiveDiscoveryClient fabric8ReactiveDiscoveryClient(KubernetesClient c
 		return new Fabric8ReactiveDiscoveryClient(fabric8DiscoveryClient);
 	}
 
+	@Bean
```
wind57, line 81: we have a situation where there are two discovery clients options:

- cacheable
- non-cacheable 

If we enable the cacheable one, the health indicator is going to fail, since it expects a non-cacheable one.

I did not catch that initially since there was no test to cover that, now there is.

wind57:
@ryanjbaxter this adds k8s-client cacheable discovery implementation + fixes a small issue

ryanjbaxter:
Will the documentation come in a separate PR?

wind57:
Yes, in the next one

PR #2074, opened 2025-10-15
Refactor without using k3s in unit tests

wind57:


wind57:
@ryanjbaxter minor thing to refactor tests not to use a real k3s container. that takes significant time to start and set-up and we can use wiremock instead. thank you

PR #2075, opened 2025-10-16
Fabric8 native leader election

wind57:


spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/leader/election/PodReadyRunner.java
```diff
@@ -0,0 +1,112 @@
...
+			"podReadyExecutor", TTL_MILLIS);
+
+	public CompletableFuture<Void> podReady(BooleanSupplier podReadySupplier) {
+
+		CompletableFuture<Void> podReadyCompletableFuture = new CompletableFuture<>();
+
+		ScheduledFuture<?> scheduledFuture = podReadySelfShutDownScheduler.scheduleWithFixedDelay(() -> {
+
+			if (podReadyCompletableFuture.isDone()) {
+				LOG.info(() -> "pod readiness is known, not running another cycle");
```
wind57: The idea here goes like this:

- there is a `CachedSingleThreadScheduler podReadySelfShutDownScheduler` that will shut down itself when there are no tasks for it to run 
- we issue the call `podReadySelfShutDownScheduler::podReady` that returns a `CompletableFuture`
- the `podReady` method schedules the `podReadySupplier` to run and returns a `ScheduledFuture`
- When `ScheduledFuture` is either done or fails, we complete the `CompletableFuture` with that status. For example:


```
if (podReadySupplier.getAsBoolean()) {
    podReadyCompletableFuture.complete(null);
}
```

- the `attachShutDownHook` makes sure that if the resulting `CompletableFuture` is completed, we also stop the `ScheduledFuture`, thus shutting down the executor; thus not leaving anything hanging.

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/leader/election/CachedSingleThreadScheduler.java
```diff
@@ -0,0 +1,119 @@
...
+
+/**
+ * This is taken from fabric8 with some changes (we need it, so it could be placed in the
+ * common package). A single thread scheduler that will shutdown itself when there are no
+ * more jobs running inside it. When all ScheduledFuture::cancel are called, the queue of
+ * tasks will be empty and there is an internal runnable that checks that.
+ *
+ * @author wind57
+ */
+public final class CachedSingleThreadScheduler {
```
wind57, line 39: the idea of this one was stolen from fabric8. I needed a thread pool that will auto-shutdown when the task that it runs is over. This was a perfect match for the readiness check: once readiness is over, we don't need the scheduler anymore.

And this is what this one achieves. Once the task it is supposed to run is over, it will shutdown itself.

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/leader/election/ConditionalOnLeaderElectionDisabled.java
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * @author wind57
+ */
+@Target({ ElementType.TYPE, ElementType.METHOD })
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+@Inherited
+@Conditional(ConditionalOnLeaderElectionDisabled.OnLeaderElectionDisabled.class)
+public @interface ConditionalOnLeaderElectionDisabled {
```
wind57, line 37: I still want to keep the previous implementation for a while, as such some conditionals that make sure the old one is the default and the new implementation can only be opted it.

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/leader/election/LeaderElectionCallbacks.java
```diff
@@ -0,0 +1,87 @@
...
+import org.springframework.core.log.LogAccessor;
+
+/**
+ * common leader election callbacks that are supposed to be used in both fabric8 and
+ * k8s-native clients.
+ *
+ * @author wind57
+ */
+public class LeaderElectionCallbacks {
+
```
wind57, line 37: these are the callbacks that fabric8 will be calling on various states that happens

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/leader/election/LeaderElectionProperties.java
```diff
@@ -0,0 +1,95 @@
+/*
+ * Copyright 2013-present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
```
wind57, line 6: all the properties that are supported by the new implementation. I tried to detail them as much as possible

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/leader/election/PodReadyRunner.java
```diff
@@ -0,0 +1,122 @@
...
+
+	public PodReadyRunner(String candidateIdentity, String candidateNamespace) {
+		this.candidateIdentity = candidateIdentity;
+		this.candidateNamespace = candidateNamespace;
+	}
+
+	// how often the inner runnable runs, or how much is the scheduler kept alive
+	private static final long TTL_MILLIS = 100;
+
+	private static final LogAccessor LOG = new LogAccessor(PodReadyRunner.class);
```
wind57, line 43: this is the entry point for the readiness check. We submit a `BooleanSupplier` to it - that will execute whatever the readiness check is; and will return back a `CompletableFuture`. 

This `CompletableFuture` is completed when : `podReadySupplier.getAsBoolean()` returns true; and at that point we do : `podReadyCompletableFuture.complete(null);`.

Another way to think about it : when pod is ready, we complete the future. 

There are two things here : 

- `ScheduledFuture` that comes from the `CachedSingleThreadScheduler` where we submit the pod ready task
- `CompletableFuture`, that is the result.

They interact with each other a lot. When `ScheduledFuture` is canceled, for example, so is the `CompletableFuture`; and in reverse.

spring-cloud-kubernetes-fabric8-leader/src/main/java/org/springframework/cloud/kubernetes/fabric8/leader/election/Fabric8LeaderElectionInitiator.java
```diff
@@ -0,0 +1,195 @@
...
+package org.springframework.cloud.kubernetes.fabric8.leader.election;
+
+import java.util.concurrent.CancellationException;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutorService;
+import java.util.function.BooleanSupplier;
+
+import io.fabric8.kubernetes.client.KubernetesClient;
+import io.fabric8.kubernetes.client.extended.leaderelection.LeaderElectionConfig;
+import jakarta.annotation.PostConstruct;
```
wind57, line 26: This is the bread and butter of this PR.

The process goes like this: 

- we first need to wait for the pod to be ready (fabric8 offers `Readiness.isPodReady(pod);`)

How do we do that?

We schedule a task in a thread pool that auto shutdown when its over: `PodReadyRunner::podReady`. The result from that call is a `CompletableFuture`. You can think that the readiness check happens in "ThreadA".
In a different thread : "ThreadB" we block and wait for the pod to be ready:

```
		podReadyWaitingExecutor.execute(() -> {
			try {
				if (waitForPodReady) {
					CompletableFuture<?> ready = attachReadinessLoggerPipeline(podReadyFuture, candidateIdentity);
					blockReadinessCheck(ready);
					startLeaderElection();
				}
				else {
					startLeaderElection();
				}
			}
			catch (Exception e) {
				LOG.error(e, () -> "failure : " + e.getMessage());
			}
		});
```

specifically : `blockReadinessCheck(ready);`

My idea here with "ThreadA" and "ThreadB" was that leader election should not block app start-up; let it run in the background and when its ready, proper callbacks will be called. 

When readiness fails, nothing happens more: we log and that's it. 

----

When readiness passes, we start the leader election process. AT this time, "ThreadA" and its pool will be stopped and will be garbage. "ThreadB" will then run : `startLeaderElection();`.

Leader election is started via native fabric8 support: 

```
leaderFuture = leaderElector(leaderElectionConfig, fabric8KubernetesClient).start();
```

It returns back a `CompletableFuture`, that will be completed ( exceptionally or not ) by either fabric8 internally, or by us, in the `@PreDestroy`.

Also, in "ThreadB" we block on this : 

```
try {
			leaderFuture.get();
}
```

From here there are a few cases possible.

----

Let's suppose we were the leader and then gave up on it, does not matter why. We did not fail, there was no exception: we just gave up on it. In this case, this code path is triggered:

```
else {
				// ok is always null; since it does not represent anything, it just passes
				// the state further
				LOG.info(() -> "leaderFuture finished normally, will re-start it for : " + candidateIdentity);
				podReadyWaitingExecutor.execute(this::startLeaderElection);
			}
```

When that happens, we simply restart leader election, since nothing actually broke.

----

Another example, is when it failed, but with a CancelationException:

```
if (error != null) {
				// only we have a reference to leaderFuture; and if it is canceled, it is
				// only possible from our own preDestroy
				if (error instanceof CancellationException) {
					LOG.info(() -> "cancel was called on the leader initiator : " + candidateIdentity);
					LOG.info(() -> "terminating leadership for : " + candidateIdentity);
				}
```

but it this case, its us calling: 

```
if (leaderFuture != null) {
			// needed to release the lock, in case we are holding it.
			// fabric8 internally expects this one to be called
			LOG.debug(() -> "leaderFuture will be canceled for : " + candidateIdentity);
			leaderFuture.cancel(true);
		}
```

Since we are the only ones that have access to it, so we are the only one possible to cancel it. As such, no restarts, this is a graceful shutdown of the app: we must not do anything besides cleanup. 

----

The last case is when we failed with an Exception ( for example a timeout to the kube-api or something ). In this case, if we restart or not the process depends on a property ( true by default ) : 

```
else {
					LOG.warn(() -> "leader failed with : " + error.getMessage());
					if (leaderElectionProperties.restartOnFailure()) {
						LOG.info(() -> "will restart leader election for : " + candidateIdentity);
						sleep(leaderElectionProperties);
						podReadyWaitingExecutor.execute(this::startLeaderElection);
					}
					else {
						LOG.warn(() -> "leader election is over for : " + candidateIdentity);
					}
				}
```

wind57:
@ryanjbaxter I want to add one more integration test in here, but otherwise I will not be changing the code. So I guess this is ready to be looked at.  I know its a lot and although we talked about this one in summer, its been too much time since then... As such, ask anything, I will try my best to answer. Thank you

docs/modules/ROOT/pages/leader-election.adoc
```diff
@@ -50,4 +50,88 @@ to `false` in `application.[properties | yaml]`:
 [source,properties]
 ----
 management.info.leader.enabled=false
+
```
ryanjbaxter, line 53: Great docs!

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/leader/LeaderUtils.java
```diff
@@ -16,25 +16,84 @@
...
+		boolean serviceAccountNamespaceExists = Files.isRegularFile(serviceAccountPath);
+		if (serviceAccountNamespaceExists) {
+			try {
+				String namespace = new String(Files.readAllBytes(serviceAccountPath)).replace(System.lineSeparator(),
+						"");
+				LOG.info(() -> "read namespace : " + namespace + " from service account " + serviceAccountPath);
+				return Optional.of(namespace);
+			}
+			catch (IOException e) {
+				throw new UncheckedIOException(e);
```
ryanjbaxter: why not just fallback to the return below in this case?

docs/modules/ROOT/pages/leader-election.adoc
```diff
@@ -50,4 +50,88 @@ to `false` in `application.[properties | yaml]`:
...
+The condition is not fulfilled, so it can't become the leader. Same result will be at `12:00:08`, `12:00:10` and so on, until `12:00:16` and this is where the TTL (`leaseDuration`) of the lock will expire and `podB` can acquire it. As such, a lower value of `leaseDuration` will mean a faster acquiring of leadership by other pods.
+
+You might have to give proper RBAC to be able to use this functionality, for example:
+
+[source]
+----
+ - apiGroups: [ "coordination.k8s.io" ]
+   resources: [ "leases", "configmaps" ]
+   verbs: [ "get", "update", "create", "patch"]
```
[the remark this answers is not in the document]
wind57: done

docs/modules/ROOT/pages/leader-election.adoc
```diff
@@ -50,4 +50,88 @@ to `false` in `application.[properties | yaml]`:
...
+
+Once a certain pod establishes itself as the leader (by acquiring the lock), it will continuously (every `spring.cloud.kubernetes.leader.election.retryPeriod`) try to renew its lease, or in other words: it will try to extend its leadership. When a renewal happens, the "record" that is stored inside the lock, is updated. For example, `renewTime` is updated inside the record, to denote when the last renewal happened. (You can always peek inside these fields by using `kubectl describe lease...` for example).
+
+Renewal must happen within a certain interval, specified by `spring.cloud.kubernetes.leader.election.renewDeadline`. By default, it is equal to 10 seconds, and it means that the leader pod has a maximum of 10 seconds to renew its leadership. If that does not happen, this pod loses its leadership and leader election starts again. Because other pods try to become leaders every 2 seconds (by default), it could mean that the pod that just lost leadership, will become leader again. If you want other pods to have a higher chance of becoming leaders, you can set the property (specified in seconds, by default it is 0) :
```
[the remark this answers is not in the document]
wind57: done

docs/modules/ROOT/pages/leader-election.adoc
```diff
@@ -50,4 +50,88 @@ to `false` in `application.[properties | yaml]`:
 [source,properties]
 ----
 management.info.leader.enabled=false
+
+'''
+
+There is another way you can configure leader election, and it comes with native support in the fabric8 library (k8s native client support is not yet implemented). In the long run, this will be the default way to configure leader election, while the previous one will be dropped. You can treat this one much like the JDKs "preview" features.
```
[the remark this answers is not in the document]
wind57: done

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/leader/LeaderUtils.java
```diff
@@ -16,25 +16,84 @@
...
+
 	// k8s environment variable responsible for host name
 	private static final String HOSTNAME = "HOSTNAME";
 
 	private LeaderUtils() {
 
 	}
 
+	/**
+	 * ideally, should always be present. If not, downward api must enable this one.
```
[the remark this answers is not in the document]
wind57: done

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/leader/LeaderUtils.java
wind57: good point. done.

ryanjbaxter:
GitHub actions appears to be having issues according to their status page.  I am going to merge this for now

PR #2088, opened 2025-11-01
fix 2087

wind57:


wind57:
@ryanjbaxter small fix over here, just replaces an import from the recent boot changes

PR #2105, opened 2025-12-19
drop old leader election implementation ( CAN ONLY GO IN THE NEXT MAJOR )

wind57:


docs/modules/ROOT/pages/examples.adoc
```diff
@@ -10,7 +10,6 @@ The same applies for `PropertySourceLocator`, where you need to add to the class
 
 The following projects highlight the usage of these dependencies and demonstrate how you can use these libraries from any Spring Boot application:
 
-* https://github.com/spring-cloud/spring-cloud-kubernetes/tree/main/spring-cloud-kubernetes-examples[Spring Cloud Kubernetes Examples]: the ones located inside this repository.
```
wind57, line 13: the only reason we kept examples, is for leader election not having integration tests where you could look on how to set-up things. This is now resolved, so we can drop it

PR #2112, opened 2025-12-24
Move leader election utilities to common

wind57:


wind57:
I'm going to need the same utilities in k8s client leader election, so I've moved them to commons

wind57:
@ryanjbaxter one more "boring" change. thank you

PR #2119, opened 2026-01-16
Simplify discovery beans for k8s native client

wind57:


spring-cloud-kubernetes-client-discovery/src/main/java/org/springframework/cloud/kubernetes/client/discovery/KubernetesClientInformerAutoConfiguration.java
```diff
@@ -92,85 +92,53 @@ List<String> k8sClientDiscoveryNamespaces(KubernetesDiscoveryProperties properti
 	}
 
 	@Bean
-	@ConditionalOnMissingBean(value = SharedInformerFactory.class, parameterizedContainer = List.class)
```
wind57, line 95: what we were doing until now:

for _each_ namespace:

- create a `SharedInformerFactory`
- create a `GenericKubernetesApi`
- create a `SharedIndexInformer`
- create a `Lister`

This is an overkill, and can be simplified to : 

- create a _single_ `SharedInformerFactory`
- create a _single_ `GenericKubernetesApi`
- create a `SharedIndexInformer` per namespace
- create a `Lister` per namespace

That is the only change I am proposing here.

PR #2120, opened 2026-01-16
create less objects in k8s native discovery configuration

wind57:


wind57:
@ryanjbaxter minor thing over here as I slowly progress towards a bug fix... thank you

spring-cloud-kubernetes-client-discovery/src/main/java/org/springframework/cloud/kubernetes/client/discovery/KubernetesClientInformerAutoConfiguration.java
```diff
@@ -110,13 +110,15 @@ List<SharedIndexInformer<V1Service>> serviceSharedIndexInformers(
 			List<SharedInformerFactory> sharedInformerFactories, List<String> selectiveNamespaces,
 			ApiClient apiClient) {
 
+		GenericKubernetesApi<V1Service, V1ServiceList> servicesApi = new GenericKubernetesApi<>(V1Service.class,
```
wind57, line 113: we use to create a `GenericKubernetesApi` for each namespace, but a single one is enough; thus creating a bit less objects

PR #2121, opened 2026-01-21
change SharedInformer to SharedIndexedInformer

wind57:


wind57:
minor change here, `SharedInformer` -> `SharedIndexedInformer`. One is a subtype of the other, thus compilation worked until now, but I want to make it clear that we are using the _indexed_ informer.

wind57:
@ryanjbaxter this one is ready too

ryanjbaxter:
Do we need to use `SharedIndexInformer`?  It seems better to use the broader subtype if we don't need to use `SharedIndexInformer`.

wind57:
my thought was that we do not want to use the other types besides the indexed informer, because we want to query always by the index. This gets more spot-on in the next work that I am doing, where I add labels support for the informer. 

So its just a way to make things clearer, to me; but its not mandatory by any means if we both know what the type is really the indexed one.

PR #2128, opened 2026-01-30
cleaner names and type

wind57:


wind57:
I changed a `final class` ( non-public) to a `record` and renamed two methods. small changes that get me closer to a kind of trivial bug fix

wind57:
@ryanjbaxter minor thing over here, thank you

PR #2136, opened 2026-02-05
Explicit poll delay

wind57:


wind57:
@ryanjbaxter many files, but a single change really, all calls that have to do with waiting now delegate to a single place.

PR #2137, opened 2026-02-05
Fix failing tests in the commons for leader election

wind57:


spring-cloud-kubernetes-commons/src/test/java/org/springframework/cloud/kubernetes/commons/leader/election/PodReadyRunnerTests.java
```diff
@@ -56,9 +56,8 @@ void readinessOKFromTheFirstCycle(CapturedOutput output) throws Exception {
 		CompletableFuture<Void> readinessFuture = podReadyRunner.podReady(readinessSupplier);
 		readinessFuture.get();
 
-		assertThat(output.getOut()).contains("Pod : identity in namespace : namespace is ready");
```
wind57, line 59: I saw in a different pipeline that these tests were failing sometimes, so I had to take a closer look.

The issue here is a race. We have a scheduled thread that completes a `CompletableFuture` and then there are two threads that : 

- one does `CompletableFuture::get` and continues with assertions on the logs
- the other does : `CompletableFuture::whenComplete` that generate the log entries

That is of course a race, so I'm changing the tests to address that

wind57:
@ryanjbaxter small bug fix in tests. thank you

ryanjbaxter:
Awesome thanks!

PR #2140, opened 2026-02-11
fix 1311

wind57:


docs/modules/ROOT/partials/_configprops.adoc
```diff
@@ -60,7 +60,10 @@
 |spring.cloud.kubernetes.discovery.cache-loading-timeout-seconds | `+++60+++` | timeout for initializing discovery cache, will abort the application if exceeded.
 |spring.cloud.kubernetes.discovery.discovery-server-url |  | 
 |spring.cloud.kubernetes.discovery.enabled | `+++true+++` | if kubernetes discovery is enabled
-|spring.cloud.kubernetes.discovery.filter |  | SpEL expression to filter services after they have been retrieved from the Kubernetes API server.
+|spring.cloud.kubernetes.discovery.filter |  | SpEL expression used to filter Services after they’re fetched from the Kubernetes API server.
```
ryanjbaxter, line 63: This is just going to get overridden when we run the build with `-Pdocs`.  IMO the example should be placed in our `discovery-client.adoc`

wind57:
in this case, no PR is needed, we already have a much better explanation and examples in that file. Do you think we can simply close the ticket than?

PR #2141, opened 2026-02-11
SpEL auto configuration minor improvement

wind57:


spring-cloud-kubernetes-client-discovery/src/main/java/org/springframework/cloud/kubernetes/client/discovery/KubernetesClientDiscoveryClientSpelAutoConfiguration.java
```diff
@@ -48,24 +48,22 @@ public final class KubernetesClientDiscoveryClientSpelAutoConfiguration {
 	@Bean
 	@ConditionalOnMissingBean
 	Predicate<V1Service> predicate(KubernetesDiscoveryProperties properties) {
+
+		String spelExpression = properties.filter();
```
wind57, line 52: this has to be the vast majority of cases, so move it up, do the check and return early, without creating any objects of any kind

wind57:
@ryanjbaxter this one is ready also. thank you

PR #2179, opened 2026-03-12
Pull dockerhub images for tests using containerd to avoid content digest issues

ryanjbaxter:
Previously we pulled images we needed for tests from DockerHub using Docker client saving the image as a tarball and then load them into the test container using containerd.  In some situations this could result in errors when loading the image tarball with containerd using the tarball because the docker client saves the image as a tarball with references for other platforms.  For example if running the tests on OSX when we save the image as a tarball using the docker client the tarball not only contains references to content digests for ARM64 but also AMD64.  When containerd loads the tarball it sees those references to AMD64 but they actually are not part of the tarball and this results in content digest not found errors.  

A more direct approach is to just have containerd pull the images inside the container, this not only avoids the content digest issue, but also avoid the pull, save, load steps we were doing before.

### DIFF
--- a/spring-cloud-kubernetes-test-support/src/main/java/org/springframework/cloud/kubernetes/integration/tests/commons/Commons.java
+++ b/spring-cloud-kubernetes-test-support/src/main/java/org/springframework/cloud/kubernetes/integration/tests/commons/Commons.java
@@ -65,6 +65,10 @@ public final class Commons {
 
 	private static final Log LOG = LogFactory.getLog(Commons.class);
 
+	private static final String DOCKER_IO = "docker.io/";
+
+	private static final String DOCKER_IO_LIBRARY = DOCKER_IO + "library/";
+
 	private Commons() {
 		throw new AssertionError("No instance provided");
 	}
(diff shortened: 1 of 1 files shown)


spring-cloud-kubernetes-test-support/src/main/java/org/springframework/cloud/kubernetes/integration/tests/commons/Commons.java
```diff
@@ -142,16 +146,56 @@ public static void load(K3sContainer container, String tarName, String imageName
...
+	 * Docker save + ctr import, which can fail with "content digest not found" for
+	 * multi-platform or OCI images. If DOCKER_HUB_USERNAME and DOCKER_HUB_PASSWORD are
+	 * set, they are passed as {@code --user username:password} for registry auth.
+	 */
+	private static void pullImageInsideK3s(K3sContainer container, String imageNameForDownload, String imageVersion) {
+		String fullImageRef = fullImageReference(imageNameForDownload, imageVersion);
+
+		final String[] ctrArgs = buildCtrPullArgs(fullImageRef);
+
+		await().atMost(Duration.ofMinutes(2)).pollInterval(Duration.ofSeconds(1)).until(() -> {
```
ryanjbaxter, line 169: I will move to `Awaitilities` when I merge this forward to main

spring-cloud-kubernetes-test-support/src/main/java/org/springframework/cloud/kubernetes/integration/tests/commons/Commons.java
```diff
@@ -142,16 +146,56 @@ public static void load(K3sContainer container, String tarName, String imageName
...
+			boolean noErrors = result.getStderr() == null || result.getStderr().isEmpty();
+			if (!noErrors) {
+				LOG.info("ctr pull stderr: " + result.getStderr());
+			}
+			return noErrors;
+		});
+	}
+
+	private static String[] buildCtrPullArgs(String fullImageRef) {
+		String username = System.getenv("DOCKER_HUB_USERNAME");
```
ryanjbaxter, line 180: This is a small enhacement I added.  One of the advantages of using the docker client on the host to pull the image was we could authenticate with docker hub.  This has several benefits, one of those is to avoid being throttled.  To avoid being throttled by containerd I wanted to supply creds to pull the image.  If these environment properties are set and not empty we will use them as creds when pulling the image.

ryanjbaxter:
@wind57 I closed the previous PR and created one against 3.3.x.  It is virtually the same with one small enhancement to supply creds to dockerhub when pulling the images

wind57:
I am working on something around this code and remembered you did these changes and have a question please. When where you getting the issues with content digest? I am asking because the step you changed, to actually download images we use for tests, should never be called inside github actions. 

That is because we have `restore-docker-images/action.yaml` that will either get images from the cache, or issue a docker pull on them ( for example when we upgrade ). In either case, it will save them to: `/tmp/docker/images` and then we have code that pull those tars from there.

Since we did not have code failing until this PR, I can only assume it was happening either for you locally or somewhere else. I am not debating that this PR is not needed, it very much is an improvement I agree with, just trying to figure out a bit. Thank you.

ryanjbaxter:
Yes it was happening locally and on Jenkins

wind57:
Thank you, that explains it!

PR #2217, opened 2026-04-13
Minor changes in tests part 2

wind57:


wind57:
@ryanjbaxter this one is ready also, I only simplified it a bit, nothing else, all the previous assertions remain in place

PR #2225, opened 2026-04-16
Fix race in tests

wind57:


wind57:
I've only changed tests here, since we had a race. 

- create a configmap with filtering label, assert that it went through the informer 
- delete this configmap
- create a new configmap without label filtering, assert that is did not go through the informer

the issue is that we delete the configmap, informer starts processing that deletion ( that has the correct informer label ), and until the `ConfigReloadUtil::reload` reads the configmap, we have already deleted the previous configmap and created a new one. 

At this point, informer logic has passed ( informer correctly saw a configmap deletion event with a proper label ), and reading now happens on a configmap without a label => a race condition. 

----

I could have fixed it in a different way, but honestly it would have been far less readable. Instead I separated the tests and now its easier to reason.

spring-cloud-kubernetes-integration-tests/spring-cloud-kubernetes-k8s-client-reload/src/test/java/org/springframework/cloud/kubernetes/k8s/client/reload/it/K8sClientConfigMapLabelEventTriggeredIT.java
```diff
@@ -133,9 +131,7 @@ void test(CapturedOutput output) {
 
 		replaceConfigMap(coreV1Api, rightConfigMapAfterChange);
 
-		// sleep for 5 seconds
```
wind57, line 136: I've also changed this test

wind57:
@ryanjbaxter ready, only touches tests and fixes the build failure we saw yesterday

PR #2227, opened 2026-04-17
rename

wind57:


spring-cloud-kubernetes-test-support/src/main/java/org/springframework/cloud/kubernetes/integration/tests/commons/fabric8_client/Fabric8KubernetesFixture.java
```diff
@@ -51,15 +51,15 @@
 /**
  * @author wind57
  */
-public final class Util {
```
wind57, line 54: this and the native client `Util` has been renamed, that's the only change. It was long overdue such a rename, really.

wind57:
@ryanjbaxter ready. I have an idea to change our integration tests to annotation based set-up, much more like spring-boot does it, plus much easier to reason about. Let's see where it takes me... this is the first step that we can do regardless or my future PR.

PR #2234, opened 2026-05-03
Try  local registry

wind57:


wind57:
I was experimenting here with a docker local registry and it works, but I'm not sure I want to continue with it. I'll think about it, but in the meanwhile, I'll close this PR

PR #2243, opened 2026-06-06
docker in docker

wind57:


wind57:
@ryanjbaxter I was looking at the issue you opened and noticed that we need this, if we want to run docker in docker ( for example I was testing that issue you reported in an ubuntu image )

PR #2248, opened 2026-06-08
Cache builder so we don't have to pull it all the time

ryanjbaxter:


.github/workflows/composites/save-pack-cache/action.yaml
```diff
@@ -0,0 +1,28 @@
+name: save pack cache
+description: Save Paketo buildpack layer cache so JVM runtimes are not re-downloaded on subsequent runs
+runs:
+  using: "composite"
+  steps:
+
+    - name: export pack cache volumes to tar
+      shell: bash
+      run: |
```
wind57, line 9: may be we should also cache the `.launch` cache:

```
CACHE_VOLS=$(docker volume ls --format '{{.Name}}' | grep 'pack-cache.*\.\(build\|launch\)$' || true)

```

.github/workflows/composites/save-pack-cache/action.yaml
```diff
@@ -0,0 +1,28 @@
...
+        CACHE_VOLS=$(docker volume ls --format '{{.Name}}' | grep 'pack-cache.*\.build' || true)
+        if [ -z "$CACHE_VOLS" ]; then
+          echo "No pack cache volumes found, skipping save"
+          exit 0
+        fi
+        mkdir -p /tmp/pack-volumes
```
wind57, line 15: I think we can do spin an alpine image only once per all volumes:

```
        mkdir -p /tmp/pack-volumes

        VOLUME_MOUNTS=""
         for vol in $CACHE_VOLS; do
           echo "Saving pack cache volume: $vol"
          VOLUME_MOUNTS="$VOLUME_MOUNTS -v $vol:/in/$vol"
         done

        # mount all selected Docker cache volumes into one helper container
        # loop over them
        # compress each one into its own tar.gz file
        # place those tar.gz files in /tmp/pack-volumes on the host
        # each volume becomes : /tmp/pack-volumes/pack-cache-foo.build.tar.gz on the host 

        docker run --rm \
          $VOLUME_MOUNTS \
          -v /tmp/pack-volumes:/out \
          alpine sh -c '
            for dir in /in/*; do
              vol=$(basename "$dir")
              tar czf "/out/${vol}.tar.gz" -C "$dir" .
            done
          '
```

.github/workflows/composites/restore-pack-cache/action.yaml
```diff
@@ -0,0 +1,31 @@
...
+
+    - name: restore pack cache from actions cache
+      uses: actions/cache/restore@v4
+      id: pack-cache-restore
+      with:
+        path: /tmp/pack-volumes
+        key: pack-cache-${{ runner.os }}-noble-java-tiny-0.0.140
+        restore-keys: pack-cache-${{ runner.os }}-noble-java-tiny-
+
+    - name: load pack cache into docker volumes
```
wind57, line 15: - name: load pack cache into docker volumes
       shell: bash
       if: steps.pack-cache-restore.outputs.cache-hit == 'true'
       run: |
         if [ -d /tmp/pack-volumes ] && [ "$(ls -A /tmp/pack-volumes/*.tar.gz 2>/dev/null)" ]; then
          VOLUME_MOUNTS=""
           for f in /tmp/pack-volumes/*.tar.gz; do
             vol=$(basename "$f" .tar.gz)
             echo "Restoring pack cache volume: $vol"
             docker volume create "$vol" 2>/dev/null || true
            VOLUME_MOUNTS="$VOLUME_MOUNTS -v $vol:/out/$vol"
           done

          # iterate over every cached archive file restored from GitHub Actions cache
          # strip path and .tar.gz
          # extract that archive into the mounted Docker volume director

          docker run --rm \
            -v /tmp/pack-volumes:/in \
            $VOLUME_MOUNTS \
            alpine sh -c '
              for tar_file in /in/*.tar.gz; do
                vol=$(basename "$tar_file" .tar.gz)
                tar xzf "$tar_file" -C "/out/$vol"
              done
            '
         else
           echo "No pack cache volumes found to restore"
         fi

wind57:
@ryanjbaxter if you don't have the time, I can pick this up, its a good enhancement. wdyt?

ryanjbaxter:
Yes that would be great!  I am swamped with other things.  I am trying to solve the annoying issue where CI builds on github actions fails with the error code 51.  I am sure you have seen that.

wind57:
ok, you can close this one then. I started working on it.

PR #2252, opened 2026-06-16
fix reload of config watcher itself

wind57:


wind57:
this is what this fix is for :

- we have configuration watcher
- it can be configured via a configmap ( the properties from `ConfigurationWatcherConfigurationProperties`  are set-up in the configmap )
- it can have the reload enabled for that configmap
- when that happens, we need to re-read those properties ( during a `/refresh` of configuration watcher itself )
- that means we either have, for example: `HttpRefreshTrigger` annotated with `@RefreshScope` or read _always_ from properties, for example via : `k8SConfigurationProperties.getActuatorPath()`. 
- `@RefreshScope` is not an option because the class is final and the interface is sealed.
- the only option is to keep reading from properties

So in two classes this is fixed ( `ConfigMapWatcherChangeDetector` and `SecretsWatcherChangeDetector` ) and in one ( `HttpRefreshTrigger` ) this is clearly documented. 


I hope this makes sense.

wind57:
@ryanjbaxter let me know what you think please

PR #2253, opened 2026-06-16
fix-1816: part 1: rename methods and add proper labels

wind57:


spring-cloud-kubernetes-controllers/spring-cloud-kubernetes-configuration-watcher/src/main/java/org/springframework/cloud/kubernetes/configuration/watcher/KubernetesSource.java
```diff
@@ -32,9 +32,14 @@ String description() {
 		}
 
 		@Override
-		String annotation() {
+		String serviceNamesAnnotation() {
```
wind57, line 35: I started to work on a bug fix and want to do it incrementally.  This change only renames a non-public method and adds a new one, that for the time being, is unused, but will be in the upcoming PRs.

wind57:
@ryanjbaxter started working on a bug fix, that will fix in small PRs

PR #2266, opened 2026-07-02
Restore pack cache

wind57:


.github/workflows/composites/resolve-java-runtime-version/action.yaml
```diff
@@ -0,0 +1,16 @@
+# pack cache must be keyed by the exact installed Java runtime version, not
+# just by the requested major version like 17 or 21. Otherwise, a cache
+# created with one runner image (for example 21.0.1) can be restored on a
+# later runner image (for example 21.0.2), look like a cache hit to GitHub
+# Actions, but still be unusable by Paketo which will re-download the layers.
+
+name: resolve java runtime version
+description: resolve the exact installed Java runtime version for pack cache keys
```
wind57, line 7: we need this one, because the layers have to be cached on the exact version, not on the short one like `17`. This is because if we now cache layers for `17.0.1` and github actions updates to `17.0.2`, store/restore would still work, but the layers would not be re-used, since they are most probably different.

.github/workflows/composites/save-docker-volumes-pack-cache/action.yaml
```diff
@@ -0,0 +1,52 @@
...
+  jdk-version:
+    description: Java version used for the buildpack cache key
+    required: true
+runs:
+  using: "composite"
+  steps:
+
+    - name: export pack cache volumes to tar
+      shell: bash
+      run: |
```
wind57, line 13: this is almost what you had initially Ryan, I just added both `build` and `launch` volumes, plus I spin the alpine image only once, per all layers; but the overall idea is exactly the same

.github/workflows/maven.yaml
```diff
@@ -109,6 +112,28 @@ jobs:
           jdk-version: ${{ env.JDK_VERSION }}
           branch: ${{ env.BRANCH_SAFE }}
 
+      # restore the cache of docker pack volumes.
+      # this is the first restore, the second one is in each matrix instance.
+      # we need to restore here, otherwise if we build, we will re-download everything again.
+      - name: restore pack cache
```
wind57: the comment above says pretty much why we need two steps of restore, this is the first one that kicks in.

wind57:
@ryanjbaxter I'm closing the PR, we can't achieve that. I have tried locally and it fails, the "volume -> tar.gz -> volume" probably does not preserve some metadata and the layers are not re-used.

I've seen that in github actions and I've tried that locally, to understand what is going on, and it does not work either. 

Also, there should be a simpler way, but may be its a bug in spring boot plugin : https://github.com/spring-projects/spring-boot/issues/50915

PR #2273, opened 2026-07-09
Fixes namespace merging across beans and configuration

### LINKED ISSUE #2272: selectiveNamespaces not working
**Describe the bug**
We're currently upgrading vom Spring Boot 3.5 to Spring Boot 4.1 and Spring Cloud 2025.1.2. We came across a weird behavior regarding the kubernetes auto-discovery.

We use Spring Cloud Kubenetes 5.0.2 with the following configuration:
```
spring:
  cloud:
    kubernetes:
      discovery:
        enabled: true
        namespaces:
          - some-namespace
```
 
after upgrading our discovery failed, because instead only querying the single given namespace it tried to do it cluster-wide. The service account, the pod has, doesn't allow this.

After debugging (for hours) we found out that creating a selectiveNamespace**s** (with s at the end) Bean with the desired namespace fixed the problem. Otherwise Spring would just collect some Strings (since it's defined as List<String>) it could find as a Bean.

I think it might be related to this commits: https://github.com/spring-cloud/spring-cloud-kubernetes/commit/df4fc11482f17ba272c4ec769d2362f4b79ead60 https://github.com/spring-cloud/spring-cloud-kubernetes/commit/10acbb91d01070852f7e0e22cf139a17318d698a https://github.com/spring-cloud/spring-cloud-kubernetes/commit/56715184e3483a8f573eecf024e6ba52518db446
They changed the name of the Bean quite a bit, but the final version is missing the **s**.

Or maybe we're doing something wrong. But I guess using some random defined String Beans for this is not the correct way anyway.

wind57:
Fixes #2272

### DIFF
--- a/spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/discovery/SelectiveNamespaces.java
+++ b/spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/discovery/SelectiveNamespaces.java
@@ -1,0 +1,39 @@
+/*
+ * Copyright 2019-present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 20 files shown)


spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/discovery/SelectiveNamespaces.java
```diff
@@ -0,0 +1,39 @@
...
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.kubernetes.commons.discovery;
+
+import java.util.Iterator;
+import java.util.List;
+
+/**
+ * @author wind57
```
wind57, line 23: this is the real fix. Introduce a new record that wrap existing `List<String> selectiveNamespaces` into it, via : `new SelectiveNamespaces(selectiveNamespaces)`.

wind57:
@ryanjbaxter ready for review. thank you

PR #2280, opened 2026-07-27
fix-1828: add HA support for config watcher

wind57:


wind57:
@ryanjbaxter can you trigger copilot in here please?

spring-cloud-kubernetes-test-support/src/main/java/org/springframework/cloud/kubernetes/integration/tests/commons/native_client/NativeClientKubernetesFixture.java
```diff
@@ -721,7 +728,8 @@ private void waitForDeploymentPodsToBeDeleted(Map<String, String> labels, String
 
 	}
 
-	private boolean isDeploymentReady(String deploymentName, String namespace) throws ApiException {
+	private boolean isDeploymentReady(String deploymentName, String namespace, int expectedReplicas)
```
wind57, line 731: in the new IT that I added, there is a need for two replicas, to really test the HA set-up

spring-cloud-kubernetes-test-support/src/main/java/org/springframework/cloud/kubernetes/integration/tests/commons/native_client/NativeClientKubernetesFixture.java
```diff
@@ -522,6 +523,11 @@ public void configWatcher(Phase phase, String refreshDelay, boolean reloadEnable
 			.value("DEBUG"));
 		envVars.add(new V1EnvVar().name("SPRING_CLOUD_KUBERNETES_SECRETS_ENABLED").value("TRUE"));
 
+		if (enableHa) {
+			envVars.add(new V1EnvVar().name("SPRING_CLOUD_KUBERNETES_LEADER_ELECTION_ENABLED").value("true"));
```
wind57, line 527: two properties are needed to enable HA

spring-cloud-kubernetes-integration-tests/spring-cloud-kubernetes-k8s-client-configuration-watcher/src/test/java/org/springframework/cloud/kubernetes/configuration/watcher/KubernetesClientConfigurationWatcherHaIT.java
```diff
@@ -0,0 +1,281 @@
...
+		TestUtil.createConfigMap(fixture, "default");
+	}
+
+	@AfterAll
+	static void afterAll(NativeClientKubernetesFixture fixture) {
+		TestUtil.deleteConfigMap(fixture, "default");
+	}
+
+	/**
+	 * <pre>
```
wind57, line 59: I've added a single IT, that goes through a cycle of leader / no leader / leader

spring-cloud-kubernetes-controllers/spring-cloud-kubernetes-configuration-watcher/src/main/java/org/springframework/cloud/kubernetes/configuration/watcher/ha/ConfigurationWatcherStateStore.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+
+package org.springframework.cloud.kubernetes.configuration.watcher.ha;
+
+/**
+ * Persistent storage for HA watcher state shared across watcher instances.
+ *
+ * @author wind57
+ */
+sealed interface ConfigurationWatcherStateStore permits LeaseConfigurationWatcherStateStore {
```
wind57, line 24: this is the definition of the resource version store. Methods in this one are executed only by the leader

spring-cloud-kubernetes-controllers/spring-cloud-kubernetes-configuration-watcher/src/main/java/org/springframework/cloud/kubernetes/configuration/watcher/ha/LeaseConfigurationWatcherStateStore.java
```diff
@@ -0,0 +1,172 @@
...
+
+	LeaseConfigurationWatcherStateStore(CoordinationV1Api api, ConfigurationWatcherHaProperties properties) {
+		this.api = api;
+		this.leaseName = properties.getLeaseName();
+		String configuredLeaseNamespace = properties.getLeaseNamespace();
+		this.leaseNamespace = StringUtils.hasText(configuredLeaseNamespace) ? configuredLeaseNamespace : "default";
+	}
+
+	@Override
+	public ConfigurationWatcherState readOrCreate() {
```
wind57, line 64: if HA is enabled and this is the first call, this will create an empty lease. Otherwise, it will read whatever is stored there.

We store in the `spring.cloud.kubernetes.configuration.watcher/configmap-resource-version` annotation, something like : "default=123,prod=456", so each namespace tracks its own resource version checkpoint. 

We need such a store because during a downtime when there is no leader at all, resource version can progress ( meaning configmap is updated ), but since there is no leader, events can get lost. As such, we always increment and "store" ( via this implementation ) the most recent resource version we have observed. This happens in the handlers `onAdd / onDelete / onUpdate`. 


So for example:
- we are now the leader and the resourceVersion is at `1`. 
- we lose leadership, so our store stays at `1`. 
- configmap progresses to `resourceVersion=2`
- another leader is elected, it reads the store, sees that it holds `1`
- starts the informer at `resourceVersion=1`, so it can replay events it has not seen

spring-cloud-kubernetes-controllers/spring-cloud-kubernetes-configuration-watcher/src/main/java/org/springframework/cloud/kubernetes/configuration/watcher/ha/ConfigurationWatcherHACoordinator.java
```diff
@@ -0,0 +1,96 @@
...
+			throw new IllegalStateException(
+					"Configuration watcher HA is enabled, but neither ConfigMap nor Secret " + "watching is enabled");
+		}
+		this.configMapDetector = configMapDetector;
+		this.secretsDetector = secretsDetector;
+		this.stateStore = stateStore;
+	}
+
+	@Override
+	public void onApplicationEvent(ApplicationEvent event) {
```
wind57, line 62: these events are triggered only when the instance became the leader

spring-cloud-kubernetes-client-config/src/main/java/org/springframework/cloud/kubernetes/client/config/reload/SecretResourceEventHandler.java
```diff
@@ -19,46 +19,63 @@
...
+		writeResourceVersion(newSecret);
 	}
 
 	@Override
 	public void onDelete(V1Secret secret, boolean deletedFinalStateUnknown) {
-		log.debug(() -> "Secret " + secret.getMetadata().getName() + " was deleted in namespace "
+		LOG.debug(() -> "Secret " + secret.getMetadata().getName() + " was deleted in namespace "
 				+ secret.getMetadata().getNamespace());
 		onEvent.accept(secret);
+		writeResourceVersion(secret);
```
wind57, line 70: after every update that we receive in the handler, write the resource version to the store

spring-cloud-kubernetes-client-config/src/main/java/org/springframework/cloud/kubernetes/client/config/reload/KubernetesClientEventBasedSecretsChangeDetector.java
```diff
@@ -91,11 +104,31 @@ public KubernetesClientEventBasedSecretsChangeDetector(CoreV1Api coreV1Api, Conf
...
+			LOG.info(() -> "config watcher HA is disabled : starting configmap informers immediately");
+			start(Map.of(), null);
+		}
+		else {
+			LOG.info(() -> "config watcher HA is enabled : deferring configmap informer startup "
+					+ "until leadership is acquired");
+		}
+	}
+
+	public final void start(Map<String, String> storedResourceVersions,
```
wind57, line 125: With HA enabled:

- @PostConstruct does nothing, so the informers are not started.
- The native leader-election callback emits `StartLeadingEvent`.
- `ConfigurationWatcherHACoordinator` receives that event.
- It calls `stateStore.readOrCreate()`
  - If the HA state Lease does not exist, it creates it and returns an empty state.
  - Otherwise, it reads the stored `ConfigMap` and `Secret` resource versions.
- The coordinator calls `start(...)` on the available detectors, passing:
  - the stored resource versions
  - a writer that persists future resource versions


- Each detector creates its `InformerResourceVersionResolver`.
- For each watched namespace, the informer sends its initial `LIST` request:
    - if a stored checkpoint exists, the request starts from that resource version
    - otherwise, it starts normally
- The informer then starts its `WATCH`.
- Resource events are handled normally:
    - ConfigMap or Secret change is processed
    - refresh or bus notification is triggered
    - the processed resource version is written to the HA state Lease
- Later `LIST/WATCH` requests use the resource version supplied by the informer, not the stored checkpoint.
- If the leader loses leadership, StopLeadingEvent is emitted.
- The coordinator stops both detectors.
- The next leader reads the persisted state and repeats the process from the stored checkpoints.

wind57:
## TL;DR

This PR adds high-availability support to the Kubernetes client-based Configuration Watcher.

The Configuration Watcher can now run with multiple replicas while ensuring that only one replica actively watches ConfigMaps and Secrets at a time. If the leader fails, another replica acquires leadership, restores the persisted informer checkpoints, and continues processing changes.

## How It Works

HA requires both properties to be enabled:

```
spring:
  cloud:
    kubernetes:
      leader:
        election:
          enabled: true
      configuration:
        watcher:
          ha:
            enabled: true
```

The two properties have separate responsibilities:
- `spring.cloud.kubernetes.leader.election.enabled` enables native Kubernetes leader election.
- `spring.cloud.kubernetes.configuration.watcher.ha.enabled` makes the Configuration Watcher start and stop its informers according to leadership.

When HA is disabled, the existing behavior is unchanged. ConfigMap and Secret informers start during normal bean initialization.

When HA is enabled:

- Informers are not started during initialization.
- The leader-election callback notifies the Configuration Watcher that leadership was acquired.
- The Configuration Watcher reads its persisted state.
- ConfigMap and Secret informers start using the stored resource-version checkpoints.
- When leadership is lost, both informers are stopped.
- The next leader restores the checkpoints and resumes watching.

The HA coordinator is initialized before the leader-election callbacks so that it is ready to receive leadership events.


**Persistent State**

The Configuration Watcher stores its state in a Kubernetes Lease. The leader-election lock and the Configuration Watcher state lease are separate resources.

The state lease stores the last processed resource version independently for:
- each watched namespace
- ConfigMaps
- Secrets

The values are stored in annotations on the lease. For example:

```
spring.cloud.kubernetes.configuration.watcher/configmap-resource-version:
  default=123,other-namespace=456
```

The default configuration is:

```
spring:
  cloud:
    kubernetes:
      configuration:
        watcher:
          ha:
            enabled: false
            lease-name: configuration-watcher-ha
            lease-namespace: default
```

The lease is created automatically when the first leader starts, if it does not already exist.


**Resource-Version Replay**

- When a new leader starts, the persisted resource version is used only for the first informer request for each namespace. This request restores the informer from the exact stored checkpoint.
- After that first request, the informer controls its own resource-version progression. Each subsequent list or watch request uses the resource version supplied by the informer.
- ConfigMap and Secret namespaces are tracked independently, so different namespaces can progress at different resource versions.
- Resource versions are persisted after the corresponding resource event has been handled. This provides at-least-once event processing:
  - If a leader processes an event and persists the checkpoint successfully, the next leader continues after that event.
  - If the leader fails after processing the event but before persisting the checkpoint, the replacement leader may process the event again.

Repeated refresh notifications must therefore be tolerated by refresh targets.

- The watcher does not refresh every ConfigMap or Secret after a leadership change. It only replays events for the configured informer scopes.

- At least two Configuration Watcher replicas are required for failover:

```
spec:
  replicas: 2
```

wind57:
@ryanjbaxter in case you missed this... can you trigger copilot please?

ryanjbaxter:
Have no missed it, just need to set aside some time to look at it.  Probably next week

wind57:
ok, thank you! as usual, Im available on any ways to answer any questions.

spring-cloud-kubernetes-controllers/spring-cloud-kubernetes-configuration-watcher/src/main/java/org/springframework/cloud/kubernetes/configuration/watcher/ha/LeaseConfigurationWatcherStateStore.java
```diff
@@ -0,0 +1,172 @@
...
+		}
+	}
+
+	private void createLease(String leaseName, String leaseNamespace) {
+		try {
+			LOG.info(() -> "Creating watcher HA lease with name : " + leaseName + " in namespace : " + leaseNamespace);
+			api.createNamespacedLease(leaseNamespace, newLease(leaseName, leaseNamespace)).execute();
+		}
+		catch (ApiException e) {
+			LOG.error(e, () -> "Failed to create watcher HA lease '" + e.getResponseBody());
```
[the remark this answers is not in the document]
wind57: fixed

spring-cloud-kubernetes-client-config/src/main/java/org/springframework/cloud/kubernetes/client/config/reload/KubernetesClientEventBasedConfigMapChangeDetector.java
```diff
@@ -89,57 +100,100 @@ public KubernetesClientEventBasedConfigMapChangeDetector(CoreV1Api coreV1Api, Co
...
+				// every change after the checkpoint.
+				// we do not need haEnabled check here, but it short-circuits fast
+				if (haEnabled && !params.watch && params.resourceVersion == null && resourceVersion != null) {
+					request.resourceVersionMatch("Exact");
+				}
+
+				return request.buildCall(null);
+			}, V1ConfigMap.class, V1ConfigMapList.class);
+
+			LOG.debug(() -> "add configmap informer for namespace : " + namespace + " with labels : " + labelSelector);
```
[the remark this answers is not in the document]
wind57: its just a log line...

spring-cloud-kubernetes-client-config/src/main/java/org/springframework/cloud/kubernetes/client/config/reload/KubernetesClientEventBasedSecretsChangeDetector.java
```diff
@@ -91,11 +104,31 @@ public KubernetesClientEventBasedSecretsChangeDetector(CoreV1Api coreV1Api, Conf
...
+		this.haEnabled = haEnabled;
 		namespaces = namespaces(kubernetesNamespaceProvider, properties, "secret");
 	}
 
 	@PostConstruct
 	void inform() {
+		// In HA mode, defer informer startup until this instance acquires leadership.
+		// The leader callback restores the persisted state and then starts the informers.
+		if (!haEnabled) {
+			LOG.info(() -> "config watcher HA is disabled : starting configmap informers immediately");
```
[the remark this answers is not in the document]
wind57: fixed

docs/modules/ROOT/pages/spring-cloud-kubernetes-configuration-watcher.adoc
```diff
@@ -98,6 +98,74 @@ This tells how many milliseconds should we wait before firing the event from con
...
+          enabled: true
+      configuration:
+        watcher:
+          ha:
+            enabled: true
+----
+
+The two settings have different responsibilities:
+
+* `spring.cloud.kubernetes.leader.election.enabled` enables the native Kubernetes leader
```
ryanjbaxter, line 123: If leader eleaction is enabled, why not just have this enabled automatically when `ha.enabled=true`?

spring-cloud-kubernetes-controllers/spring-cloud-kubernetes-configuration-watcher/src/main/java/org/springframework/cloud/kubernetes/configuration/watcher/ha/LeaseConfigurationWatcherStateStore.java
ryanjbaxter: Is that necessary?  In reality if a value was updated while there was no monitor and then updated again the app likely only cares about what the latest value is and don't really care about what it was changed to inbetween.

spring-cloud-kubernetes-controllers/spring-cloud-kubernetes-configuration-watcher/src/main/java/org/springframework/cloud/kubernetes/configuration/watcher/ha/LeaseConfigurationWatcherStateStore.java
```diff
@@ -0,0 +1,172 @@
...
+
+	private final String leaseName;
+
+	private final String leaseNamespace;
+
+	LeaseConfigurationWatcherStateStore(CoordinationV1Api api, ConfigurationWatcherHaProperties properties) {
+		this.api = api;
+		this.leaseName = properties.getLeaseName();
+		String configuredLeaseNamespace = properties.getLeaseNamespace();
+		this.leaseNamespace = StringUtils.hasText(configuredLeaseNamespace) ? configuredLeaseNamespace : "default";
```
ryanjbaxter, line 60: Could we not use `KubernetesNamespaceProvider` here instead of falling back to `default`?

spring-cloud-kubernetes-client-config/src/main/java/org/springframework/cloud/kubernetes/client/config/reload/KubernetesClientEventBasedConfigMapChangeDetector.java
```diff
@@ -89,57 +100,100 @@ public KubernetesClientEventBasedConfigMapChangeDetector(CoreV1Api coreV1Api, Co
...
+
+		ConfigMapResourceEventHandler handler = new ConfigMapResourceEventHandler(this::onEvent, resourceVersionWriter);
+
+		namespaces.forEach(namespace -> {
+			SharedIndexInformer<V1ConfigMap> informer;
+			SharedInformerFactory factory = new SharedInformerFactory(apiClient);
+			factories.add(factory);
+			informer = factory.sharedIndexInformerFor((CallGeneratorParams params) -> {
+
+				String resourceVersion = resourceVersionResolver.resolve(namespace, params.resourceVersion);
```
ryanjbaxter, line 154: This will march the resource consumed before the actual API call to consume the resource, what happens if the subsiquent call does not actually succeed?  The resource will be marked consumed when it hasn't been

spring-cloud-kubernetes-controllers/spring-cloud-kubernetes-configuration-watcher/src/main/java/org/springframework/cloud/kubernetes/configuration/watcher/ha/ConfigurationWatcherHACoordinator.java
```diff
@@ -0,0 +1,96 @@
...
+		}
+		else if (event instanceof StopLeadingEvent stopLeadingEvent) {
+			onStopLeading(stopLeadingEvent);
+		}
+	}
+
+	void onStartLeading(StartLeadingEvent event) {
+		LOG.info(() -> "configuration watcher with identity : " + event.candidateIdentity() + " became leader at : "
+				+ Instant.ofEpochMilli(event.getTimestamp()));
+		ConfigurationWatcherState state = stateStore.readOrCreate();
```
ryanjbaxter, line 74: If this fails and throws an exception is the exception swallowed?  If so doesn't that mean that the detectors wont start?

spring-cloud-kubernetes-controllers/spring-cloud-kubernetes-configuration-watcher/src/main/java/org/springframework/cloud/kubernetes/configuration/watcher/ha/LeaseConfigurationWatcherStateStore.java
```diff
@@ -0,0 +1,172 @@
...
+
+		try {
+			V1Lease lease = api.readNamespacedLease(leaseName, leaseNamespace).execute();
+			Map<String, String> annotations = existingAnnotations(lease);
+			return new ConfigurationWatcherState(parseResourceVersions(annotations.get(CONFIGMAP_ANNOTATION)),
+					parseResourceVersions(annotations.get(SECRET_ANNOTATION)));
+		}
+		catch (ApiException e) {
+			if (e.getCode() == 404) {
+				createLease(leaseName, leaseNamespace);
```
ryanjbaxter, line 76: It might be worth handling a `409` here in `createLease` in the case where there are 2 pods thinking they are the leader at the same time and one wins in the race the create the lease first

spring-cloud-kubernetes-controllers/spring-cloud-kubernetes-configuration-watcher/src/main/java/org/springframework/cloud/kubernetes/configuration/watcher/ha/LeaseConfigurationWatcherStateStore.java
```diff
@@ -0,0 +1,172 @@
...
+	public void writeSecretResourceVersion(String namespace, String resourceVersion) {
+		writeResourceVersion(SECRET_ANNOTATION, namespace, resourceVersion);
+	}
+
+	private void writeResourceVersion(String annotation, String namespace, String resourceVersion) {
+		try {
+
+			LOG.debug(() -> "Updating lease with name: " + leaseName + " in namespace: " + leaseNamespace);
+
+			V1Lease currentLease = api.readNamespacedLease(leaseName, leaseNamespace).execute();
```
ryanjbaxter, line 99: Probably also want to handle a potential `409` here as well

spring-cloud-kubernetes-client-config/src/main/java/org/springframework/cloud/kubernetes/client/config/reload/ConfigMapResourceEventHandler.java
```diff
@@ -55,13 +62,23 @@ public void onUpdate(V1ConfigMap oldConfigMap, V1ConfigMap newConfigMap) {
 		else {
 			onEvent.accept(newConfigMap);
 		}
+		writeResourceVersion(newConfigMap);
```
ryanjbaxter, line 65: Should this only occur during the else branch above?

spring-cloud-kubernetes-client-config/src/main/java/org/springframework/cloud/kubernetes/client/config/reload/SecretResourceEventHandler.java
```diff
@@ -19,46 +19,63 @@
...
 				+ newSecret.getMetadata().getNamespace());
 
 		if (KubernetesClientEventBasedSecretsChangeDetector.equals(oldSecret.getData(), newSecret.getData())) {
-			log.debug(() -> "data in secret has not changed, will not reload");
+			LOG.debug(() -> "data in secret has not changed, will not reload");
 		}
 		else {
 			onEvent.accept(newSecret);
 		}
+		writeResourceVersion(newSecret);
```
ryanjbaxter, line 62: Should this only occur during the else branch above?

ryanjbaxter (changes requested):
This might also be an opportunity to unify the logic in KubernetesClientEventBasedConfigMapChangeDetector and KubernetesClientEventBasedSecretsChangeDetector as well as ConfigMapResourceEventHandler and SecretResourceEventHandler since they are mostly identifcal.

wind57:
I stated looking at your comments, some of them can be made as separate PRs so that its easier to review this way. I'll tag you when those are ready.

docs/modules/ROOT/pages/spring-cloud-kubernetes-configuration-watcher.adoc
wind57: this is indeed a valid point, I changed the code to reflect this

spring-cloud-kubernetes-client-config/src/main/java/org/springframework/cloud/kubernetes/client/config/reload/ConfigMapResourceEventHandler.java
wind57: not relevant anymore, because I removed the writing

spring-cloud-kubernetes-client-config/src/main/java/org/springframework/cloud/kubernetes/client/config/reload/SecretResourceEventHandler.java
wind57: not relevant anymore, because I removed the writing

spring-cloud-kubernetes-controllers/spring-cloud-kubernetes-configuration-watcher/src/main/java/org/springframework/cloud/kubernetes/configuration/watcher/ha/LeaseConfigurationWatcherStateStore.java
wind57: not relevant anymore, since I removed the writing

spring-cloud-kubernetes-controllers/spring-cloud-kubernetes-configuration-watcher/src/main/java/org/springframework/cloud/kubernetes/configuration/watcher/ha/LeaseConfigurationWatcherStateStore.java
wind57: not relevant, as there is no `createLease` anymore

spring-cloud-kubernetes-controllers/spring-cloud-kubernetes-configuration-watcher/src/main/java/org/springframework/cloud/kubernetes/configuration/watcher/ha/ConfigurationWatcherHACoordinator.java
wind57: not relevant anymore

spring-cloud-kubernetes-client-config/src/main/java/org/springframework/cloud/kubernetes/client/config/reload/KubernetesClientEventBasedConfigMapChangeDetector.java
wind57: not relevant anymore

wind57:
> Is that necessary? In reality if a value was updated while there was no monitor and then updated again the app likely only cares about what the latest value is and don't really care about what it was changed to inbetween.

I've been debating about this also, but decided initially to start _with it_, I really happy you commented on this and now that I looked for a few days into it, I agree, it is not needed so I dropped the idea in general. Meaning there is no storage at all happening in the code anymore.

> Could we not use KubernetesNamespaceProvider here instead of falling back to default?

Per the comment above, we do not require this anymore at all. 

> This will march the resource consumed before the actual API call to consume the resource, what happens if the subsiquent call does not actually succeed? The resource will be marked consumed when it hasn't been

Same as above, this logic was dropped.


@ryanjbaxter I've simplified the code and dropped the idea of storing resource versions. You can have another look and also trigger copilot if you may. thank you

spring-cloud-kubernetes-controllers/spring-cloud-kubernetes-configuration-watcher/src/main/java/org/springframework/cloud/kubernetes/configuration/watcher/ha/ConfigurationWatcherHACoordinator.java
```diff
@@ -0,0 +1,80 @@
...
+
+	private final ObjectProvider<@NonNull KubernetesClientEventBasedConfigMapChangeDetector> configMapDetector;
+
+	private final ObjectProvider<@NonNull KubernetesClientEventBasedSecretsChangeDetector> secretsDetector;
+
+	ConfigurationWatcherHACoordinator(
+			ObjectProvider<@NonNull KubernetesClientEventBasedConfigMapChangeDetector> configMapDetector,
+			ObjectProvider<@NonNull KubernetesClientEventBasedSecretsChangeDetector> secretsDetector) {
+		if (configMapDetector.getIfAvailable() == null && secretsDetector.getIfAvailable() == null) {
+			throw new IllegalStateException(
```
ryanjbaxter, line 49: Under what condition would both change detectors not be available?  It feels like that it would be a prerequisite to enabling HA

ryanjbaxter, line 49: In a similar vein what happens if the user sets `leader.election.publish-events=false`?

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/leader/election/ConditionalOnLeaderElectionEnabled.java
```diff
@@ -23,21 +23,19 @@
...
+ * Configuration Watcher HA is enabled.
  *
  * @author wind57
  */
 @Target({ ElementType.TYPE, ElementType.METHOD })
 @Retention(RetentionPolicy.RUNTIME)
 @Documented
 @Inherited
-@ConditionalOnProperty(value = LEADER_ELECTION_ENABLED_PROPERTY, havingValue = "true", matchIfMissing = false)
+@Conditional(LeaderElectionEnabledCondition.class)
```
ryanjbaxter, line 38: I think this change will impact `Fabric8LeaderAutoConfiguration`.  If for some reason `spring.cloud.kubernetes.configuration.watcher.ha.enabled=true` leaks into an app where `Fabric8LeaderAutoConfiguration` is being used today `@ConditionalOnLeaderElectionEnabled` will be true meaning `@ConditionalOnLeaderElectionDisabled` will be false and `Fabric8LeaderAutoConfiguration` won't be used.

```
@ConditionalOnLeaderElectionDisabled
public class Fabric8LeaderAutoConfiguration {
```

spring-cloud-kubernetes-commons/src/main/java/org/springframework/cloud/kubernetes/commons/leader/election/ConditionalOnLeaderElectionEnabled.java
```diff
@@ -23,21 +23,19 @@
...
 public @interface ConditionalOnLeaderElectionEnabled {
```
ryanjbaxter, line 39: Couldn't we do this and get rid of `LeaderElectionEnabledCondition`?

```
@Conditional(ConditionalOnLeaderElectionEnabled.OnLeaderElectionEnabled.class)
public @interface ConditionalOnLeaderElectionEnabled {

    class OnLeaderElectionEnabled extends AnyNestedCondition {
        OnLeaderElectionEnabled() {
            super(ConfigurationPhase.REGISTER_BEAN);
        }

        @ConditionalOnProperty(prefix = LeaderUtils.LEADER_ELECTION_PROPERTY_PREFIX, name = "enabled", havingValue = "true")
        static class OnLeaderElectionPropertyEnabled { }

        @ConditionalOnConfigurationWatcherHAEnabled
        static class OnConfigurationWatcherHaEnabled { }
    }
}
```

spring-cloud-kubernetes-client-config/src/main/java/org/springframework/cloud/kubernetes/client/config/reload/KubernetesClientEventBasedConfigMapChangeDetector.java
```diff
@@ -62,39 +62,67 @@ public class KubernetesClientEventBasedConfigMapChangeDetector extends Kubernete
...
+			LOG.info(() -> "config watcher HA is disabled : starting configmap informers immediately");
+			start();
+		}
+		else {
+			LOG.info(() -> "config watcher HA is enabled : deferring configmap informer startup "
+					+ "until leadership is acquired");
+		}
+	}
+
+	public final void start() {
```
ryanjbaxter, line 103: Does this need to be public?  And if so can we guard against it being called more than once and precent starting the informers again?

spring-cloud-kubernetes-client-config/src/main/java/org/springframework/cloud/kubernetes/client/config/reload/KubernetesClientEventBasedConfigMapChangeDetector.java
```diff
@@ -62,39 +62,67 @@ public class KubernetesClientEventBasedConfigMapChangeDetector extends Kubernete
 	private final KubernetesResourceEventHandler<V1ConfigMap> handler = new KubernetesResourceEventHandler<>(
 			this::onEvent);
 
+	// HA enabled for configuration watcher
+	private final boolean haEnabled;
```
ryanjbaxter, line 66: Having any HA logic baked into this class when it is also used when not doing HA feels wrong to me.  Maybe we can subclass it and stick the logic there?

spring-cloud-kubernetes-controllers/spring-cloud-kubernetes-configuration-watcher/src/main/java/org/springframework/cloud/kubernetes/configuration/watcher/ha/ConfigurationWatcherHACoordinator.java
```diff
@@ -0,0 +1,80 @@
...
+					"Configuration watcher HA is enabled, but neither ConfigMap nor Secret " + "watching is enabled");
+		}
+		this.configMapDetector = configMapDetector;
+		this.secretsDetector = secretsDetector;
+	}
+
+	@Override
+	public void onApplicationEvent(ApplicationEvent event) {
```
ryanjbaxter, line 57: Instead of using an `ApplicationListener` could we use `@EventListener`s here?

```
    @EventListener
    void onStartLeading(StartLeadingEvent event) {
        LOG.info(...);
        configMapDetector.ifAvailable(KubernetesClientEventBasedConfigMapChangeDetector::start);
        secretsDetector.ifAvailable(KubernetesClientEventBasedSecretsChangeDetector::start);
    }

    @EventListener
    void onStopLeading(StopLeadingEvent event) {
        LOG.info(...);
        secretsDetector.ifAvailable(KubernetesClientEventBasedSecretsChangeDetector::stop);
        configMapDetector.ifAvailable(KubernetesClientEventBasedConfigMapChangeDetector::stop);
    }
```

ryanjbaxter (changes requested):

spring-cloud-kubernetes-controllers/spring-cloud-kubernetes-configuration-watcher/src/main/java/org/springframework/cloud/kubernetes/configuration/watcher/ha/ConfigurationWatcherHACoordinator.java
wind57: I did this on purpose. 

`@EventListener` methods are discovered by Spring’s `EventListenerMethodProcessor` after singleton initialization. The leader-election initiator can start during earlier bean lifecycle callbacks and publish `StartLeadingEvent` before that processor has registered the coordinator’s annotated methods.

As such : the coordinator bean existed, but its `@EventListene`r method was not registered yet, so it missed the event.

I actually caught that in the integration test and this is why I changed it.

ryanjbaxter: Got it, that is probably worth calling out in a comment in the code

spring-cloud-kubernetes-client-config/src/main/java/org/springframework/cloud/kubernetes/client/config/reload/KubernetesClientEventBasedConfigMapChangeDetector.java
wind57: this is an interesting idea and initially I really liked it. 

we could easily have `KubernetesClientEventBasedConfigMapChangeDetector` and for example : `KubernetesClientEventBasedConfigMapHAChangeDetector` and their implementations would actually be rather clean ( I tried it ).

But then, the problem would become that lots of new classes would have to be created, because we currently have: 

```
abstract sealed class ConfigMapWatcherChangeDetector extends KubernetesClientEventBasedConfigMapChangeDetector
```

and there are two implementations of the above: `HttpBasedConfigMapWatchChangeDetector` and `BusEventBasedConfigMapWatcherChangeDetector`.

That means we would have to create, for example:

```
ConfigMapWatcherChangeDetector
    extends KubernetesClientEventBasedConfigMapChangeDetector

HAConfigMapWatcherChangeDetector
    extends KubernetesClientEventBasedHAConfigMapChangeDetector
```

same applies to Secrets.

The HTTP/Bus concrete detectors also need normal and HA variants so each auto-configuration creates the correct hierarchy, meaning we would need:

```
HttpBasedConfigMapWatchChangeDetector
    extends ConfigMapWatcherChangeDetector
    // non-HA

HAHttpBasedConfigMapWatchChangeDetector
    extends HAConfigMapWatcherChangeDetector
    // HA
```

and so on. 

I'm going to wait for your input here cause it matters how I work on the other PR comments.

ryanjbaxter: Yeah I see what you mean.

I would be nice if KubernetesClientEventBasedConfigMapChangeDetector was an interface and we had 2 implementations and [Config/Secret]MapWatcherChangeDetector took in the implementation.  

But IDK if that would be possible without breaking the whole class structure...

PR #2288, opened 2026-08-06
move to outer class

wind57:


wind57:
@ryanjbaxter I'm working in HA for config watcher, and some classes become a bit too large, so decided to move the handlers to separate classes in this PR. ready

PR #2294, opened 2026-08-13
Keep Running All Test Jobs and Add Retry Functionality

ryanjbaxter:
Remove fail fast so other test runners continue running on failure. 
Add retry functionality to retry failed tests

### DIFF
--- a/.github/workflows/composites/cache/action.yaml
+++ b/.github/workflows/composites/cache/action.yaml
@@ -13,8 +13,11 @@ runs:
       with:
         attempt_limit: 3
         action: buildjet/cache@v3
+        # '~/.m2/wrapper' holds the maven distribution that ./mvnw downloads for itself.
         with: |
-          path: ~/.m2/repository
-          key: ${{ runner.os }}-cache-jdk${{ inputs.jdk-version }}-${{ env.BRANCH_NAME }}-${{ hashFiles('**/pom.xml') }}
-          restore-keys: ${{ runner.os }}-cache-jdk${{ inputs.jdk-version }}-${{ env.BRANCH_NAME }}-${{ hashFiles('**/pom.xml') }}
+          path: |
+            ~/.m2/repository
+            ~/.m2/wrapper
+          key: ${{ runner.os }}-cache-v2-jdk${{ inputs.jdk-version }}-${{ env.BRANCH_NAME }}-${{ hashFiles('**/pom.xml') }}
+          restore-keys: ${{ runner.os }}-cache-v2-jdk${{ inputs.jdk-version }}-${{ env.BRANCH_NAME }}-${{ hashFiles('**/pom.xml') }}
         
--- a/.github/workflows/composites/scripts/retry-command.sh
+++ b/.github/workflows/composites/scripts/retry-command.sh
@@ -1,0 +1,92 @@
+#!/usr/bin/env bash
+
+# Runs a command and re-runs it from scratch if it fails.
+#
+# This exists for the steps that build the docker images, which happen before any test
+# runs and are therefore not covered by run-tests-with-retry.sh. Those steps fail for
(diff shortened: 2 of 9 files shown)


ryanjbaxter:
@wind57 appreciate your review here

wind57:
that is a very interesting idea :) will take a look most probably tomorrow

wind57:
Suppose a shard contains TestA, TestB, and TestC from moduleA, moduleB, and moduleC. If TestA fails in moduleA, Maven stops the reactor before moduleB and moduleC, so TestB and TestC never run. This is because of the implicit `--fail-fast`

`failed_test_classes` then returns only TestA, because it is the only test with a failure report. The retry consequently runs with `testsToRun=TestA`. If TestA passes, modules without a matching TestA succeed, while running no tests (`failIfNoTests` defaults to false), making the shard green even though TestB and TestC were never executed.

That is the only issue I see here, there is `--fail-at-end` that we can use

ryanjbaxter:
Good catch!  I just pushed a change that should address this.

wind57:
one more edge case and I think we are golden.

Suppose a shard needs to run `TestA,TestB`. Then initially : `outstanding="TestA,TestB"`. 
Now suppose both tests run successfully and produce reports, but Maven later fails during packaging:

```
outstanding = "TestA,TestB"
failed      = ""
never_ran   = ""
```

It is only updated later, immediately before a retry: `outstanding="${retry_set}"`

That causes the problematic comparison:

```
[[ "${never_ran}" == "$(clean_list "${outstanding}")" ]]
```

which is really:

```
[[ "" == "TestA,TestB" ]]
```

IMHO, we should add: 


```
if [[ -z "${failed}" ]]; then
  echo "Maven failed, but no failed test report was found."
  echo "Not retrying."
  return "$rc"
fi
```

ryanjbaxter:
Another good catch!

ryanjbaxter:
And I added another piece to the retry functionality.

Retry when running build images.

This tends to be another flakey part of the CI jobs due to network issues many times.  This has nothing to do with tests so the other retry functionality never ran.

.github/workflows/composites/scripts/retry-command.sh
```diff
@@ -0,0 +1,92 @@
+#!/usr/bin/env bash
+
+# Runs a command and re-runs it from scratch if it fails.
+#
+# This exists for the steps that build the docker images, which happen before any test
+# runs and are therefore not covered by run-tests-with-retry.sh. Those steps fail for
+# reasons that have nothing to do with this project: the paketo buildpacks download a JRE
+# and the buildpack images from the network on every build, so a reset connection or a
+# slow mirror fails the whole shard about twenty minutes before it would have finished.
```
wind57, line 9: I like this one! _a lot_! I can't even tell you how many commits I had over the years that have one message "re-trigger". kudos fot this

.github/workflows/composites/cache/action.yaml
```diff
@@ -13,8 +13,11 @@ runs:
...
         attempt_limit: 3
         action: buildjet/cache@v3
+        # '~/.m2/wrapper' holds the maven distribution that ./mvnw downloads for itself.
         with: |
-          path: ~/.m2/repository
-          key: ${{ runner.os }}-cache-jdk${{ inputs.jdk-version }}-${{ env.BRANCH_NAME }}-${{ hashFiles('**/pom.xml') }}
-          restore-keys: ${{ runner.os }}-cache-jdk${{ inputs.jdk-version }}-${{ env.BRANCH_NAME }}-${{ hashFiles('**/pom.xml') }}
+          path: |
+            ~/.m2/repository
+            ~/.m2/wrapper
```
ryanjbaxter, line 20: By caching the wrapper we won't try and download it over and over again hopefully helping with 429 errors

PR #2300, opened 2026-08-27
move config to the parent project

wind57:


wind57:
@ryanjbaxter I started on adding support for native images and I have a few PRs until I can really start that work. This one is the first one and it is ready. thank you

spring-cloud-kubernetes-controllers/pom.xml
```diff
@@ -29,6 +33,17 @@
 	</build>
 
 	<profiles>
+		<profile>
+			<id>controller-image-name-from-environment</id>
+			<activation>
+				<property>
+					<name>env.IMAGE</name>
```
ryanjbaxter, line 40: Shouldn’t this be negated?

wind57: nope, the old per-module profile used `!env.IMAGE` because it was creating a fallback value only when IMAGE was missing, but now the default is already always set as a normal property:

```
<spring-cloud-kubernetes.controllers.image.name>
  springcloud/${project.artifactId}:${project.version}
</spring-cloud-kubernetes.controllers.image.name>
```

wind57: that being said... is `env.IMAGE` still a thing? It's used only for controllers. If not, we can drop it and my next PR would simplify things even further

ryanjbaxter: Ah its jus overriding it.  Got it.

Not sure if its still a thing to be honest

wind57: can we then "break it"? :) if we hear any complaints, I will revert it

wind57: but for this PR, this is irrelevant anyway... so this one is ready, just FYI

ryanjbaxter: I think we leave it for now as it allows someone to customize the image name if they want

PR #2301, opened 2026-08-27
Move common build plugin

wind57:


wind57:
Moved the common Spring Boot image build configuration to the root pom.xml, so modules that build regular JVM images can inherit the same spring-boot-maven-plugin setup instead of duplicating it.

Current behavior:

- Integration test apps use the inherited root config directly, so their image name remains:
  `docker.io/springcloud/${project.artifactId}:${project.version}`
- Controller apps also inherit the common build-image execution/env config, but keep a small local override for image naming because they already supported `IMAGE` from the environment.
- Controller default image name remains: `springcloud/${project.artifactId}:${project.version}`
- If `${env.IMAGE}` is provided, controllers use that. 

So the duplicated build-image mechanics are centralized, while the controller-specific image naming behavior is preserved.

@ryanjbaxter one more small PR ready. thank you

PR #2305, opened 2026-09-02
Unify Event Handlers

wind57:


wind57:
@ryanjbaxter this one is ready: it unifies the 2 handlers via your ( excellent ) suggestion in the other PR.

ryanjbaxter:
Awesome, any reason this can't go into 5.0.x as well?

wind57:
we did not extract handlers separately in `5.0.x`, we only did in main, so that is not strait-forward atm. I mean `5.0.x` does not have `ConfigMapResourceEventHandler` and `SecretResourceEventHandler`, also there are some other things that `5.0.x` does not have in this regard compared to `main`.

I can try to port everything, but I would not like it :) tell me your input and I will follow it

ryanjbaxter:
Not sure I understand, we just branched 5.0.x from main on Monday of this week, they are practically identical at this point

ryanjbaxter:
The HA PR will only go in main since it is a new feature that I think warrants being in a new feature release, however the simplification here should go in both places as long as it is not a breaking change (and I don't think it is)

PR #2307, opened 2026-09-03
unify event handlers

wind57:


wind57:
@ryanjbaxter Im sorry and thank you, you were right... I was looking in a different place :( it is indeed trivial

PR #2309, opened 2026-09-05
start unifying detectors

wind57:


wind57:
@ryanjbaxter same idea we talked in the PR about HA support

PR #2310, opened 2026-09-06
simpler handlers

wind57:


wind57:
@ryanjbaxter this is the last improvement in k8s-client module from the HA setup discussion. Next I'm going to port the fabric8 to the same logic.

PR #2311, opened 2026-09-06
90 minutes timeout on the mvn job

wind57:


wind57:
Since refactoring the pipeline to retry failed tests and adding --fail-at-end, I’ve noticed that some jobs occasionally fail without retaining any useful logs. For example:

https://github.com/spring-cloud/spring-cloud-kubernetes/actions/runs/34035659441/job/101493657710

In this case, the entire job reached its two-hour timeout. It appears that when GitHub terminates a job at this limit, its logs may not be finalized or included in the downloadable archive. I found several reports describing similar behavior.
To make these failures diagnosable, this change adds a 90-minute timeout to the Maven test steps themselves. If Maven reaches that limit, only the step should time out while the runner remains alive long enough to finalize the job and preserve its logs. The overall shard timeout remains two hours.

wind57:
@ryanjbaxter let me know what you think about this one also please. thank you

ryanjbaxter:
So this is basically lowering the timeout so we can figure out why the job is taking so long?

wind57:
right. because the pipeline is killed, we have no logs at all. 

so instead of killing the pipeline, I am proposing to kill the mvn step itself, and that should preserve the logs. 

I will monitor this of course and see what is going on

PR #2316, opened 2026-09-09
unify handlers in fabric8 client reload

wind57:


wind57:
@ryanjbaxter we have done this for k8s-client, now in fabric8 also.

PR #2317, opened 2026-09-09
fix tests

wind57:


wind57:
While working on another change, I noticed that some tests were failing intermittently. The affected tests shared this inherited static field:

```
protected static KubernetesClient kubernetesClient;
```

Each test could overwrite the shared reference. When one test finished, its cleanup closed the client, potentially closing the client still being used by another test. This caused failures such as IllegalStateException: Client is closed.

Each test now owns its own mock `KubernetesClient` instance.

wind57:
@ryanjbaxter ready for review, it only changes tests. thank you

PR #2321, opened 2026-09-16
fix merge

wind57:


wind57:
@ryanjbaxter this is a left over from some work a while ago, it does not change anything just aligns the code to be exactly like we have it in config map equivalent class.

PR #2322, opened 2026-09-16
fix colliding ports

wind57:


wind57:
The issue was caused by hardcoded `NodePort` values in the integration test manifests.


Several tests created NodePort Services with:

```
nodePort: 32321
```

which means that this port is occupied. 

In the test that I saw failing, Traefik ( most probably ) took the same `32321` port and of course the Service creation failed and this point in time.

I propose here two main changes:

- have a range of ports available for k3s : `32321-32325`
- let k8s assign the NodePort itself, finding one that is not used.

We adjust the code because of that, so that it can read the Service NodePort dinamically, instead of hardcoding it to `32321`.

wind57:
@ryanjbaxter ready to be looked at, fixes some tests that I saw failing. thank you

